PR #4384, opened 2018-05-11
fixing CTS failure issue in master branch

ren-zhijun-oracle:


kguelzau:
@ren-zhijun-oracle this seems to be the fix for CVE-2018-14371.
Is it true that also the 2.2.x branch is affected by this vulnerability? The CVE seems to match _all_ versions below 2.3.7:
https://nvd.nist.gov/vuln/detail/CVE-2018-14371/cpes?expandCpeRanges=true

And I can't see that the 2.2.x is affected by the "loc" parameter:
https://github.com/javaserverfaces/mojarra/blob/166d2173544b40dcfe5359d8ecbb5dda9cb53385/jsf-ri/src/main/java/com/sun/faces/application/resource/ResourceManager.java#L590

Issue #3943, opened 2015-05-22
ScopedAttributeELResolver.setValue() does not check if view is transient before calling getViewMap()


ren-zhijun-oracle:
com.sun.faces.el.ScopedAttributeELResolver.setValue() does not check if view root is transient before calling getViewMap(). This causes warnings like '@ViewScoped beans are not supported on stateless views'.
#### Affected Versions
[2.3.0-m02]

PR #4448, opened 2018-08-20
Prevent ScopedAttributeELResolver to create Session/ViewScope as a si…

kalgon:
Fix for #3943 (previously known as JAVASERVERFACES-3939).

When calling `setValue(...)` or `getValue(...)` on `ScopedAttributeELResolver`, those methods do not check if a session exists before calling `ExternalContext.getSessionMap()` or if a view-scope exists before calling `UIViewRoot.getViewMap()`. This could mean that a superfluous session could be created or that a view-scope could be created in a transient view (and produce some confusing warnings in the logs).

I know this issue has already been closed (for the wrong reason in my opinion) but I would like it to be re-examined.


kalgon:
[Reproducer](https://gist.github.com/kalgon/7770707b316fcd68903abbf16e28013d)

Issue #4458, opened 2018-09-18
Remove code generation for jsf config populator and jsf-ri-runtime.xml


arjantijms:
The Mojarra build generates a temporary file called `jsf-ri-runtime.xml`, from which it generates a Java file `JsfRIRuntimePopulator_Generated`. This last file is used to config Mojarra internally.

While this has some small advantages (keeping the config in sync), the price in build complexity is not quite balanced with the reward.

PR #4459, opened 2018-09-18
#4458 - Remove jsf-ri-runtime.xml generation

arjantijms:


ggam:
This cleanup makes some tools files unused and so those can be removed. If the idea is to remove the tools module altogether, it might be easier to remove files as they are available.

Following @manfredriem advice I made a PR in case you want to make the change https://github.com/arjantijms/mojarra/pull/1/files

PR #4491, opened 2018-10-14
Fix version of parent

pzygielo:
With previous value project cannot be even cleaned against fresh/empty Maven repo.

Signed-off-by: Piotrek Zygielo <piotr@zygielo.pl>


arjantijms:
Thx! It passed the test compile, but probably not the bare clean.

PR #4508, opened 2018-11-01
Bug 4262 - don't restore rowIndex in UIData#invokeOnComponent, if it …

fcaputo:
…was not modified

setRowIndex is called in finally, even if the rowIndex was not modified
by invokeOnComponent. This leads to unexpected method calls. This commit
reduces the scope of the try catch finally.


arjantijms:
@fcaputo Thanks Franks! lgtm!

PR #4512, opened 2018-11-07
Refactor CDI code into same package

manfredriem:


manfredriem:
Signed-off by: Manfred Riem (manfred.riem@oracle.com)

PR #4517, opened 2018-11-23
fix #4516 Ajax update fails due to invalid characters in response XML (DoS)

### LINKED ISSUE #4516: Ajax update fails due to invalid characters in response XML (DoS)
I noticed that the ``<f:ajax />`` update fails when the updated form contains unicode characters like \u000C, which are not allowed in the [XML 1.0 spec](https://www.w3.org/TR/REC-xml/#charsets). 

# 1) Environment
WildFly 14.0.1
Mojarra 2.3.5.SP2
JDK10

# 2) Expected Behaviour
If the update response contains characters that are not allowed in XML, they should be filtered by Mojarra before writing the response.

# 3) Actual Behaviour
Illegal XML characters are not filtered and therefore the browser fails to parse the response. The ajax oncomplete handler tries to read the ``responseXml`` field from the ``XMLHttpRequest`` (see [jsf.js](https://github.com/eclipse-ee4j/mojarra/blob/2ba2a3b5dfd339745a7adba6f5ca77100d97dc68/impl/src/main/resources/META-INF/resources/javax.faces/jsf-uncompressed.js#L2932)). Because the XML contained invalid characters, the browser was not able to parse the response, which causes the ``responseXML`` to be null.

# 4) Steps to reproduce
I created a small github project to reproduce this behaviour: <https://github.com/cnsgithub/mojarra-ajax>
To reproduce:
- ``git clone https://github.com/cnsgithub/mojarra-ajax``
- run ``mvn package wildfly:run`` 
- after the server has started, open <http://localhost:8080/ajax/index.xhtml>
- Click the button, the error should occur

The issue also occurs with user supplied inputs:
- open <http://localhost:8080/ajax/input.xhtml> 
- Enter the [\u000C](https://r12a.github.io/uniview/?charlist=%0C#title) character into the input field
- Click the button

Issue #4145 also seems to describe this problem for an older mojarra version. 

This issue should be addressed with high priority since it is security related (might be exploited for Denial of Service).

cnsgithub:
Implement a FilterWriter to filter characters not allowed according to XML 1.0 spec.

### DIFF
--- a/impl/src/main/java/com/sun/faces/context/IllegalXmlCharacterFilterWriter.java
+++ b/impl/src/main/java/com/sun/faces/context/IllegalXmlCharacterFilterWriter.java
@@ -1,0 +1,56 @@
+package com.sun.faces.context;
+
+import java.io.FilterWriter;
+import java.io.IOException;
+import java.io.Writer;
+
+public class IllegalXmlCharacterFilterWriter extends FilterWriter {
+
+	/**
+	 * Create a new filtered writer.
+	 *
+	 * @param out a Writer object to provide the underlying stream.
+	 * @throws NullPointerException if <code>out</code> is <code>null</code>
+	 */
+	public IllegalXmlCharacterFilterWriter(Writer out) {
+		super(out);
+	}
+
+	@Override
+	public void write(int c) throws IOException {
+		super.write(xmlEncode((char) c));
+	}
+
+	@Override
+	public void write(char[] cbuf, int off, int len) throws IOException {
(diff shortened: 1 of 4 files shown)


cnsgithub:
Just in case someone decides to revive this PR, please have a look at MyFaces' solution (similar code, but much better performance) here: https://github.com/apache/myfaces/blob/master/impl/src/main/java/org/apache/myfaces/util/IllegalXmlCharacterFilterWriter.java

PR #4521, opened 2018-11-30
Do not urlEncode token twice.

hrotkogabor:
When we have a link to a token protected page with param, the token will be urlEncoded twice.

```
<p:link id="button_non_postback" value="Go to protected view with viewParam" outcome="/csrf_protected_page">
  <f:param name="myParam" value="myValue" />
</p:link>
```

Correct example token in the link:
javax.faces.Token=AZ5P3wukOTnNUTQESns2ELAcg3kBTOZfDCEqpdA%3D

actual:
javax.faces.Token=AZ5P3wukOTnNUTQESns2ELAcg3kBTOZfDCEqpdA%**25**3D


manorrock:
@hrotkogabor Please include a reproducer that only uses Mojarra itself, not PrimeFaces.

hrotkogabor:
Attaching a project without primefaces.
[csrfTestMaven.zip](https://github.com/eclipse-ee4j/mojarra/files/2636298/csrfTestMaven.zip)

PR #4522, opened 2018-12-06
port check fix, when checking referer protectedView

hrotkogabor:
There is a case when the referer have no port defined (-1), but the context return non -1 for requestServerPort, e.g. 80.
In this case we should use the port for the referer from the scheme.
http-> 80, https-> 443


arjantijms:
Should be fixed via https://github.com/eclipse-ee4j/mojarra/pull/4579 

If not, this can be re-opened.

PR #4534, opened 2019-01-21
2.3.9 - [Bug-4516] Ajax update fails due to invalid characters in response XML (DoS)

juneau001:
Applied changes that were made by cnsgithub in [PR4517](https://github.com/eclipse-ee4j/mojarra/pull/4517)
- Added com.sun.faces.context.IllegalXmlCharacterWriterFilter
- Added tests/com.sun.faces.context.PartialResponseWriterTest
- Modified com.sun.faces.context.PartialViewContextImpl to invoke IllegalXmlCharacterWriterFilter


impl/src/main/java/com/sun/faces/context/IllegalXmlCharacterFilterWriter.java
```diff
@@ -0,0 +1,56 @@
+package com.sun.faces.context;
+
+ import java.io.FilterWriter;
+import java.io.IOException;
+import java.io.Writer;
+
+ public class IllegalXmlCharacterFilterWriter extends FilterWriter {
```
cnsgithub: As suggested in https://github.com/eclipse-ee4j/mojarra/pull/4517#issuecomment-441930544 I would highly recommend to consider the performance considerations implemented in https://github.com/apache/myfaces/blob/master/impl/src/main/java/org/apache/myfaces/util/IllegalXmlCharacterFilterWriter.java, see https://issues.apache.org/jira/browse/MYFACES-4266 for the reasons.

cnsgithub:
@arjantijms Can you give reasons for closing this PR? This is the second PR addressing #4516 that has been closed now without any comments. #4517 was closed by @mojarra-bot.

cnsgithub:
@juneau001 #4517 contained a bug: `IllegalXmlCharacterWriterFilter.isInvalidChar` should not return `true` for the ends of the character ranges. Especially, `\u0020` (the space character) should not be considered invalid. This might hurt performance. Could you please adjust like so:

````java
    private boolean isInvalidChar(char c) {
        if (Character.isSurrogate(c)) {
            return true;
        }
        if (c == '\u0009' || c == '\n' || c == '\r') {
            return false;
        }
        if (c >= '\u0020' && c <= '\uD7FF') {
            return false;
        }
        if (c >= '\uE000' && c <= '\uFFFD') {
            return false;
        }
        return true;
    }
````

Reference: https://www.w3.org/TR/REC-xml/#charsets.

PR #4537, opened 2019-02-04
Upgrade EE4J Branch Javadoc Plugin and Transaction API

MattGill98:
Build currently fails with NPE in Javadoc plugin and artifact not found for `jakarta.transaction:jakarta.transaction-api:1.3.1`. Both were fixed by upgrading them by one patch version.


arjantijms:
Funny, this should have been resolved for the GF 5.1 release, since otherwise how could it have been build and included ⭕️ 

At any length, the PR should be done against the 2.3 branch now, since EE4J_8 is "done".

Issue #4547, opened 2019-03-05
ELFlash ArrayIndexOutOfBoundsException on invalid Cookie value


codylerum:
From: https://github.com/javaserverfaces/mojarra/issues/4386

This can easily reproduced by sending a csfcfc cooke with value `7aXBAow%3D`

codylerum:
@arjantijms should I port in the version from 2.4 here as well? It appears the encrypt and decrypt methods in https://github.com/eclipse-ee4j/mojarra/blob/master/impl/src/main/java/com/sun/faces/util/ByteArrayGuardAESCTR.java have been simplified.

jasonex7:
This is fixed with PR #4668

didiez:
We are using `2.3.14` and the bug is still there. 
Are there any plans to release another version in the `2.3` branch? The bugfix seems backported to `2.3` in 7144ebcd43ad963dd594ea584ec1d9b98b6f3848

manorrock:
Please fix on 3.0 and master as well

melloware:
This looks like it was already ported to 3.0 and 4.0 and can be closed.

PR #4548, opened 2019-03-05
move decrypt into the try catch so we handle if bytes are not valid

codylerum:
This place the ByteArrayGuardAESCTR.decrypt inside the existing try catch so that general errors such as the follow are properly caught and the cookie is reported as bad. This still allows the  InvalidKeyException  to propagate up to it's specific catch handling.

See: #4547 and https://github.com/javaserverfaces/mojarra/issues/4386

```
java.lang.ArrayIndexOutOfBoundsException
	at java.lang.System.arraycopy(Native Method)
	at com.sun.faces.util.ByteArrayGuardAESCTR.decrypt(ByteArrayGuardAESCTR.java:133)
	at com.sun.faces.context.flash.ELFlash$PreviousNextFlashInfoManager.decode(ELFlash.java:1391)
	at com.sun.faces.context.flash.ELFlash.getCurrentFlashManager(ELFlash.java:1214)
	at com.sun.faces.context.flash.ELFlash.doPrePhaseActions(ELFlash.java:618)
	at com.sun.faces.lifecycle.Phase.handleBeforePhase(Phase.java:190)
	at com.sun.faces.lifecycle.Phase.doPhase(Phase.java:74)
	at com.sun.faces.lifecycle.RestoreViewPhase.doPhase(RestoreViewPhase.java:109)
	at com.sun.faces.lifecycle.LifecycleImpl.execute(LifecycleImpl.java:177)
	at javax.faces.webapp.FacesServlet.executeLifecyle(FacesServlet.java:707)
	at javax.faces.webapp.FacesServlet.service(FacesServlet.java:451)
```

### DIFF
--- a/impl/src/main/java/com/sun/faces/context/flash/ELFlash.java
+++ b/impl/src/main/java/com/sun/faces/context/flash/ELFlash.java
@@ -1379,7 +1379,7 @@ public class ELFlash extends Flash {
             String temp;
             String value;
             
-            String urlDecodedValue = null;
+            String urlDecodedValue;
             
             try {
                 urlDecodedValue = URLDecoder.decode(cookie.getValue(), "UTF-8");
@@ -1387,9 +1387,9 @@ public class ELFlash extends Flash {
                 urlDecodedValue = cookie.getValue();
             }
             
-            value = guard.decrypt(urlDecodedValue);
-            
             try {
+                value = guard.decrypt(urlDecodedValue);
+
                 int i = value.indexOf("_");
 
                 // IMPORTANT: what was "next" when the cookie was
@@ -1442,15 +1442,16 @@ public class ELFlash extends Flash {
                     }
                     nextRequestFlashInfo.setFlashMap(flashMap);
                 }
+            } catch(InvalidKeyException e) {
+                throw e;
             } catch (Throwable t) {
                 context.getAttributes().put(CONSTANTS.ForceSetMaxAgeZero, Boolean.TRUE);
                 if (LOGGER.isLoggable(Level.SEVERE)) {
                     LOGGER.log(Level.SEVERE,
                             "jsf.externalcontext.flash.bad.cookie",
-                            new Object [] { value });
+                            new Object [] { urlDecodedValue });
                 }
             }
-
         }
 
 	/**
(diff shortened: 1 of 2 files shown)


juneau001:
Reviewed code and ran standard test suite.  However, please provide tests for this repair, as it is a standard practice to provide a test and I want to ensure that it doesn't break anything.  Thanks for the contribution, it is appreciated!

vonnai:
Hi, when are you going to merge this fix into master? It's fixed a half of year and nothing in production. It's real problem with not valid cookie value, application using mojarra is unusable. Thanks so much.

codylerum:
@juneau001 do you have a similar test you could point me at to replicate? I'm not familiar with the whole suite and project layout to really know where to look.

vonnai:
Guys, is it really safe, that function decrypt of class ByteArrayGuardAESCTR doesn't check the input? There is not any null check, length check of parameter value, then there is a copy of array which requests the 16 bytes. and thats the problem. You should catch NullPointerException and ArrayIndexOutOfBoundExcetion too. The invalid csfcfc cookie can be e.g:

- AAAABBBBCCCCDDDDEEEEFFFF, which is not decryptable, array copy doens't fail, InvalidKeyException is thrown, Application works
- hjkfsd%kd, array copy failed, ArrayIndexOutOfBoundException is thrown, Application doesn't work, user have to delete cookie.

juneau001:
@codylerum I will look into making a quick test for this one.  I appreciate the contribution.  @vonnai Thanks for the follow up, I appreciate the information.

jasonex7:
I've just open PR #4668, which, in my opinion, is a better a approach to resolve this issue because it deals with class ByteArrayGuardAESCTR which is propagating an unnecessary unchecked exception when cookie value doesn't has IV (Initialization Vector) bytes, instead it should propagate a checked exception indicating the value is not valid.

juneau001:
> I've just open PR #4668, which, in my opinion, is a better a approach to resolve this issue because it deals with class ByteArrayGuardAESCTR which is propagating an unnecessary unchecked exception when cookie value doesn't has IV (Initialization Vector) bytes, instead it should propagate a checked exception indicating the value is not valid.

I will take a look a the PR, thanks.

codylerum:
Handled via #4668

Issue #4551, opened 2019-03-19
Only one CDIValidator is invoked when defined via annotation @FacesValidator


soul2zimate:
https://github.com/eclipse-ee4j/mojarra/blob/2.3.9-RELEASE/impl/src/main/java/com/sun/faces/facelets/tag/jsf/ValidatorTagHandlerDelegateImpl.java#L123-L132 compares class name to skip found validator from existing validators list such as:

```
        for (Validator validator : validators) {
            if (validator.getClass().equals(v.getClass())) {
                found = true;
                break;
            }
        }
        
        if (!found) {
            evh.addValidator(v);
        }
```

However, if there are two CDIValidators, this only takes one instance into account.

When JSF CDI injection is enabled. [InstanceFactory.createValidator](https://github.com/eclipse-ee4j/mojarra/blob/2.3.9-RELEASE/impl/src/main/java/com/sun/faces/application/applicationimpl/InstanceFactory.java#L576) calls [CdiUtils.java](https://github.com/eclipse-ee4j/mojarra/blob/2.3.9-RELEASE/impl/src/main/java/com/sun/faces/cdi/CdiUtils.java#L163) to create new CDIValidators.

If JSF CDI injection is not enabled or configured by xml, createCDIValidator() returns null and InstanceFactory creates validator by keeping its Class type. Thus, it won't be skipped in above ValidatorTagHandlerDelegateImpl comparison.

```
        Validator<?> validator = createCDIValidator(validatorId);
        if (validator != null) {
            return validator;
        }

        validator = newThing(validatorId, validatorMap);
```

PR #4552, opened 2019-03-19
Fix issue 4551, don't compare class name and skip CDIValidator

soul2zimate:
This is a fix for https://github.com/eclipse-ee4j/mojarra/issues/4551


soul2zimate:
Can anyone review this change please?

bmaxwell:
Red Hat JBoss Enterprise Application users have reported this issue where the @FacesValidator only invokes the first validator when it should invoke others such that it behaves the same as when configured via faces-config.xml , can someone please review the proposed changed to make the annotations work inline with the xml configuration ?

juneau001:
If possible, please provide a test for this fix.  It is standard practice to include a test that can be integrated into the test suite to prove the fix repairs the issue.  We appreciate your time and contribution!

PR #4559, opened 2019-04-08
Fix issue 4556, HIGH-LEVEL VULNERABILITY WITHIN MOJARRA JSF V2.2

ruolli:


juneau001:
Can you please provide a test to accompany this change?  It looks okay to me, but I believe the standard practice is to provide a test.  Thanks for your contribution...it is appreciated!

ruolli:
The fix was incomplete, closing this Pull Request. 
Instead, a new one would be raised soon.

PR #4578, opened 2019-06-13
[issue-4577] Protected view JSF page can not be accessed with port 80

tmiyargi:
Fix for #4577 
2.3 PR: #4579
RestoreViewPhase is ignoring default ports 443 and 80
To validate run test Bug22995287IT on port 80 by setting integration.url=http://host/test-javaee7-protectedView

Issue #4577, opened 2019-06-13
Protected view JSF page can not be accessed with port 80


tmiyargi:
With the following conditions, the protected xhtml can not be accessed even if the page transition is valid.
set <protected-views> for a xthml in faces-config.xml
place the xhtml except for the root directory
Application server listens on port 80
The request will be failed with following Exception.

`
Caused by: javax.faces.application.ProtectedViewException: JSF1099: Referer [sic] header value http://localhost/jsfcontroller/aaa/hello.xhtml does not appear to be a protected view. Preventing display of viewId /aaa/welcome.xhtml
	at com.sun.faces.lifecycle.RestoreViewPhase.maybeTakeProtectedViewAction(RestoreViewPhase.java:348)
	at com.sun.faces.lifecycle.RestoreViewPhase.execute(RestoreViewPhase.java:237)
	at com.sun.faces.lifecycle.Phase.doPhase(Phase.java:100)
	at com.sun.faces.lifecycle.RestoreViewPhase.doPhase(RestoreViewPhase.java:133)
	at com.sun.faces.lifecycle.LifecycleImpl.execute(LifecycleImpl.java:201)
	at javax.faces.webapp.FacesServlet.service(FacesServlet.java:670)
	... 46 more`

tmiyargi:
2.3 PR: #4579
master PR: #4578

PR #4579, opened 2019-06-13
[issue-4577] Protected view JSF page can not be accessed with port 80

tmiyargi:
issue: #4577 
master PR: #4578

RestoreViewPhase is ignoring default ports 443 and 80
To validate run test Bug22995287IT on port 80 by setting integration.url=http://host/test-javaee7-protectedView


arjantijms:
Thx! :)

PR #4589, opened 2019-07-15
Remove unused code

pzygielo:
I see it was added intentionally in [javaserverfaces/mojarra](https://github.com/javaserverfaces/mojarra/commit/0129ffe2aa#diff-7c962b50d924954596ed4f8326d6c83aR88), but I'm not following the reason.


arjantijms:
A million thanks, that's unused code indeed that should not have made it in.

PR #4593, opened 2019-07-18
Avoid NPE in ELUtils.addEL3_0_Resolvers when streamELResolver is null

TomasHofman:
This is upstream PR for https://github.com/eclipse-ee4j/mojarra/pull/4594

During previous refactoring [1] of method `ELUtils#addEL3_0_Resolvers`, a block catching `Throwable` was replaced with block catching multiple specific exceptions:

```
        if (getStreamELResolverMethod != null) {
            try {
                ELResolver streamELResolver = (ELResolver) getStreamELResolverMethod.invoke(expressionFactory, (Object[]) null);
                composite.addRootELResolver(streamELResolver);  <-- can throw NPE

                // Assume that if we have getStreamELResolver, then we must have
                // javax.el.staticFieldELResolver
                composite.addRootELResolver((ELResolver) newInstance("javax.el.StaticFieldELResolver"));

-           } catch (Throwable t) {
+           } catch (IllegalAccessException | IllegalArgumentException | InvocationTargetException | InstantiationException t) {
                // This is normal on containers that do not have these ELResolvers
            }
        }
```

The code in the try block can however also throw a NPE, which would've been caught before the refactoring and is not caught now. NPE is thrown inside the `composite.addRootELResolver(streamELResolver)` call (specifically at `DemuxCompositeELResolver:80`), when `streamELResolver` is null.

This NPE then fails deployment.

[1] https://github.com/jboss/mojarra/commit/606a22f2aaef9aa551221eda955b9dee02b5c821#diff-bc4fc2106782297b4905f575b100fa6e

### DIFF
--- a/impl/src/main/java/com/sun/faces/el/ELUtils.java
+++ b/impl/src/main/java/com/sun/faces/el/ELUtils.java
@@ -212,12 +212,13 @@ public class ELUtils {
         if (getStreamELResolverMethod != null) {
             try {
                 ELResolver streamELResolver = (ELResolver) getStreamELResolverMethod.invoke(expressionFactory, (Object[]) null);
-                composite.addRootELResolver(streamELResolver);
+                if (streamELResolver != null) {
+                    composite.addRootELResolver(streamELResolver);
 
-                // Assume that if we have getStreamELResolver, then we must have
-                // javax.el.staticFieldELResolver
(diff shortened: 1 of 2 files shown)


TomasHofman:
This issue also affects Wildfly/JBoss EAP users.

PR #4595, opened 2019-07-22
Escaping CDATA end tags

### LINKED ISSUE #4392: Partial rendering: insufficient CDATA encoding (XSS)
Migrated from https://github.com/javaserverfaces/mojarra/issues/4370

## 1) Environment
- Mojarra version: 2.3.2

## 2) Expected behavior
Using CDATA end tag `]]>` within submitted ajax strings must not break things when included in partial response.

## 3) Actual behavior
Client side partial response processing gets aborted. This issue may lead to partial response (XML) injection/XSS as seen in https://github.com/primefaces/primefaces/issues/3623 and reported in https://github.com/primefaces/primefaces/issues/3629 also.
This issue was also mentioned in https://github.com/javaserverfaces/mojarra/issues/3042 but has not yet been addressed.

## 4) XHTML
```xml
        <h:form id="form">
            <h:commandButton value="Submit" action="#{bean.submit}">
                <f:ajax event="click" />
            </h:commandButton>
            <h:panelGroup id="panel" />
        </h:form>
```

## 5) Bean
```java
@ManagedBean
public class Bean {

    public void submit() {
        FacesContext ctx = FacesContext.getCurrentInstance();
        ExternalContext extContext = ctx.getExternalContext();
        if (ctx.getPartialViewContext().isAjaxRequest()) {
            try {
                extContext.setResponseContentType("text/xml");
                extContext.addResponseHeader("Cache-Control", "no-cache");
                PartialResponseWriter writer = ctx.getPartialViewContext().getPartialResponseWriter();
                writer.startDocument();
                writer.startUpdate("form:panel");
                writer.startElement("span", null);
                writer.writeAttribute("id", "form:panel", null);
                writer.startElement("script", null);
                writer.writeText("alert('User input: " + escapeJavascript(getUserInput()) + "')", null);
                writer.endElement("script");
                writer.endElement("span");
                writer.endUpdate();
                writer.endDocument();
                writer.flush();
                ctx.responseComplete();
            }
            catch (Exception e) {
                throw new FacesException(e);
            }
        }
    }
    
    private String getUserInput() {
        return "foo]]>bar";
    }
    
    private String escapeJavascript(String string) {
        // would use OWASP javascript escaping for example
        // no need to escape ]]> in javascript strings, so it remains unchanged
        return string;
    }
}
```

## 5) Misc
- Possible solution: In `PartialResponseWriter`/`HtmlResponseWriter.writeText` replace `]]>` with `]]]]><![CDATA[>` when `dontEscape && writingCdata == true`.
- Further readings: https://stackoverflow.com/questions/223652/is-there-a-way-to-escape-a-cdata-end-token-in-xml and https://www.owasp.org/index.php/Testing_for_XML_Injection_(OTG-INPVAL-008)

As this might be a security risk, it should be addressed with higher priority. Just ignoring it again seems not to be the right solution.

Btw, you may have a look at owasp's solution to get an idea how to fix it: https://github.com/OWASP/owasp-java-encoder/blob/master/core/src/main/java/org/owasp/encoder/CDATAEncoder.java

TomasHofman:
Fixes #4392

Signed-off-by: Tomas Hofman <thofman@redhat.com>

The issue is that when HtmlResponseWriter is writing CDATA text inside `<script>` or `<styles>` elements, special characters escaping is disabled. However the end CDATA sequence `]]>` should IMO still be escaped, which is the change in this PR. Not escaping `]]>` breaks AJAX response handling, see related issue.

This issue also affects Wildfly/JBoss EAP users.

### DIFF
--- a/impl/src/main/java/com/sun/faces/renderkit/html_basic/HtmlResponseWriter.java
+++ b/impl/src/main/java/com/sun/faces/renderkit/html_basic/HtmlResponseWriter.java
@@ -867,7 +867,12 @@ public class HtmlResponseWriter extends ResponseWriter {
 
         closeStartIfNecessary();
         if (dontEscape) {
-            writer.write(text);
+            if (writingCdata) {
+                charHolder[0] = text;
+                writeUnescapedCData(charHolder, 0, 1);
+            } else {
+                writer.write(text);
+            }
         } else if (isPartial || !writingCdata) {
             charHolder[0] = text;
             HtmlUtils.writeText(writer, escapeUnicode, escapeIso, buffer, charHolder);
@@ -906,7 +911,11 @@ public class HtmlResponseWriter extends ResponseWriter {
         closeStartIfNecessary();
 
         if (dontEscape) {
-            writer.write(text);
+            if (writingCdata) {
+                writeUnescapedCData(text, 0, text.length);
+            } else {
+                writer.write(text);
+            }
         } else if (isPartial || !writingCdata) {
             HtmlUtils.writeText(writer, escapeUnicode, escapeIso, buffer, text);
         } else { // if writingCdata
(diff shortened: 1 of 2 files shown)


xstefank:
@arjantijms review ping, can you take a look, please?

arjantijms:
@xstefank thanks for the ping :) I've reviewed and tested locally. I'll run the TCK against this PR and some of the other ones next.

Issue #4596, opened 2019-07-23
Scripts with CDATA cause "empty response" error on Ajax render


tmiyargi:
Migrated from https://github.com/javaserverfaces/mojarra/issues/4388

Regression caused by fix https://github.com/javaserverfaces/mojarra/issues/4358

PR #4597, opened 2019-07-23
4596 - Scripts with CDATA cause "empty response" error on Ajax render

tmiyargi:
This fixes a regression that is not present in master

Issue: https://github.com/eclipse-ee4j/mojarra/issues/4596


xstefank:
@arjantijms review ping. Can you take a look, please?

PR #4616, opened 2019-08-07
Travis build for master

pzygielo:


arjantijms:
LGTM, the javax jar had been recently updated in 2.3 as well, but good to have it here too. As mentioned in another comment, we may move the current master to a new branch called "prototype" or something, and create a new master based on 2.3 for the javax -> jakarta package renaming.

PR #4617, opened 2019-08-07
Travis build for 2.3 branch

pzygielo:
I can't find build for 2.3 branch in https://jenkins.eclipse.org/mojarra<sup>1</sup>, so perhaps travis could be used to ensure compilability for PRs at least?

But this test https://github.com/eclipse-ee4j/mojarra/blob/5122f2bc9456d0853809759ddd90dab597ac8a77/impl/src/test/java/javax/faces/FactoryFinderTestCase.java#L107

is flickering. On the same sources it [failed earlier](https://travis-ci.org/eclipse-ee4j/mojarra/builds/568951794#L23350), and passed now.

--- 

1. other than [1_mojarra-build-and-stage](https://jenkins.eclipse.org/mojarra/job/1_mojarra-build-and-stage/) which seems to be manual

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,5 @@
+language: java
+
+jdk:
+  - openjdk8
+


arjantijms:
The job should have been https://jenkins.eclipse.org/mojarra/job/Mojarra_2_3_build but it was set to the wrong branch. I've updated it to 2.3 now.

Travis is still handy for PRs indeed.

pzygielo:
Please wait.

pzygielo:
I'd like advice regarding testFacesConfigCase.
IMO it could be disabled, perhaps converted to issue to investigate and re-enable it.

arjantijms:
There's a bunch of unstable tests indeed. Very generally speaking everything in /test/servlet30isolated is very unstable.

I'd have to look at `mojarra/impl/src/test/java/javax/faces/FactoryFinderTestCase.java` again.

pzygielo:
But it's the only one I encountered to fail from time to time in `test` phase locally as well.
I had impression that servlet30-isolated (at least on master) are executed as integration tests, and skipped in UT. Currently presented travis goes only for UTs I think (default script: `mvn test`, and in install phase all tests are skipped).
So if you'd like me to do anything else, I'm open to do changes.

pzygielo:
Also - on master - `TEST_FOLDER` `test/servlet30-isolated` is excluded from execution.

arjantijms:
> Also - on master - TEST_FOLDER test/servlet30-isolated is excluded from execution.

Yes, these tests are based on the old Cactus system, very hard to maintain, and make far too much assumptions about very deep internals most of the time.

pzygielo:
So just in that case of occasional failure - either job can be restarted by admin, retriggered (with new commit, like I did here) or simply ignored by reviewer, right?

pzygielo:
How about [new profile activated on travis only](https://github.com/pzygielo/ee4j-mojarra/compare/travis_2.3...pzygielo:exclude-in-profile?expand=1)?
It would run this one test along others locally and on jenkins.
I would prefer to have this exclusion to save time of unnecessary investigation of kind-of allowed failure.

PR #4626, opened 2019-08-22
Allow sending blobs via AJAX

ggam:
This is a first try to cover the issue I posted to the mailing list: https://www.eclipse.org/lists/mojarra-dev/msg00064.html

I've removed the iframe workaround that's not needed on moderately modern browsers (IE >=10) and use FormData JS object to add the user provided parameters.

With this code, I can now do:
```javascript
canvas.toBlob(function(blob) {
    jsfCommandScript({"blob": blob});
}, "image/jpeg");
```

And receive the part from the backing bean:
```java
public void jsfCommandScript() throws IOException, ServletException {
    var request = (HttpServletRequest) FacesContext.getCurrentInstance().getExternalContext().getRequest();
    request.getParts().stream().filter(p -> p.getSubmittedFileName() != null).forEach(System.out::println);
}
```
The code is not ready and I have only tested my usecase but I'd lke to get some feedback on the approach and whether is acceptable to drop IE 9 support for Mojarra 3.0

### DIFF
--- a/impl/src/main/resources/META-INF/resources/javax.faces/jsf-uncompressed.js
+++ b/impl/src/main/resources/META-INF/resources/javax.faces/jsf-uncompressed.js
@@ -123,229 +123,12 @@ if (!((jsf && jsf.specversion && jsf.specversion >= 23000 ) &&
          * @ignore
          */
         var getTransport = function getTransport(context) {
-            var returnVal;
-            // Here we check for encoding type for file upload(s).
-            // This is where we would also include a check for the existence of
-            // input file control for the current form (see hasInputFileControl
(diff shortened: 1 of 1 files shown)


arjantijms:
@BalusC can you take a look at this?

manorrock:
This sounds like a specification enhancement where we need to make sure jsf.js properly supports binary data instead of hacking it in.

Note on the notion of dropping support for older browsers, now is the time to drop support for older browsers in the version under development. We cannot drop support in an already released version.

arjantijms:
@ggam if you still support this, can you work with @manorrock on a specification enhancement for this? I'll close this one for now. Thanks!

Issue #4631, opened 2019-09-13
Memory leak in FlashScope - expired elements are not cleared


spyrkob:
Objects in FlashScope is not cleared and keep accumulating even if POST-REDIRECT-BACK is completed.

If we put a huge object to FlashScope continuously as the following code, Java heap space will be exhausted soon.

```
@Model
public class Action {

    private static final int oneHundredMB = 1024 * 1024 * 100;

    public String next() {
        Flash flash = FacesContext.getCurrentInstance().getExternalContext().getFlash();
        flash.putNow("huge", new byte[oneHundredMB]);
        return "index.xhtml?faces-redirect=true";
    }
}

```
When OutOfMemoryError is happened, flashInnerMap field in com.sun.faces.context.flash.ELFlash is bloated.

```
Class Name                                                             | Ref. Objects | Shallow Heap | Ref. Shallow Heap | Retained Heap
-----------------------------------------------------------------------------------------------------------------------------------------
java.lang.Thread @ 0xeea24380  default task-1 Thread                   |            1 |          120 |                64 |        20,800
'- <Java Local> com.sun.faces.context.flash.ELFlash @ 0xee9e4608       |            1 |           40 |                64 |            64
   '- flashInnerMap java.util.concurrent.ConcurrentHashMap @ 0xee9e4630|            1 |           64 |                64 |   314,574,048
-----------------------------------------------------------------------------------------------------------------------------------------
```

spyrkob:
One solution I found is expiring the 'previous' map at the end of ResponseRender phase.
If I understand the code right, at this point 'previous' map is either no longer needed - the values to pass to next cycle are in 'next' map at this point.
I tried to write some simple tests and I didn't see any regressions, but I don't know if there are any cases I missed.

arjantijms:
To be absolutely honest, none of the current Mojarra developers really dares to touch that code :(

It's all just utterly weird and seems to have been coded for something else initially, then somehow re-used to implement The Flash.

spyrkob:
@arjantijms it does look confusing :) 

I found my initial idea was wrong and it caused some regression. Instead less invasive fix might be making sure the 'next' map is removed from innerMaps during cookie decoding. Currently it's just set to null which means it's never properly cleared up.
Updated the PR

erickdeoliveiraleal:
please, merge it into 2.3 branch

PR #4633, opened 2019-09-18
Memory leak in FlashScope - expired elements are not cleared

spyrkob:
Issue: #4631 

Expire 'previous' flash map and empty 'next' map at the end of RenderResponse phase. 
If the 'previous' map is needed in next request cycle, it would be moved to 'next' map - otherwise it seems it can be safely removed.

### DIFF
--- a/impl/src/main/java/com/sun/faces/context/flash/ELFlash.java
+++ b/impl/src/main/java/com/sun/faces/context/flash/ELFlash.java
@@ -1083,13 +1083,6 @@ public class ELFlash extends Flash {
             }
             contextMap.put(CONSTANTS.DidWriteCookieAttributeName, Boolean.TRUE);
         } else {
-            // if the cookie is removed, make sure the empty maps are removed from innerMap
-            if (nextFlash != null) {
-                flashManager.expireNext();
-            }
-            if (prevFlash != null) {
-                flashManager.expirePrevious();
-            }
             removeCookie(extContext, toSet);
         }
     }
--- a/impl/src/main/java/com/sun/faces/context/flash/ELFlash.java
+++ b/impl/src/main/java/com/sun/faces/context/flash/ELFlash.java
@@ -1083,6 +1083,13 @@ public class ELFlash extends Flash {
             }
             contextMap.put(CONSTANTS.DidWriteCookieAttributeName, Boolean.TRUE);
         } else {
+            // if the cookie is removed, make sure the empty maps are removed from innerMap
+            if (nextFlash != null) {
+                flashManager.expireNext();
+            }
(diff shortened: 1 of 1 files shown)


fjuma:
@arjantijms We have a Red Hat JBoss EAP customer that is running into this issue and needs a fix. Would you be able to review this PR to see if the proposed fix looks ok to you? Thanks!

arjantijms:
@fjuma thanks for the ping. At a quick glance it seems okay, though the FlashScope implementation in Mojarra is very troublesome and brittle, and every fix often breaks something else. I'll do the review today, or tomorrow morning (CEST).

Although EAP of course does its own builds, the plan for Mojarra itself is to have the next build with the contributed fixes from last month out by the end of this week.

fjuma:
Thanks for the quick response, @arjantijms, it's much appreciated!

spyrkob:
Hi @arjantijms, did you have a chance to take a look at this PR? If I can help with anything, let me know!

arjantijms:
@spyrkob There has been a little delay due to other tasks, but release should still go out this week.

Something that's always helpful is running the Mojarra tests on the changed code,

spyrkob:
@arjantijms I run the tests, but I used the travis.yml as template and assumed the commented out tests are outdated or disused and shouldn't be run.

I just tried running the other tests. I was only able to run test/servlet30 successfully, others were failing even without any code modifications - I'm not sure if they need some additional setup.

I found one of the tests in the servlet30 suite fails because of this PR. This should be fixed by the latest update.

Sorry about the trouble...

arjantijms:
@spyrkob thanks for testing. 

The best target to do the patch against and therefor also test against is the 2.3 branch. Master is actually for the coming 3.0 release, but due to all changes necessary for EE4J has fallen behind. We're probably going to prune it soon and replace it with the current 2.3 branch, and then work from there again.

For running the tests you need to set up a GlassFish instance and then define properties for it (for example in a settings.xml in .m2):

```xml
<glassfish.cargo.home>/opt/glassfish5</glassfish.cargo.home>
 <glassfish.patch.home>/opt/glassfish5</glassfish.patch.home>
```

Then after having build Mojarra using `mvn clean install`, cd in the test folder and then execute `./bin/run-tests.sh `

erickdeoliveiraleal:
this pull request should be merged into 2.3 branch too

BalusC:
Checking ...

BalusC:
https://github.com/eclipse-ee4j/mojarra/commit/bbc9d4d5108b5a1d5432ff1ce884ae2fdfc5db0b and https://github.com/eclipse-ee4j/mojarra/pull/4653 are already present in 2.3.

PR #4669, opened 2020-02-14
Don't install and deploy tests

dmatej:
It is not useful to install and deploy test artifacts to any repository.


dmatej:
Seems after last changes this is not happening any more, so I am closing the PR.

Issue #4721, opened 2020-06-11
Cannot build 2.3 branch based on README


codylerum:
Trying to build so I can create a test and debug an issue, but I'm getting errors trying to build the 2.3 branch based on the readme 

I'm running into this issue as well as a similar issue with a disabled h:selectOneCheckbox as referenced in https://issues.redhat.com/browse/WFLY-13571

The readme states JDK8, Ant and Maven which I have installed and then a `mvn clean install` in the root of the project.

```
codylerum@BORIS:~/projects/mojarra$ java -version
openjdk version "1.8.0_252"
OpenJDK Runtime Environment (build 1.8.0_252-8u252-b09-1ubuntu1-b09)
OpenJDK 64-Bit Server VM (build 25.252-b09, mixed mode)

codylerum@BORIS:~/projects/mojarra$ ant -version
Apache Ant(TM) version 1.10.7 compiled on October 24 2019

codylerum@BORIS:~/projects/mojarra$ mvn -version
Apache Maven 3.6.3
Maven home: /usr/share/maven
Java version: 1.8.0_252, vendor: Private Build, runtime: /usr/lib/jvm/java-8-openjdk-amd64/jre
Default locale: en, platform encoding: UTF-8
OS name: "linux", version: "4.19.104-microsoft-standard", arch: "amd64", family: "unix"

codylerum@BORIS:~/projects/mojarra$ mvn clean install
[INFO] Scanning for projects...
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[FATAL] Non-resolvable parent POM for com.sun.faces.test.javaee6:viewParamNullValueAjax:3.0.0-m01-SNAPSHOT: Could not find artifact com.sun.faces.test.javaee6:pom:pom:3.0.0-m01-SNAPSHOT and 'parent.relativePath' points at wrong local POM @ line 21, column 11
 @
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]
[ERROR]   The project com.sun.faces.test.javaee6:viewParamNullValueAjax:3.0.0-m01-SNAPSHOT (/home/codylerum/projects/mojarra/test/javaee6/viewParamNullValueAjax/pom.xml) has 1 error
[...]

PR #4722, opened 2020-06-11
Update test pom version to fix 2.3 build

codylerum:


codylerum:
Fix for #4721

arjantijms:
Thx! Hope this is the final time and it never comes back again ;)

PR #4723, opened 2020-06-12
travis config for 2.3 branch

codylerum:
Here is a PR that enables a travis build for the 2.3 branch. This should help new committers to more easily test their changes.

One thing is that the versions for the root pom and the tests seemed out of sync and caused failures. In this PR they are now all at 2.3.15-SNAPSHOT which matches the project root pom

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,36 @@
+os: linux
+dist: bionic
+
+language: java
+
+jdk:
+  - openjdk8
+
+env:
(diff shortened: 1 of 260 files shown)


arjantijms:
Thanks! Do note that we also have a CI for Mojarra at http://ci.eclipse.org/mojarra, but that's only really accessible to committers and can't easily take PRs. (we can add a Jenkinsfile to this project, but it's a bit of a hassle at the moment)

Issue #4734, opened 2020-07-22
ViewParam + Ajax Call issue


erickdeoliveiraleal:
I still have a problem with JSF 2.3 related to ViewParam, I tested it with WildFly 21 (released october 2020) and the bug is still there. The code is in the following repositories: Jetty example: https://github.com/erickdeoliveiraleal/primefaces-test/tree/viewparam Wildfly example https://gitlab.com/erickdeoliveiraleal/simpleproject

Note that Wildfly already incorporated all the fixes recently done recently in Mojarra 2.3 repository.

I posted a issue in official repo but got no success to resolve the issue: https://github.com/eclipse-ee4j/mojarra/issues/4734

When you click on button the second time a exception happens, this is not the case when using mojarra 2.2, myfaces 2.2 or myfaces 2.3

XHTML:
```
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml"
	xmlns:h="http://java.sun.com/jsf/html"
	xmlns:f="http://java.sun.com/jsf/core"
	xmlns:ui="http://java.sun.com/jsf/facelets"
	xmlns:fn="http://java.sun.com/jsp/jstl/functions">

<f:metadata>
	<f:viewParam id="id" name="id" value="#{testView.testClass.id}" />
</f:metadata>

<h:form>

	<h:commandButton>
		<f:ajax execute="@form"  />
	</h:commandButton>

	<h:selectOneMenu value="#{testView.testClass}">
		<f:selectItem itemValue="null" itemLabel="Sel. classe" />
		<f:selectItems value="#{testView.testClasses}" var="cl"
			itemValue="#{cl}" itemLabel="#{cl.id}" />
	</h:selectOneMenu>

</h:form>

</html>
```
Bean
```
package org.primefaces.test;

import java.io.Serializable;
import java.util.ArrayList;
import java.util.List;

import javax.annotation.PostConstruct;
import javax.faces.view.ViewScoped;
import javax.inject.Named;

@Named
@ViewScoped
public class TestView implements Serializable {

	private TestClass testClass;
	private List<TestClass> testClasses;

	@PostConstruct
	public void init() {
		testClass = new TestClass();
		testClasses = new ArrayList<>();
		testClasses.add(new TestClass(1));
	}

	public TestClass getTestClass() {
		return testClass;
	}

	public void setTestClass(TestClass testClass) {
		this.testClass = testClass;
	}

	public List<TestClass> getTestClasses() {
[...]

PR #4801, opened 2020-11-15
https://github.com/eclipse-ee4j/mojarra/issues/4734

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/4734

Fix PropertyNotFoundException when nested entity is null while basic input renderer attempts to locate the converter by value expression type even though the submitted value is null


soul2zimate:
I made a quick test with this change on WildFly master, The PropertyNotFoundException mentioned in #4734 no longer appears (since it was caught and logged at FINE level).

melloware (approved):

Issue #1498, opened 2010-01-10
onevent handler on f:ajax is broken


ren-zhijun-oracle:
Simple example:
<h:commandLink value="Click me">
<f:ajax onevent="callback" />
</h:commandLink>

JS is defined as:
callback = function()
{
alert('Moin);
}

Rendered HTML:
<a onclick="mojarra.ab(this,event,'action',0,0,

{'onevent':callback}

);return
false" href="#" id="j_idt12" name="j_idt12">Click me</a>

When I click the link, ajax request is triggered (OK), BUT I get three alert()
calls saying "undefined" ...

==> version: 2.0.2-b10
#### Environment
Operating System: All
Platform: All
#### Affected Versions
[2.0.2]

PR #4804, opened 2020-11-16
#1497+ #1498

tandraschko:


arjantijms:
Looks good, but the ECA check failed. To pass is you need to create an Eclipse account and accept the ECA agreement. After that the commit needs to be signed-off.

See the Eclipse Contributor Checkmark in for instance my Eclipse account: 

https://accounts.eclipse.org/users/atijmse7y

See example of the sign-off for e.g. this commit:

https://github.com/eclipse-ee4j/mojarra/commit/5a15a98a74296874af9c48ee8779925f556f04cb

tandraschko:
@arjantijms 
Its here: https://accounts.eclipse.org/users/tandraschko

how can i add a footer in the commit? AFAICS i would need to add: Signed-off-by: tandraschko <andraschko.thomas@gmail.com>

arjantijms:
You can use the sing-off feature of git, see e.g. https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for 

Most IDEs have a checkbox for it. But you can also just have it as the last line of the commit comment. If you amend your last commit with as comment the signoff, then force push, it should work.

PR #4808, opened 2020-11-16
Performance-optimized version of ComponentSupport.findChildByTagId()

zjavaperf:
Hi,

this is Marc from IBM. Not sure where the right place to say "Hello" is, so I will try here. Since I haven't contributed to Mojarra yet, let me quickly introduce myself. I am a Java Performance Analyst and work in the IBM Lab in Boeblingen, Germany. Nowadays, I spend most of my time improving the performance of IBM products, but I also have a long history of analyzing and optimizing the performance of customer Java applications.

One of our IBM Mainframe customers is running their COVID-19 tracing application based on Eclipse Mojarra and they asked me to analyze its performance, since they are struggling with the high amount of CPU that their application consumes. I found a number of tuning opportunities in the Mojarra code, and this patch here is one of them. Since the customer hasn't been able to test my patches yet, I can only provide an educated guess, but my expectation is that the combination of all patches will save ca. 20-25% of all Java method calls in the customer's situation.

Let me explain my `findChildByTagId()` patch a bit: In the original version of the code, the logic in `ComponentSupport.java` performs a BFS for the child with `MARK_ID` "id". While this works very well in small component trees, it quickly becomes quite a heavy call when the component tree underneath of the "parent" `UIComponent` is large. In the customer's situation, I have seen tree depths of up to 12 and breadths of the individual nodes of more than 140 in the traces. So you can imagine how many visits it may take until the search is complete. This is particularly heavy when the child doesn't exist and the entire tree has to be searched almost all of the time.

What my patch does is that it stores the `MARK_ID`s of **all** descendants of a `UIComponent` in a hash map that belongs to the component, along with references to the descendant components. I called this the `descendantMarkIdCache`. Now when `findChildByTagId()` is called in `ComponentSupport`, the only operation that has to be performed is a lookup the `descendantMarkIdCache` hash map of "parent". There is no more searching in the component tree.

Of course, remembering all descendants of a component in a hash map comes at a certain cost, but I have done a number of measurements in order to validate the effectiveness of my patch and the outcome was that it is still much cheaper compared to the current BFS approach. In one test case, the pages/sec throughput increased by ca. 23%, which is equivalent to ca. 18% savings in CPU consumption. The more complex - i.e. nested - the Facelet is, the more the `descendantMarkIdCache` pays off.

I still have to perform a more in depth analysis regarding the additional Java heap consumption caused by the `findChildByTagId()` patch, but from my initial cursory check the increase in heap consumption should be rather moderate.

[...]

PR #4809, opened 2020-11-16
Performance-optimized version of ResourceImpl.getURL()

zjavaperf:
This is the second of my patches for the customer situation mentioned in pull request #4808. The idea of this patch here is rather simple: I noticed that in the customer's traces, there are a number of expensive, **twofold** `ResourceImpl.getURL()` calls on the _exact_ same `ResourceImpl` instance. Those expensive calls are caused by `ResourceImpl` calling `ClasspathResourceHelper.getURL()`, which in turn has to scan .jar files, etc.

My assumption is that the URL of a `ResourceImpl` instance doesn't change during the lifetime of a request. Note that I'm not proposing to statically cache a `ResourceImpl`'s URL. I only try to avoid performing a _second_ expensive call to `ClasspathResourceHelper.getURL()` inside of a `ResourceImpl` instance's `getURL()` method.

Even with my patch enabled, the URL of a `ResourceImpl` instance is determined every time a resource is accessed for the first time in a request. My patch only avoids the second - and all subsequent - URL resolutions.

Again, I have run every Mojarra integration test that completed successfully without my patches also with my patches enabled: glassfish, javaee6/6web/7/8, servlet30/31/40. The outcome was that all integration tests passed successfully.

Signed-off-by: Marc Beyerle <marc.beyerle@de.ibm.com>

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/resource/ResourceImpl.java
+++ b/impl/src/main/java/com/sun/faces/application/resource/ResourceImpl.java
@@ -101,6 +101,12 @@ public class ResourceImpl extends Resource implements Externalizable {
      */
     private long maxAge;
 
+    /**
+     * The URL of this {@link ResourceImpl} object is valid exactly as long as this {@link ResourceImpl} object itself is valid.
+     * Therefore, resolve the URL only once and re-use the already resolved URL value for subsequent calls to {@link ResourceImpl#getURL()}.
+     */
+    private URL resolvedUrl = null;
+
 
     // ------------------------------------------------------------ Constructors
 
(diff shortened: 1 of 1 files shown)


tandraschko:
JFYI: we do the same in MF and is battle tested there.

melloware (approved):

christophs78:
Probably we should get this into Master (=4.0.) too.

arjantijms:
> Probably we should get this into Master (=4.0.) too.

If you could do a PR for that I'd be more than happy to merge it. Thanks!

PR #4810, opened 2020-11-16
Performance-optimized version of ELUtils.isCompositeComponentExpr()

zjavaperf:
For now, this is the last patch in my series of patches for the customer situation I mentioned in pull request #4808. While analyzing the traces of the customer, I noticed that there are many thousand calls to `ELUtils.isCompositeComponentExpr()` that contribute measurably to the high amount of CPU consumption that the customer is currently experiencing.

When looking at `ELUtils.isCompositeComponentExpr()`, I found that all it basically does is regular expression matching. My proposed patch creates a small - 1000 entries at the moment - cache that stores the regular expression evaluation results for the incoming strings. In order to implement this cache, I used an admittedly very simple LRU cache based on a FIFO queue - see the `rememberEvaluationResult()` method in `ELUtils`.

Instead of implementing a complex _real_ LRU algorithm, which would probably eat away the savings generated by the cache immediately, I only use a `LinkedList` to evict cache entries that hung around for a while. This approach has proven to work well in many other customer situations that I was involved in.

Note that the 1000 entries are not cast in stone. If you like the idea of a cache for `isCompositeComponentExpr()`, we can of course also make the size configurable, for example.

Signed-off-by: Marc Beyerle <marc.beyerle@de.ibm.com>

### DIFF
--- a/impl/src/main/java/com/sun/faces/el/ELUtils.java
+++ b/impl/src/main/java/com/sun/faces/el/ELUtils.java
@@ -32,6 +32,8 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


tandraschko:
@zjavaperf you can even check if the expression string contains a "cc" before using regex. We dont have a cache in MF but the contains("cc") check

tandraschko:
also a static cache could >maybe< lead to bugs in containers where Mojarra is deployed in the container, not in the WAR. Therefore a "app scoped" cache like using the JSF ApplicationMap is a better fit here.

kguelzau:
I had a similar approach for #4082 see here: https://github.com/kguelzau/mojarra/commit/e7949bf645efe3681abba44a858e149dd60f20af#diff-27aacf0c7d38e4f63a3fdc221058109f0bbe598ec8c4aa381595883682df3934R116
But storing the results in TagAttributeImpl instead of using a cache turned out to be much simpler.

christophs78:
Probably we should get this into Master (=4.0.) too.

Issue #4812, opened 2020-11-23
New WARNING error message on undeploy when there are duplicate registrations of ConfigureListener


rmartinc:
Mainly the same issue fixed in #3801 but with a new second message. The new severe message was introduced by the change in #4587 with this commit https://github.com/jboss/mojarra/commit/b45a4e4a7dda21b0a0c8b30d43e249e44a5b1f2a.

This new [line](https://github.com/jboss/mojarra/commit/b45a4e4a7dda21b0a0c8b30d43e249e44a5b1f2a#diff-3f75b18a1a7fd1e67a74dc03b9b8ef4a6987c6429e3211134ed55f60be6737a9R363) displays again the same error not checking if the context was destroyed before.

I'm going to send a PR for 2.3 branch with the same fix that was already done for the previous #3801 but for the new line.

BalusC:
Already in 2.3.15.

Issue #3801, opened 2015-02-26
SEVERE error message on undeploy when there are duplicate registrations of ConfigureListener


ren-zhijun-oracle:
The fix for #3615 causes problems when running on WildFly. In particular, since <tt>ConfigureListener</tt> can be registered twice when deploying a JSF application (duplicate registrations are possible because of a workaround for a GlassFish issue that was added to Mojarra 2.x, as described in #3193), the following error message now occurs when undeploying JSF applications:

```
13:01:50,432 SEVERE [javax.enterprise.resource.webcontainer.jsf.config] (MSC service thread 1-4) Unexpected state during contextDestroyed: no ConfigManager instance in current ServletContext but one is expected to exist.
```

The first time <tt>ConfigureListener.contextDestroyed()</tt> is called, the error doesn't occur, as expected. However, when <tt>ConfigureListener.contextDestroyed()</tt> gets called a second time, <tt>configManager</tt> will evaluate to <tt>null</tt> in <tt>ConfigureListener.contextDestroyed()</tt> since it was already removed from the <tt>ServletContext</tt> the first time <tt>ConfigureListener.contextDestroyed()</tt> was called, resulting in the error message being logged.

The error message will also get logged if <tt>ConfigureListener</tt> gets registered for an app that doesn't actually use JSF (this also happens when running with WildFly). In this case, <tt>configManager</tt> will evaluate to <tt>null</tt> in <tt>ConfigureListener.contextDestroyed()</tt> since it gets [removed](https://github.com/jboss/mojarra/blob/2.2.10-jbossorg-1/jsf-ri/src/main/java/com/sun/faces/config/ConfigureListener.java#L185) from the <tt>ServletContext</tt> when <tt>ConfigureListener.contextInitialized()</tt> is called.

In both cases mentioned, the <tt>ConfigManager</tt> isn't actually expected to exist in the current <tt>ServletContext</tt> when <tt>ConfigureListener.contextDestroyed()</tt> is called so the SEVERE error message shouldn't be logged. The following patch fixes this problem by adding an additional check to determine if the <tt>WebConfiguration</tt> instance is non-null:

```
diff --git a/jsf-ri/src/main/java/com/sun/faces/config/ConfigureListener.java b/jsf-ri/src/main/java/com/sun/faces/config/ConfigureListener.java
index 9bdfc53..96436bd 100644
--- a/jsf-ri/src/main/java/com/sun/faces/config/ConfigureListener.java
[...]

PR #4813, opened 2020-11-23
[Issue 4812] New WARNING error message on undeploy when there are duplicate registrations of ConfigureListener

rmartinc:
Fix for issue #4812 

As commented the same idea applied before for #3801 is used. The message is only displayed if the `WebConfiguration` instance is not null (if it is null then it was removed before by the first listener).

@arjantijms The PR is only sent to the 2.3 branch, if you need it to other branches (3.0 for example) please just let me know.


rmartinc:
Change the name of the PR and the commit. The message now is at WARNING level.

Issue #4800, opened 2020-11-12
Resources: set cache-control-header instead of expires-header


christophs78:
https://github.com/eclipse-ee4j/mojarra/blob/master/impl/src/main/java/com/sun/faces/application/resource/ResourceImpl.java#L184 currently set´s expires-header.

IMO it´s today state-of-the-art to set `cache-control: max-age=XXX` instead of `expires` for resources which should be cached. And set `cache-control: no-cache` for resources which should not be cached.

Details: https://developer.mozilla.org/de/docs/Web/HTTP/Headers/Cache-Control

This would also simplify code a bit.

BalusC:
Already in 2.3.15.

PR #4816, opened 2020-12-08
#4800: Resources: set cache-control-header instead of expires-header (JSF 2.3)

christophs78:
Signed-off-by: Christoph Straßer <christoph.strasser78@gmail.com>

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/resource/ResourceImpl.java
+++ b/impl/src/main/java/com/sun/faces/application/resource/ResourceImpl.java
@@ -188,16 +188,11 @@ public class ResourceImpl extends Resource implements Externalizable {
                 responseHeaders = new HashMap<>(6, 1.0f);
             }
 
-            long expiresTime;
             if (FacesContext.getCurrentInstance().isProjectStage(Development)) {
-                expiresTime = new Date().getTime();
+                responseHeaders.put("Cache-Control", "no-store, must-revalidate");
             } else {
-                expiresTime = new Date().getTime() + maxAge;
+                responseHeaders.put("Cache-Control", "max-age=" + (maxAge/1000));
             }
-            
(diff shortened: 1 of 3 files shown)


christophs78:
Hi @arjantijms!
My first contribution to mojarra...
Due to some reasons Travis CI does not kick in. (even after 6 hours) Travis-job is "received" but does not get queued.
image
Maybe we should think about moving from Travis CI to Github Action. We did this with PrimeFaces (and addon-projects) in october which worked quite well. (https://github.com/primefaces/primefaces/issues/6391)

arjantijms:
Hi Chris, thanks for the contribution! Travis is currently not in the best condition. Most EE4J projects use the Eclipse provided Jenkins for CI though.

@BalusC can you review this one?

impl/src/main/java/com/sun/faces/application/resource/ResourceImpl.java
```diff
@@ -188,16 +188,11 @@ public URL getURL() {
                 responseHeaders = new HashMap<>(6, 1.0f);
             }
 
-            long expiresTime;
             if (FacesContext.getCurrentInstance().isProjectStage(Development)) {
-                expiresTime = new Date().getTime();
+                responseHeaders.put("Cache-Control", "no-cache");
```
BalusC, line 192: Should be `no-store, must-revalidate`.

christophs78: Ok, I changed this.
IMO this is to some degree a matter of taste. Even englisch and german MDN - page for https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control prefer different solutions.

BalusC: Not a matter of taste. If the resource was already (accidentally) cached, `no-cache` doesn't wipe it.

PR #4817, opened 2020-12-08
#4800: Resources: set cache-control-header instead of expires-header (JSF 4.0)

christophs78:
Signed-off-by: Christoph Straßer <christoph.strasser78@gmail.com>

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/resource/ResourceImpl.java
+++ b/impl/src/main/java/com/sun/faces/application/resource/ResourceImpl.java
@@ -172,16 +172,11 @@ public class ResourceImpl extends Resource implements Externalizable {
                 responseHeaders = new HashMap<>(6, 1.0f);
             }
 
-            long expiresTime;
             if (FacesContext.getCurrentInstance().isProjectStage(Development)) {
-                expiresTime = new Date().getTime();
+                responseHeaders.put("Cache-Control", "no-store, must-revalidate");
             } else {
-                expiresTime = new Date().getTime() + maxAge;
+                responseHeaders.put("Cache-Control", "max-age=" + (maxAge/1000));
             }
-
-            SimpleDateFormat format = new SimpleDateFormat(RFC1123_DATE_PATTERN, US);
-            format.setTimeZone(GMT);
-            responseHeaders.put("Expires", format.format(new Date(expiresTime)));
 
             URL url = getURL();
             InputStream in = null;
@@ -195,6 +190,8 @@ public class ResourceImpl extends Resource implements Externalizable {
                 if (lastModified == 0) {
                     lastModified = initialTime;
                 }
+                SimpleDateFormat format = new SimpleDateFormat(RFC1123_DATE_PATTERN, US);
+                format.setTimeZone(GMT);
                 responseHeaders.put("Last-Modified", format.format(new Date(lastModified)));
                 if (lastModified != 0 && contentLength != -1) {
                     responseHeaders.put("ETag", "W/\"" + contentLength + '-' + lastModified + '"');


christophs78:
This PR is lying around for more the 3 months without any feedback. 
Ideally we should port some new features which where build from @tandraschko for MyFaces back to Mojarra to get them into the JSF4-spec. But why should i be motivated when we get no feedback at all for existing PRs?
Sometimes i even ask myself why we split our very limited resources into two JSF implementations.  Shouldn´t we favour one well maintained high-quality implementation?

tandraschko:
CC @arjantijms

arjantijms:
> Sometimes i even ask myself why we split our very limited resources into two JSF implementations. Shouldn´t we favour one well maintained high-quality implementation?

Actually, to be totally honest I've been thinking about the exact same thing. We more or less brushed on this topic when we discussed the TCK recently; that the JSF market has become too small to warrant having anyone put a lot of resources into porting the entire TCK to Facelets/CDI.

I understand it's super frustrating when there's no feedback on the PRs. My last chunk of time I had free for JSF was from late November. Thanks to @manfredriem and @BalusC for stepping in after that.

tandraschko:
Same feeling from my side.
If both Mojarra and MyFaces were one project, without the spec overhead and so on, it would be much easier to maintain and we could progress much easier.

TBH i dont have time to port all my changes from MyFaces to Mojarra, thats more than the double effort for me.

Also if we continue with 1 project, i dont think that we will ever agree on a codebase. Mojarra is a bit cleaner code, but bit more buggy and bit slower. MyFaces is a bit more "hacky" to get the most out of it. In general i reworked the whole codebase in 2.3-next and master/4.0 and its much better now.

i personally could invest good amount of time in JSF API changes and get 4.0 bit more modern and add missing features. But its just killing me that there are so many unmerged PRs in Mojarra, thats just loosing faith in the spec idea.

tandraschko:
another point:
the last active frameworks are PrimeFaces and OmniFaces
its really sad that working on JSF itself is so much overhead compared to PrimeFaces, which envolves great each release cycle.
it would be much easer for the PrimeFaces community to fork e.g. MyFaces and make a "PrimeFaces Core" out of it, with different packages and so on.

tandraschko:
It would be definitely good to have more frequent Faces API releases BUT it dont think it will change our issue, that we have to implement it 2 times.

Another big downside is probably that JSF is delivered inside the EE containers. If it would be just a normal JAR framework, it would be much easier. Even in old containers, you can just deliver the newest Faces + PrimeFaces jar and we dont have hacks anymore in Prime or OmniFaces.
No specs, 1 API, 1 Impl and thats it. Just as PrimeFaces, as external lib.

Also i dont think that merge Faces + PrimeFaces + OmniFaces is a good think for Faces itself. Its actually good that Faces only supports basic HTML, without styles, themes and so on. Thats really related to a UI library, not to a core framework. Just take Angular or Vue. The core isnt about styling.
But i can really think about something like porting all major and cool core features of PrimeFaces and OmniFaces to Faces.

I definitely think that we should not enhance flows. And if we switch Faces e.g. to a own project, we should really think about to make a module out of it. I think core just be as small as possible, which important features.


Indepedent of this exact topic, we could even think about how could a modern Server-Side-Java-Rendering UI framework could look like? How can we reuse e.g. shadow dom, use REST for binding instead of rendering table-content as HTML and so on.
Can we make java developers happy with a more modern approach but allow a similar developing strategy like with Faces? ManagedBeans, ViewScoped and so on? LocalStorageScoped?

christophs78:
From my personal perspective i did my first PrimeFaces PRs back in 2018. Originally i did not plan to do that much more. Just improve/modify a few existing things. But it was kind of rewarding. @tandraschko, @melloware and @Rapster are a good team. They provide good feedback to newbies. And they give you the feeling, you can make a difference and move the project forward.
So i´m still there. :-)
Sometimes you can even improve the project with comparable limited invested time.
Overall we made - as Thomas said - together with the PrimeTek-guys good progress over the last years.

On my day-job we decided in 2017 to use spring instead of JavaEE / JakartaEE. We "only" use things like JPA und JSF from JavaEE / JakartaEE. We really like to be flexible and deploy the versions of Spring, Hibernate, Mojarra, PrimeFaces .... we prefer into most containers. (It was a discussion process because our Ops-people were used to classic WebSphere. But overall it was IMO a good decision for our organisation.)

Having a split between core UI technology (=JSF) and component-libraries (=PrimeFaces) IMO makes sense. Probably there are a few things within PrimeFaces which should be moved into core JSF. (They were added to PrimeFaces because it was the easiest way to get this out of the door.)
Omnifaces is IMO more tightly coupled to core JSF. But i would really miss the excellent OmniFaces-showcase and documentation.
Having two core implementations which we need to maintain in multiple versions does not makes me happy. We need to maintain 2.3 because that´s what people use. We need to maintain 3.0 because it weird to fix things only for 2.3 and not put them into 3.0. And we naturally must put all into master (4.0).
On the PrimeFaces-side of things we need to ensure to be compatible with Mojarra and MyFaces. We introduced https://github.com/primefaces-extensions/primefaces-integration-tests last summer and made good progress. (Some for PrimeFaces Selenium.) Recently we added Mojarra 2.3 to our already extensive test-matrix. (We need to run our tests against Chrome, Firefox, Safari, ....)
Maybe we have some ideas how to solve this "Gordian knot" some time down the road.

I´m currently happy to see some life coming back into core JSF. (Thanks  @arjantijms , @BalusC, @manorrock , @tandraschko )
Maybe i spend the next month´s a bit more time on core JSF and a bit less on PrimeFaces.

The last thing i want to mention for now is tooling support. It´s really weird to loose a far to big part of JSF-support in IntelliJ with the move from javax to jakarta. My organisation moved 40+ dev´s last autumn from Eclipse to IntelliJ. (As a long-term Eclipse user (~15 years) i finally had to agree IntelliJ is far ahead of Eclipse and probably Netbeans.)
Without proper tooling support it´s almost impossible to "sell" a upgrade from e.g. JSF 2.3 to 4.0 in my organisation.

PR #4432, opened 2018-07-27
[#4431] NPE in ELFlash.get() on first request

spyrkob:
Possible fix for https://github.com/eclipse-ee4j/mojarra/issues/4431

PR #4823, opened 2021-01-04
[#4431] NPE in ELFlash.get() on first request

spyrkob:
backport of #4432 for 2.3 branch


manorrock (approved):
As this is a patch without a test I am requesting another reviewer to have a look at it

arjantijms:
Thanks for this! If we ever have time (one can dream), we may wanna try to find out  why `SessionHelper` is null in the first place.

arjantijms (approved):

Issue #4820, opened 2020-12-18
Faces 3.0 Does Not Support JavaEE NameSpace in Flow.xml Files


volosied:
**UPDATE:  Sorry for the numerous comments. It's been a discovery process.**

https://github.com/eclipse-ee4j/mojarra/blob/1c904bfeb193530624f95628a66da029f270bb1d/impl/src/main/java/com/sun/faces/config/processor/FacesFlowDefinitionConfigProcessor.java#L281-L282

`xpath.setNamespaceContext(new FacesConfigNamespaceContext());`
should be changed to handle both https://jakarta.ee/xml/ns/jakartaee and http://xmlns.jcp.org/xml/ns/javaee

FacesConfigNamespaceContext only handles the Jakarta Namespace.

The code later runs into parsing errors since the namespaces don't' match (Using Javaee on the newer Jakartaee 3.0 version) 

____________

I  noticed that the flows misbehave when exiting a flow (through the return button). 

I tested this failing scenario with the 3.0.0 jar bundled in the  [javaee6web/basicFlow app](https://github.com/eclipse-ee4j/mojarra/tree/master/test/javaee6web/basicFlow) on Open Liberty with the facesContainer-3.0 feature. 

Steps to reproduce: 

1. Go to test-javaee6web-basicFlow/faces/index.xhtml 
2. Click "start flow on valid entry node"
3. Click Next
4. Enter value
5. Click Next
6. Click Return 

The return just reloads the page, and doesn't display "Page navigated to upon return from bounded-task-flow" as expected (when using the 2.4.0 release) 

In another application, I noticed the following error:

    Unable to find matching navigation case with from-view-id '/simple/simple-2.xhtml' for action 'goReturn' with outcome 'goReturn' 

Something may be off with the navigation case handling?

volosied:
For the other application I tested, I see that the behavior changed between the RC3 and RC4 releases.

Looks it it narrows it down to these commits: https://github.com/eclipse-ee4j/mojarra/compare/3.0.0-RC3-RELEASE...3.0.0-RC4-RELEASE 

I've add this second application here: https://github.com/volosied/sample-apps/tree/main/flows-failure

The pom uses the RC3 jar but the additional war contains the RC4 jar (I don't see the maven coordinates for RC4) 

Clicking next (exit flow) on /flow-failure-rc4/simple/simple-2.xhtml just refreshes the page when it shouldn't.

volosied:
[...]

PR #4826, opened 2021-01-31
#4820 Faces 3.0 Does Not Support JavaEE NameSpace in Flow.xml Files

hussainnm:
Introduce a constructor to use given namespace instead of default

### DIFF
--- a/impl/src/main/java/com/sun/faces/config/processor/FacesConfigNamespaceContext.java
+++ b/impl/src/main/java/com/sun/faces/config/processor/FacesConfigNamespaceContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2021 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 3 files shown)


manorrock (commented):
Can you add some tests?

hussainnm:
Added Tests

manorrock (commented):
Can you also create a PR against the master branch? Please refer to the issue in its comments.

impl/src/main/java/com/sun/faces/config/processor/FacesConfigNamespaceContext.java
```diff
@@ -23,9 +23,19 @@
 
 class FacesConfigNamespaceContext implements NamespaceContext {
 
+    String defaultNamespace = null;
```
pzygielo, line 26: Could it be private?

arjantijms:
> Can you also create a PR against the master branch? Please refer to the issue in its comments.

We can additionally cherry pick, it's no super big deal, though PR may be more convenient ;)

impl/src/main/java/com/sun/faces/config/processor/FacesConfigNamespaceContext.java
```diff
@@ -23,9 +23,19 @@
...
+
+    public FacesConfigNamespaceContext() {
```
pzygielo, line 28: I think extending accessibility of that constructor (current (existing) default one isn't public) is not necessary. Given type is package-accessible anyway.

manorrock (approved):

PR #4843, opened 2021-03-26
Log throwables using throwable parameter in AjaxExceptionHandlerImpl

### LINKED ISSUE #4839: AjaxExceptionHandlerImpl logs stacktrace as message
I am currently integrating [Sentry application monitoring](https://sentry.io) into a JSF application that is running on Mojarra 2.3. Events for Sentry are captured by integrating a `java.util.logging.Handler` into the logging stack. This handler then takes every logged error event and sends the corresponding message and stacktrace to Sentry where it is processed, for example by grouping stacktraces together.

While monitoring Sentry events for my application I noticed that `com.sun.faces.context.AjaxExceptionHandlerImpl` formats exceptions into the message field of the log entry instead of using the `Throwable` parameter. This messes up grouping in Sentry and probably most other log monitoring software that uses log handlers to capture events.

This behavior was observed in Mojarra 2.3, but seems to be present in `master` as well. The following code formats the stacktrace into the log message:
https://github.com/eclipse-ee4j/mojarra/blob/559f09126f63d955e07548724788bc649e1354f3/impl/src/main/java/com/sun/faces/context/AjaxExceptionHandlerImpl.java#L180-L184

Curiously, just a few lines further down in a `catch` block, another exception is logged "properly" by using the `Throwable` parameter:
https://github.com/eclipse-ee4j/mojarra/blob/559f09126f63d955e07548724788bc649e1354f3/impl/src/main/java/com/sun/faces/context/AjaxExceptionHandlerImpl.java#L188-L191

Is there a reason for formatting the stacktrace into the log message instead of using the `Throwable` parameter? And if this is not the case, can you consider changing it?

RagingCactus:
```
Formatting the stacktrace of throwables into the log message prevents
analysis tools that integrate using log handlers from analyzing the
exception properly. Log using the Throwable parameter instead, which
should look almost the same in classical text logs, but enables
analytics using log handler integrations to work properly.

Fixes #4839
```

---

In #4839 I was instructed to open a PR for this and add tests. However, I don't quite know how I would go about testing logging output.

I am happy to add a test if you can point me to a similar test already in the repo. Otherwise I believe the change is small and obvious enough to be mergeable as-is.

### DIFF
--- a/impl/src/main/java/com/sun/faces/context/AjaxExceptionHandlerImpl.java
+++ b/impl/src/main/java/com/sun/faces/context/AjaxExceptionHandlerImpl.java
@@ -17,8 +17,6 @@
 package com.sun.faces.context;
 
 import java.io.IOException;
-import java.io.PrintWriter;
-import java.io.StringWriter;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedList;
@@ -178,10 +176,7 @@ public class AjaxExceptionHandlerImpl extends ExceptionHandlerWrapper {
             writer.endDocument();
 
             if (LOGGER.isLoggable(Level.SEVERE)) {
-                StringWriter sw = new StringWriter();
-                PrintWriter pw = new PrintWriter(sw);
-                t.printStackTrace(pw);
-                LOGGER.log(Level.SEVERE, sw.toString());
+                LOGGER.log(Level.SEVERE, t.toString(), t);
             }
 
             context.responseComplete();


arjantijms:
Thanks so much! Yes, most if not all logging should be done by passing the exception instance to the logger.

One small thing is that following the JDK 8+ pattern you can pass a lambda to the logger and avoid the if statement around it. But it's clear as is. Thanks again!

PR #4879, opened 2021-03-28
Fix failing f:websocket tests because of HtmlUnit internal issues

BalusC:
Currently used HtmlUnit version appears to be unable to wait until a websocket is opened and also to dislike having more than one websocket open in same session. So tests had to be split. While at it, tests have also been polished.


arjantijms (approved):
Nice!👍

Issue #4881, opened 2021-03-30
Empty lists are discarded during state-saving


christophs78:
.. so this state is not saved and EL is re-evaluated during the next request-cycle.

This causes following issues:
https://github.com/primefaces/primefaces/issues/7026
https://github.com/primefaces/primefaces/issues/7181
https://github.com/primefaces/primefaces/issues/6427

MyFaces preserves emtpy lists.

melloware:
Nice work @christophs78

melloware:
Shouldn't this fix be ported to 3.0 and 4.0 as well?

christophs78:
Yes, we should port it.

BalusC:
Please don't PR/merge this one yet.

The 3.0 and 4.0 are currently not fully up to date with recent 2.3 changes.

Among others https://github.com/eclipse-ee4j/mojarra/commit/57932b88d44a408ed16e7cf6025d5ebe8e55e17e is still not in 3.0/4.0 and overlapping your changes.

I'll sort out it.

BalusC:
Everything is in 2.3 / 3.0 / 4.0 now and I've made sure that 2.3 is synced into 3.0 and then into 4.0.

Thank you for your contribution!

PR #4882, opened 2021-03-30
#4881 Empty lists are discarded during state-saving

christophs78:
Compare with MyFaces BehaviorBase#restoreAttachedState as reference.
https://github.com/apache/myfaces/blob/master/api/src/main/java/jakarta/faces/component/behavior/BehaviorBase.java#L198

PrimeFaces-issues related to this PR:
primefaces/primefaces#7026
primefaces/primefaces#7181
primefaces/primefaces#6427

After this is merged we should do the same for master. (=4.0)

### DIFF
--- a/impl/src/main/java/javax/faces/component/ComponentStateHelper.java
+++ b/impl/src/main/java/javax/faces/component/ComponentStateHelper.java
@@ -114,6 +114,8 @@ class ComponentStateHelper implements StateHelper, TransientStateHelper {
      */
     @Override
     public Object put(Serializable key, String mapKey, Object value) {
+        initMap(key);
+
         if (MARK_CREATED.equals(mapKey)) {
             if (PropertyKeys.attributes.equals(key)) {
                 UIComponent parent = component.getParent();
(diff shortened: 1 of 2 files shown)


christophs78:
@tandraschko Maybe you can add some feedback to this....
MyFaces works with _AttachedListStateWrapper to wrap lists during state-saving. Mojarra does not seem to have something similar.  What had been the reasons to introduce _AttachedListStateWrapper?

tandraschko:
Out of my scope/history

christophs78:
I tried to fix this within the existing Mojarra-codebase. 

Plan "B" would be to do it MyFaces-style. 
Look at BehaviorBase#restoreAttachedState as reference.
https://github.com/apache/myfaces/blob/master/api/src/main/java/jakarta/faces/component/behavior/BehaviorBase.java#L198

tandraschko:
great work @christophs78 

@arjantijms i would be happy if such power users can be invited to be part of the Mojarra team ;)

arjantijms:
@tandraschko absolutely!

@christophs78 if you're up to it, I'd absolutely like to nominate you to become Mojarra committer. What do you think?

arjantijms (approved):

christophs78:
> @tandraschko absolutely!
> 
> @christophs78 if you're up to it, I'd absolutely like to nominate you to become Mojarra committer. What do you think?

Yes, please. We all have limited time budgets. I'm happy trying to find some time for core JSF and contribute to move JSF forward.

melloware (approved):

arjantijms:
@tandraschko I can nominate you as regular committer first, after which I think you can already manage issues and labels and such. Whether you actually commit code or not beyond that is up to you, and of course not a requirement.

tandraschko:
all right, thanks :)

Issue #4811, opened 2020-11-23
RESTORE_PHASE is slow on Mojarra 2.3


wafa-ogh:
Hello All,

After upgrading from Mojarra 2.2.13 to Mojarra 2.3.9, RESTORE_PHASE becomes slow. In my case, the response time of requests from views with at least 5000 components, has increased 12 times compared to 2.2, an ajax action that used to take ~600 to ~800 milliseconds, takes now several seconds to complete :(    
From looking at the differences on the two versions, it was the ComponentSupport.findChildByTagId() method that caused the issue on 2.3.
![image](https://user-images.githubusercontent.com/35977998/99955851-50892e80-2d85-11eb-9c7e-209d5cdc9a59.png)
On 2.2, ComponentSupport.findChildByTagId() does not perform a tree search If the phase Id is equal to RESTORE_PHASE or if the request is not a postback.
The method has undergone 2 changes :

1. The first change removed the check on the phase Id
        The commit description point this bug https://java.net/jira/browse/JAVASERVERFACES-4124 (Ajax file upload fails when 
        com.sun.faces.namespaceParameters is set to true)

2. The second change removed the check on the request type, the related commit description point to these bugs : 
	https://java.net/jira/browse/JAVASERVERFACES-4142 "ENABLE CLIENTSTATESAVINGPASSWORD BY DEFAULT IN JSF 1.2"
	track bugdb bug#23126505: JSF ERRORS AFTER APPLYING WLS 12.1.3 PSU5/PSU6 also fix: Bug 22879626 - REGRESSION IN 
        JSF PAGE RENDERING IN GLASSFISH 3.1.2.12

Although the fix could be to stay on the 2.2.13 version right now, but as Mojarra user and interested on knowing about the updates, I wanted to better understand these changes in the 2.3.9 version : Are they really related to changes for "com.sun.faces.ClientStateSavingPassword" to be enabled by default (4142) and the "Ajax issue when com.sun.faces.namespaceParameters is set to true" (4124) or did I misunderstood the commits comments ?
Also, will the ComponentSupport.findChildByTagId() stay as is on next releases especially because it has impacted Mojarra performance when handling big component trees  ? Could this be considered a bug on 2.3 ?

Environment : JSF 2.3.9 SP06 / Wildfly 19.0.0

BalusC:
[...]

PR #4905, opened 2021-04-06
Move logic from UIComponent to ComponentSupport

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/4811


christophs78:
This PR essentially (from a performance-point-of-view) replaces `getDescendantMarkIdCache()` with `component.getTransientStateHelper().getTransient(MARK_ID_CACHE)`?
So we should not expect an significant performance-degradation due to this PR. 
https://github.com/eclipse-ee4j/mojarra/blob/master/impl/src/main/java/jakarta/faces/component/ComponentStateHelper.java#L490 is plain simple and has no costly logic. So from my side this looks good!

christophs78 (approved):

PR #4653, opened 2019-10-31
[Issue 4652] Memory leak in FlashScope when response is already commi…

spyrkob:
…tted

Issue: https://github.com/eclipse-ee4j/mojarra/issues/4652

PR #4906, opened 2021-04-07
[Issue 4652] Memory leak in FlashScope when response is already commi…

spyrkob:
…tted

Backport of #4653 for 2.3 branch

Signed-off-by: Bartosz Spyrko-Smietanko <spyrkob@gmail.com>


BalusC:
Yup confirmed present in 4.0 but absent in both 2.3 and 3.0. This PR is only vor 2.3 but I'll make sure it also ends up in 3.0.

BalusC:
It's in 2.3/3.0/4.0 now. Thank you for your contribution!

PR #4922, opened 2021-05-03
Revert #4808 due to being possible cause of failing TCK

BalusC:
Ran TCK against this branch, it passed successfully, so it was indeed the cause.

### DIFF
--- a/impl/src/main/java/com/sun/faces/facelets/tag/jsf/ComponentSupport.java
+++ b/impl/src/main/java/com/sun/faces/facelets/tag/jsf/ComponentSupport.java
@@ -16,28 +16,10 @@
 
 package com.sun.faces.facelets.tag.jsf;
 
-import com.sun.faces.RIConstants;
-import com.sun.faces.context.StateContext;
-import com.sun.faces.facelets.tag.jsf.core.FacetHandler;
 import static com.sun.faces.config.WebConfiguration.BooleanWebContextInitParameter.PartialStateSaving;
-import com.sun.faces.util.Util;
-
-import javax.faces.FacesException;
-import javax.faces.component.UIComponent;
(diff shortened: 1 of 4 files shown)


arjantijms (approved):

christophs78:
That's weird. Our release-build failed today too.
https://travis-ci.org/github/eclipse-ee4j/mojarra/builds/769310675?utm_source=github_status&utm_medium=notification
Is there a difference between CI during "normal" commits and CI during release-builds (triggered by setting a version-tag??).

zjavaperf:
Could the reason be as simple as one or more missing or malformed `pom.xml` files? See the job log:

```
[WARNING] The POM for org.glassfish:jakarta.faces:jar:2.3.15-SNAPSHOT is missing, no dependency information available
...
[ERROR] Failed to execute goal on project absoluteOrdering: Could not resolve dependencies for project com.sun.faces.test.servlet30:absoluteOrdering:war:2.3.15-SNAPSHOT: Could not find artifact org.glassfish:jakarta.faces:jar:2.3.15-SNAPSHOT in sonatype-snapshots (https://oss.sonatype.org/content/repositories/snapshots/) -> [Help 1]
```

From what I remember when running the Mojarra integration test suite locally on my PC, I had to manually "massage" the `pom.xml` files underneath of the `test` subdirectory because of the `-SNAPSHOT` suffix in the version string.

Just scanned my notes and this is what I had to do in the `test` subdirectory:

`find . -name "pom.xml" | xargs -i sed -i s/2.3.14-SNAPSHOT/2.3.14/g '{}'`

At the time, I didn't think anything of it, but perhaps this is either the root cause of the failed tests or at least related to it. Would love to hear your thoughts on this!

Issue #4924, opened 2021-05-04
Add back Performance-optimized version of ComponentSupport.findChildByTagId()


BalusC:
#4808 has been reverted for 2.3.15 as per #4922

This needs to be added back but first we need to identify the failing TCK test and replicate the test in Mojarra project.

@arjantijms can you do this? Then I can take a second look at the impl.

arjantijms:
Yes, I'll look at adding the test in 2.3 first, and then move it to 3.0 and 4.0 as well.

BalusC:
Merged and upmerged in 3.0 and 4.0 as well.

https://github.com/eclipse-ee4j/mojarra/commit/77e0e04a3027c73bc62383a2ac3cde3b86fe3f32
https://github.com/eclipse-ee4j/mojarra/commit/7ebc063432f58f8a5fb0d9fea565f46ed44a29b5

PR #4931, opened 2021-05-16
Restore #4808 and add work around for failing TCK test

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/4924

### DIFF
--- a/impl/src/main/java/com/sun/faces/facelets/tag/jsf/ComponentSupport.java
+++ b/impl/src/main/java/com/sun/faces/facelets/tag/jsf/ComponentSupport.java
@@ -16,10 +16,28 @@
 
 package com.sun.faces.facelets.tag.jsf;
 
+import com.sun.faces.RIConstants;
+import com.sun.faces.context.StateContext;
+import com.sun.faces.facelets.tag.jsf.core.FacetHandler;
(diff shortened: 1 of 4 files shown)


melloware (approved):

zjavaperf:
Hi @BalusC,

just stumbled across this pull request. Seems like you actually found the root cause of the failed TCK tests, meaning the `null` `FacesContext` in case the tests are running under the TCK - awesome :+1: I assumed that it must have been something simple like this, but you nailed it. Thank you very much for your time and the effort you put into this.

Cheers,
Marc

BalusC:
Unfortunately the test is still failing. It turns out that `FacesContext#getCurrentInstance()` was not null during the test, but it was `FacesContext#getCurrentPhaseId()` which unexpectedly returned `null` instead of some sort of default such as `ANY_PHASE`. But this in turn highlighted another problem, an `Object#equals()` test was performed to compare the current phase ID while it is actually a constant. The `==` should have been used instead. It's not only compiletime-typesafe but also nullsafe.

I'll have to redo the restore in a new branch.

BalusC:
https://github.com/eclipse-ee4j/mojarra/pull/4934

PR #4943, opened 2021-06-11
CDI - make sure current state works with CDI 4.

manovotn:
I was testing WFLY with latest CDI API (4.0.0.Alpha1) where there was a number of removals of long deprecated methods.
See eclipse-ee4j/cdi#472

This causes mojarra to crash during TCK runs so I had to create a custom artifact and replace it within WFLY - this PR a sum of changes I needed to make it work.

Since you are bound to eventually bounce into the very same issue, I thought I might as well contribute. But whether you accept it or not is OFC completely up to you :-)

Oh and I should note - the current state of this PR will work with CDI 2/3 as well as CDI 4 so there should be no harm in having it applied right away.

### DIFF
--- a/impl/src/main/java/com/sun/faces/cdi/CdiProducer.java
+++ b/impl/src/main/java/com/sun/faces/cdi/CdiProducer.java
@@ -148,12 +148,7 @@ abstract class CdiProducer<T> implements Bean<T>, PassivationCapable, Serializab
         return false;
     }
 
-    /**
-     * Is this nullable.
-     *
-     * @return false.
-     */
-    @Override
+    // TODO to be removed once using CDI API 4.x
     public boolean isNullable() {
         return false;
     }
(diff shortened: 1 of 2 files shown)


melloware (approved):

manorrock (commented):
Can you explain what the getEvent() method does in this context?

arjantijms:
Thanks @manovotn! Perhaps before long we should start with a build of GF 7 which incorporates as many alpha or milestones APIs and implementations as we have available.

manovotn:
> Can you explain what the getEvent() method does in this context?

It gives you an `Event<Object>` from where you can specify more concrete type and qualifiers if needed and fire either sync or async events. Basically the same as if you did `@Inject Event<Object>` just using `BeanManager` instead.

manorrock:
@arjantijms @BalusC Can either of you review/merge this?

PR #4962, opened 2021-09-08
#1581 Implement spec issue 1581 - Programmatic Facelets instances

arjantijms:
Minimalistic implementation for https://github.com/eclipse-ee4j/faces-api/issues/1581


tandraschko:
Looks good for me, i just wonder about:

1) FacesViewId is a mapping to a view, just like ViewActionMapping will be
I think we should align both namings. Sorry, im a bit fussy regarding namings.

2) AFAICS (i currently review via phone), a "managed facelet" can also be used in ui:include. We have to choose a different naming then, relates to 1)

3) ResourceDependency handling will only work with #createComponent instead of 'new HtmlForm', i think samples should use it

impl/src/main/java/jakarta/faces/annotation/FacesViewId.java
```diff
@@ -0,0 +1,60 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import jakarta.enterprise.util.AnnotationLiteral;
+import jakarta.inject.Qualifier;
+
+/**
+
+ *
```
BalusC, line 17: Javadoc missing?

test2/faces40/javaPage/src/main/java/com/sun/faces/test/faces40/javapages/Hello.java
```diff
@@ -0,0 +1,62 @@
...
+import jakarta.faces.context.FacesContext;
+import jakarta.faces.view.facelets.Facelet;
+
+@FacesViewId("/hello.xhtml")
+@ApplicationScoped
+public class Hello extends Facelet {
+
+    @Override
+    public void apply(FacesContext facesContext, UIComponent root) throws IOException {
+        if (!facesContext.getAttributes().containsKey(IS_BUILDING_INITIAL_STATE)) {
```
BalusC, line 25: Shouldn't this be `ComponentHandler.isNew(root)`?

arjantijms: Nope, that one only checks if the component has a parent. This here has to do with the fact the Facelets compiler actually builds the view twice, and then the second time it deletes and adds components.  The difference between the first and second time is the `IS_BUILDING_INITIAL_STATE` constant being set.

At some point we may need to look at this Facelets internal weirdness, but that's far beyond the scope of this issue.

impl/src/main/java/jakarta/faces/annotation/FacesViewId.java
arjantijms: Indeed, I didn't add Javadoc or spec text yet. I'll do that after feedback.

arjantijms:
@tandraschko 
> FacesViewId is a mapping to a view, just like ViewActionMapping will be
I think we should align both namings. Sorry, im a bit fussy regarding namings.

Don't be sorry, I'm always quite fussy about it too ;)

Here, `FacesViewId` is a qualifier for whatever "bean query" needs to be queried by view id. 

In this case the query is for type `Facelet`, but not just any facelet type, but those that handle view id X, where X is the value of  `FacesViewId` annotation.

For the `ViewAction` to be used in this way, we could work with 2 annotations then:

```java
    @RequestScoped
    @ViewAction
    public class IndexPageBean implements Serializable {
 
      /**
       * Execute the page.
       * 
       * @return /index.xhtml
       */
      @FacesViewId("/index")
      public String execute() {
        return "/index.xhtml";
      }
    }
```

WDYT?

(p.s it would be nice if the CDI bean queries allowed to select on method types and/or method annotations. I'll pitch this on the CDI list)

tandraschko:
I will think a bit about it and come back in weekend, i already have something in mind

tandraschko:
do we want to support wildcard stuff in the `@ViewActionMapping`? Otherwise we could reuse the same qualifier

arjantijms:
Maybe we want so support wildcard stuff indeed. We can simply specify that the `value` of `@FacesViewId` can hold a pattern as well?

BalusC (approved):

test2/faces40/javaPage/src/main/java/com/sun/faces/test/faces40/javapages/Hello.java
BalusC: Ok. Agreed

tandraschko:
all right
should we stay with the "Faces" prefix of the annotation? I think it was only introduced for Converter/Validator because of naming conflicts

I would then maybe just suggest "View" or "FacesView" as annotation name as "Id" sounds like not usable with wildcards or else later

For the "ViewAction" mapping it could then be used like this:

```
public void myMethod(@Observes @View("index.xhtml") ViewActionEvent event)
```

this is what i would like to see for other events, too like implemented in DS

arjantijms:
> should we stay with the "Faces" prefix of the annotation?

Not necessarily indeed. Sometimes its handy to be more explicit with names, sometimes it's better to be short. For parameter annotations short is typically preferred.

tandraschko:
then lets go with `@View`?

arjantijms:
> then lets go with @View?

Okay, I like the idea. Let me update the PR then.

impl/src/main/java/jakarta/faces/annotation/FacesViewId.java
arjantijms: Done

impl/src/main/java/jakarta/faces/view/facelets/Facelet.java
```diff
@@ -32,6 +32,27 @@
  */
 public abstract class Facelet {
 
+    /**
+     * <p>
+     * <span class="changed_modified_4_0">The</span> passed UIComponent parent will be populated/restored in accordance with
```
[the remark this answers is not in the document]
arjantijms: Agree, and done

tandraschko:
Looks good to me now :)

I just wonder about the package. Should it go into view or annotation package?
We dont have a rule for it now

arjantijms:
>  Should it go into view or annotation package?
> We dont have a rule for it now

During the last spec cycle someone wanted all annotations in a separate annotation package. As far as I can remember nobody else really objected, so then it became the annotation package. We might want to set some rules for those things indeed.

tandraschko:
Ahh, OK
I couldnt remember
+1 

I will also implement it in MF tomorrow

arjantijms:
Thanks! :)

PR #4970, opened 2021-09-12
Spec 1508 - Implement automatic extensionless mapping

arjantijms:
Mojarra implementation of https://github.com/eclipse-ee4j/faces-api/issues/1508

### DIFF
--- a/test2/faces40/pom.xml
+++ b/test2/faces40/pom.xml
@@ -34,7 +34,6 @@
     <modules>
 
     	  <module>ajax</module>
-        <module>cdi</module>
         <module>extensionless-mapping</module>
         
         <module>doctype</module>
--- a/test2/faces40/pom.xml
+++ b/test2/faces40/pom.xml
@@ -37,7 +37,6 @@
         <module>cdi</module>
         <module>extensionless-mapping</module>
         <module>javaPage</module>
-        <module>javaPageWithMetadata</module>
         
         <module>doctype</module>
 
--- a/test2/faces40/pom.xml
+++ b/test2/faces40/pom.xml
@@ -38,7 +38,6 @@
         <module>extensionless-mapping</module>
         <module>javaPage</module>
         <module>javaPageWithMetadata</module>
-        <module>selectItemGroups</module>
         
         <module>doctype</module>
 
--- a/test2/faces40/pom.xml
+++ b/test2/faces40/pom.xml
@@ -36,7 +36,6 @@
     	  <module>ajax</module>
         <module>cdi</module>
         <module>extensionless-mapping</module>
-        <module>javaPage</module>
         
         <module>doctype</module>
 
--- a/impl/src/main/java/com/sun/faces/RIConstants.java
+++ b/impl/src/main/java/com/sun/faces/RIConstants.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 1997, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2021 Contributors to Eclipse Foundation.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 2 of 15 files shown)


tandraschko:
without a deep review: is it the same feature like implemented in MF?

if yes, what happens when browse e.g. "/index" - how are e.g. links rendered? "index" or "index.xhtml"?
AFAIR MF renders index.xhtml, which isnt nice

arjantijms:
> without a deep review: is it the same feature like implemented in MF?

Yes, and the same as blogged about here: 

https://arjan-tijms.omnifaces.org/2017/08/extensionless-urls-with-jsf-23.html

You are right about the links. I did a more advanced implementation for OmniFaces that does take care of the links. See https://showcase.omnifaces.org/facesviews/ExtensionlessURLs

Since Faces core is always about minimalism I didn't include that, but if there's demand (I think there is now) I can include that part as well.

tandraschko:
That would be great to include it now, otherwise it feels a bit broken

arjantijms:
> That would be great to include it now, otherwise it feels a bit broken

Okay, sure, I'll look at including that as well then.

arjantijms:
@tandraschko This appeared to be investing, I finally got back to this topic, and actually it already works in Mojarra out of the box.

With extensionless mapping, a source such as:

```xml
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:h="http://xmlns.jcp.org/jsf/html">
      
  <h:head>
    <title>Test JSF 4.0 Extensionless</title>
  </h:head>
  
  <h:body>
    This is page foo <br/>
    <h:link outcome="bar.xhtml" value="go to bar"/>
  </h:body>
</html>
```

renders to:

```xml
<html xmlns="http://www.w3.org/1999/xhtml">
  <head id="j_idt3">
    <title>
      Test JSF 4.0 Extensionless
    </title>
  </head>
  <body>
    
    This is page foo 
    <br/>
    <a href="/test-faces40-ajax/bar">
      go to bar
    </a>
  </body>
</html>
```

If the link uses `bar` instead of `bar.xhtml` it also works.

It's because of the support for exact mapping that was added, and which I refactored a while ago:

```java
        // ### Deal with exact mapping

        if (mapping.getMappingMatch() == EXACT) {
            if (viewId.contains(".")) {
                for (String extension : configuredExtensions) {
                    if (viewId.endsWith(extension)) {
                        String exactViewId = viewId.substring(0, viewId.lastIndexOf(extension));
                        if (isViewIdExactMappedToFacesServlet(exactViewId)) {
                            return contextPath + exactViewId;
                        }
                    }
                }
            } else {
                if (isViewIdExactMappedToFacesServlet(viewId)) {
                    return contextPath + viewId;
                }
            }

            // No exact mapping for the requested view id, see if Facelets service is mapped to
            // e.g. /faces/* or *.xhtml and take that mapping
            mapping = getFirstWildCardMappingToFacesServlet(context.getExternalContext());

            if (mapping == null) {

                // If there are only exact mappings and the view is not exact mapped,
                // we can't serve this view

                throw new IllegalStateException("No suitable mapping for FacesServlet found. To serve views that are not exact mapped "
                        + "FacesServlet should have at least one prefix or suffix mapping.");
            }
        }
```

Like always in Faces, it uses the mapping of the current request to calculate the mapping of the target. In this example, it will see `bar.xhtml`, loop over the configured extensions (by default here only .xhtml), so we get `bar`. It will then check if `bar` is exact mapped. Since automatic extensionless mapping maps all views reported by VDLs, this will always be true, so it will return the context path + `bar` as the action URL..

So by default the current request is leading for the kind of mapping that us used for the target. This makes a small and subtle difference when multiple mappings are active.

tandraschko:
cool, thats great
so +1 from my side, i will add it in MF next week

test2/faces40/pom.xml
```diff
@@ -32,8 +32,16 @@
     <name>Mojarra ${project.version} - Test - Faces 4.0</name>
 
     <modules>
-        <module>ajax</module>
+
+    	  <module>ajax</module>
+        <module>cdi</module>
```
pzygielo, line 37:
```suggestion
```

test2/faces40/pom.xml
```diff
@@ -32,8 +32,16 @@
...
+        <module>extensionless-mapping</module>
+        <module>javaPage</module>
```
pzygielo, line 39:
```suggestion
```

test2/faces40/pom.xml
```diff
@@ -32,8 +32,16 @@
...
+        <module>javaPageWithMetadata</module>
```
pzygielo, line 40:
```suggestion
```

test2/faces40/pom.xml
```diff
@@ -32,8 +32,16 @@
...
+        <module>selectItemGroups</module>
```
pzygielo, line 41:
```suggestion
```

arjantijms: Thanks, there was a merge error there. I already fixed it locally as obviously the tests completely failed. But I can just as well accept those changes. Thanks again!

arjantijms:
Implemented https://github.com/eclipse-ee4j/faces-api/issues/1508 on the Mojarra side.

Issue #4642, opened 2019-10-07
Add BeanManager lookup fallback to CDI.current


tandraschko:
See: https://issues.apache.org/jira/browse/OWB-1298

Currently Mojarra only try to lookup the BeanManager via JDNI (Util#getCdiBeanManager).
But this requires some configration of jetty/tomcat and web.xml, which is annoying for the users.

It would be great if you could try CDI.current instead of returning null. We do the same in MyFaces

arjantijms:
CDI.current first might be problematic because of e.g. this report:

https://issues.jboss.org/browse/WELD-2510

See also this one:

https://github.com/eclipse-ee4j/soteria/pull/243

Given that, it might be an option to try CDI.current *last*. 

The idea would be that in runtimes that support EARs, JNDI is basically always supported as well. Jetty/Tomcat don't support EARs, so they would fall back to the problematic CDI.current, which however won't be problematic to them.

What do you think?

tandraschko:
yep, why not
it doesn't matter if first or last, the important thing is to avoid NPEs for users :)

arjantijms:
Yes, avoiding that is important, although avoiding unpredictable behaviour is also important ;)

I'm not entirely sure what to think of CDI.current and where its current problems must be fixed. The Weld team said it's an integration issue, so that would put the blame on GlassFish etc, but I'm not 100% sure.

rmannibucau:
Using CDI.current() or a JNDI lookup should end with the same behavior if we want applications to be portables (it is the only defined way to get a contextual instance) - whatever the integration handles it or the CDI core impl is an implementation details so it can be the full server integration or weld, it is not important.
That said, using CDI.current() as a last fallback does not hurt in all cases (it is better than failling or returning null).
So maybe just reverse the "first" of the issue ;)

christophs78:
You may assign this issue to me. PR propably comming within a week.

christophs78:
This one may be closed because already implemented by @BalusC with https://github.com/eclipse-ee4j/mojarra/commit/98189bea9d485c8bf459718eae1e127b063c4fad#diff-6e59848a7dc34c64f4c25e5afc623099e1262d1161223a33eec9570569242aeb.
Is it an option to release 2.3.17?

BalusC:
Indeed already implemented in 2.3.17.

PR #4974, opened 2021-09-24
#4642 Add BeanManager lookup fallback to CDI.current

christophs78:


christophs78:
Migrated some code from MyFaces (https://github.com/apache/myfaces/blob/114fd7af5c42865c48ef0d98e185dd102e2b9395/impl/src/main/java/org/apache/myfaces/webapp/FacesInitializerImpl.java#L534) to Mojarra.

Can you merge this PR (targeted for 2.3) also into 3.0 and 4.0 or should i provide separate PR´s?

Not sure whether we still need the WELD-special-treatment:
`result = (BeanManager) applicationMap.get("org.jboss.weld.environment.servlet.javax.enterprise.inject.spi.BeanManager");`

IMO MyFaces does the JNDI-lookup a bit more structured. When you wish, i may port this part of to Mojarra too.

christophs78:
Hm. We may do this without of reflections because JSF since 2.3 requires CDI.

melloware (approved):

tandraschko:
at least you can do it without reflection in 4.0 but 2.3 may also be used with CDI 1.0

BalusC:
But Mojarra is already doing that? https://github.com/eclipse-ee4j/mojarra/blob/8eb32a83c40c3666fa568a32ca450bbf1b63875d/impl/src/main/java/com/sun/faces/util/Util.java#L1572

christophs78:
OK. Fixed but not released. Missed this detail.

tandraschko:
Also fixed in 2.3 or only newer?

christophs78:
Also fixed for 2.3.
(Analysed this issue with 2.3.16 and had tomatoes on my eyes when i did this PR.)

PR #4978, opened 2021-10-15
Fix #4977: Flash: does not work in scenarios with rendered response

### LINKED ISSUE #4977: Flash: does not work in scenarios with rendered response
When we try to pass Java-objects via Flash to a PrimeFaces Dialog Framework (which is technically a IFRAME; http://primefaces.org/showcase/ui/df/data.xhtml) this does not work due to 
`faces.externalcontext.flash.response.already.committed=JSF1095: The response was already committed by the time we tried to set the outgoing cookie for the flash.  Any values stored to the flash will not be available on the next request.`

Debugging (with Mojarra 2.3.16) show´s following:

`FaceletViewHandlingStrategy#renderView`  calls via some intermediate steps `HtmlResponseWriter#endDocument`. And `HtmlResponseWriter#endDocument` calls `flush`.
Mojarra_Flash_01_Flush

After this `FaceletViewHandlingStrategy#renderView`  calls `ctx.getExternalContext().getFlash().doPostPhaseActions(ctx);` which want´s to set the Flash-cookie. This can´t work because of the flush before.
Mojarra_Flash_02_Flush

Ironically this works when you use https://showcase.omnifaces.org/filters/GzipResponseFilter because this filter kind of ignores flush. (We used GzipResponseFilter up to now. After switching to GZIP provided by the container our existing usecases broke.)

(MyFaces does not have this issue.)

And i´ll try to provide a PR. So you may assign this issue to me.

christophs78:


melloware:
@BalusC should review this one but it looks good to me!

arjantijms (approved):

arjantijms:
LGTM, @BalusC I'll merge this now, but if you could still review it after the merge, the code can be adjusted then if needed.

christophs78:
Can you please merge 2.3-branch into 3.0 and master (4.0)? (So we get this modifications also into the newer branches?)

impl/src/main/java/com/sun/faces/application/view/FaceletViewHandlingStrategy.java
```diff
@@ -439,14 +440,18 @@ public void renderView(FacesContext ctx, UIViewRoot viewToRender) throws IOExcep
...
                 viewToRender.encodeAll(ctx);
+                contextMap.remove(ELFlash.DELAYED_END_DOCUMENT);
                 try {
                     ctx.getExternalContext().getFlash().doPostPhaseActions(ctx);
                 } catch (UnsupportedOperationException uoe) {
                     if (LOGGER.isLoggable(FINE)) {
                         LOGGER.fine("ExternalContext.getFlash() throw UnsupportedOperationException -> Flash unavailable");
                     }
                 }
+                ctx.getPartialViewContext().getPartialResponseWriter().endDocument();
```
BalusC, line 454: This does not satisfy the existing comment on line 441. One of them should be adjusted.

impl/src/main/java/com/sun/faces/application/view/FaceletViewHandlingStrategy.java
```diff
@@ -439,14 +440,18 @@ public void renderView(FacesContext ctx, UIViewRoot viewToRender) throws IOExcep
 
             //  Don't call startDoc and endDoc on a partial response
             if (ctx.getPartialViewContext().isPartialRequest()) {
+                Map<Object, Object> contextMap = ctx.getAttributes();
+                contextMap.put(ELFlash.DELAYED_END_DOCUMENT, true);
                 viewToRender.encodeAll(ctx);
+                contextMap.remove(ELFlash.DELAYED_END_DOCUMENT);
```
BalusC, line 446: Shouldn't this also be removed when encodeAll throws an exception?

impl/src/main/java/com/sun/faces/context/PartialViewContextImpl.java
```diff
@@ -322,7 +323,9 @@ public void processPartial(PhaseId phaseId) {
                 renderState(ctx);
                 renderEvalScripts(ctx);
 
-                writer.endDocument();
+                if (!ctx.getAttributes().containsKey(ELFlash.DELAYED_END_DOCUMENT) || !(Boolean) ctx.getAttributes().get(ELFlash.DELAYED_END_DOCUMENT)) {
```
BalusC, line 326: `if (ctx.getAttributes().get(ELFlash.DELAYED_END_DOCUMENT) == Boolean.TRUE)` will make the `containsKey()` call superfluous.

BalusC:
I looked at remaining of PartialViewContextImpl code and saw this:
```
                if (isRenderAll()) {
                    renderAll(ctx, viewRoot);
                    renderState(ctx);
                    writer.endDocument();
                    return;
                }
```
I guess the flash cookie should also be written before that moment, else the same bug may reoccur when `render="@all"` is used. 

On a second thought I think it's probably better to just move the whole try-block which calls `ctx.getExternalContext().getFlash().doPostPhaseActions(ctx);` from FaceletViewHandlingStrategy to PartialViewContextImpl.

PR #4982, opened 2021-10-21
Fix #4979: MenuRenderer#createCollection fails when collection is an instance of java.util.Arrays$ArrayList

### LINKED ISSUE #4979: MenuRenderer#createCollection fails when collection is an instance of java.util.Arrays$ArrayList
Happens when the ArrayList was created via `Arrays.asList()`.

Maybe we should fall back in https://github.com/eclipse-ee4j/mojarra/blob/8a2124c7a5c66642f898ff32924887f55990bc5b/impl/src/main/java/com/sun/faces/renderkit/html_basic/MenuRenderer.java#L698-L702 to fallBackType.

Or map `java.util.Arrays$ArrayList` to `java.util.ArrayList`.

Or???

I need to do some more analysis.

christophs78:
Not sure this is the best way to solve this issue, but at least it works.
MyFaces (i did not debug this aspect in detail) seems to avoid this issue at all.


arjantijms:
That's a good question. You'd think there should be a better solution, but let's keep that in mind then for the future.

PR #4983, opened 2021-11-02
JSF 2.3: Query param values containing equals sign not correctly parsed in NavigationHandlerImpl findImplicitMatch

kmix:
URI query parameters are allowed to contain non-escaped equals signs. The existing method that splits the values between the ampersands of the query fragment defaults to a limit of 0, which will result in a string array of unlimited length:

From [NavigationHandlerImpl.java](https://github.com/eclipse-ee4j/mojarra/blob/master/impl/src/main/java/com/sun/faces/application/NavigationHandlerImpl.java):

`String[] elements = Util.split(appMap, queryElements[i], "=");`

and from [Util.java](https://github.com/eclipse-ee4j/mojarra/blob/master/impl/src/main/java/com/sun/faces/util/Util.java):

```
public synchronized static String[] split(Map<String, Object> appMap, String toSplit, String regex) {
        Map<String, Pattern> patternCache = getPatternCache(appMap);
        Pattern pattern = patternCache.get(regex);
        if (pattern == null) {
            pattern = Pattern.compile(regex);
            patternCache.put(regex, pattern);
        }
        return pattern.split(toSplit, 0);
    }
```

Additionally, if using key=value convention in the query parameters and the value portion is a padded hash or similar that ends in one or more equals signs, the trailing equals signs are silently stripped from the end of the value (e.g. "Zm9vQmFyVmFsdWU=" becomes "Zm9vQmFyVmFsdWU"). While the missing padding won't always be an issue, it can be if the application is expecting to exactly match a string value.

A fix for this issue would be to only split on the first equals sign in the parameter, leaving any other equals signs untouched in the value. This PR overloads the existing Util.split method with an optional splitLimit instead of the current default of 0, thought it leaves the existing Util.split signature in place with a limit of 0.

This PR also includes a minor typo fix in the Util class - `PATTERN_CACKE_KEY` renamed to `PATTERN_CACHE_KEY`.

My particular use case is with 2.3, but if this PR is acceptable I'm happy to create PRs for master and 3.0, as well.

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/NavigationHandlerImpl.java
+++ b/impl/src/main/java/com/sun/faces/application/NavigationHandlerImpl.java
@@ -875,7 +875,7 @@ public class NavigationHandlerImpl extends ConfigurableNavigationHandler {
 
                 String[] queryElements = Util.split(appMap, queryString, "&amp;|&");
                 for (int i = 0, len = queryElements.length; i < len; i++) {
-                    String[] elements = Util.split(appMap, queryElements[i], "=");
+                    String[] elements = Util.split(appMap, queryElements[i], "=", 2);
                     if (elements.length == 2) {
                         String rightHandSide = elements[1];
                         String sanitized = null != rightHandSide && 2 < rightHandSide.length() ? rightHandSide.trim() : "";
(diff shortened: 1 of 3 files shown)


melloware (approved):

codylerum:
@arjantijms Would be nice to have this fixed for the 4.0 release. Should a PR be created against master as well?

arjantijms:
> Would be nice to have this fixed for the 4.0 release. Should a PR be created against master as well?

Yes, that would indeed be a good idea.

arjantijms (approved):

PR #5016, opened 2021-12-09
Cleanup: the @HandlesTypes of FacesInitializer does not anymore contain only annotation classes

BalusC:


BalusC:
Invalid approach, will improve later.

PR #5024, opened 2021-12-14
Cleanup document generation

BalusC:
Corresponding PR in faces https://github.com/jakartaee/faces/pull/1597


impl/src/main/java/com/sun/faces/facelets/tag/faces/html/HtmlLibrary.java
```diff
@@ -35,77 +67,77 @@
...
 
-        addHtmlComponent("selectOneMenu", "jakarta.faces.HtmlSelectOneMenu", "jakarta.faces.Menu");
+        addHtmlComponent("selectOneMenu", HtmlSelectOneMenu.COMPONENT_TYPE, "jakarta.faces.Menu");
 
-        addHtmlComponent("selectOneRadio", "jakarta.faces.HtmlSelectOneRadio", "jakarta.faces.Radio");
+        addHtmlComponent("selectOneRadio", HtmlSelectOneRadio.COMPONENT_TYPE, "jakarta.faces.Radio");
 
-        addHtmlComponent("title", "jakarta.faces.Output", "jakarta.faces.Title");
+        addHtmlComponent("title", UIOutput.COMPONENT_TYPE, "jakarta.faces.Title");
     }
```
BalusC, line 141: We should probably introduce new constants like `HtmlSelectOneRadio.DEFAULT_RENDERER_TYPE` instead of hardcoding those renderer types as well.

Also I'm wondering where those for "html" and "title" are coming from? They do actually not exist as components ...

arjantijms: Especially for the new programmatic facelets, maybe at least the HTML one should exist.

arjantijms (approved):

Issue #5036, opened 2022-01-17
Follow up to XML update parsing error because of the 0x0C (Form Feed) control character


j-be:
As #4145 was closed, and as requested there, I open a new bug report, as the issue is still reproducible with WildFly 26.0.0.Final.

I'm not going to reiterate the full bug (see #4145 for that). I will however show how to reproduce it.

* Deploy the `war`contained in [mojarra4145.zip](https://github.com/eclipse-ee4j/mojarra/files/7881074/mojarra4145.zip) to WildFly 26.0.0.Final
* Open a browser on http://localhost:8080/mojarra4145/
  * Hint: Open your browser's developer tools now
* Switch to tab "Invalid Char Demo"
* Add an invalid entry using the "Add invalid entry" button
  * Look at the browser's developer tools - you should already see some error regarding invalid XML
  * Notice how no entry appears in the list
* Open a second browser window, open http://localhost:8080/mojarra4145/, and go to tab "Invalid Char Demo"
  * Notice how the tab is completely empty
* Switch back to "Dummy" tab and click "Remove invalid entry"
* Everything is back to normal now in both browser windows

The source code to this reproducer can be found at https://github.com/j-be/mojarra-4145.

BalusC:
The reproducer contains PrimeFaces. This is not unambiguous. I'll have to change the reproducer myself in order to prepare a proper integration test. In future Mojarra-targeted issue reports, it would be very welcome if the reproducer does not contain any third party libraries.

j-be:
@BalusC I see, I used it mosty because due of the `tabs` feature. If you'd like I can try to rewrite it only using JSF features and report back after I succeeeded.

I just thought it would be a good idea to immediatly raise this again before it gets forgotten - I still consider this a being a potential DoS.

BalusC:
I've boiled down an unambiguous reproducer which indeed reproduced the problem in Mojarra 2.3.17.

View:
```
<html
    xmlns:f="http://xmlns.jcp.org/jsf/core"
    xmlns:h="http://xmlns.jcp.org/jsf/html"
>
    <h:head />
    <h:body>
        <h:form>
            <h:commandButton value="Submit" action="#{bean.submit}">
                <f:ajax render="output" />
            </h:commandButton>
            <h:outputText id="output" value="#{bean.output}" />
        </h:form>
    </h:body>
</html>
```

Model:
```
package com.example;

import javax.enterprise.context.RequestScoped;
import javax.inject.Named;

[...]

PR #5037, opened 2022-01-17
Fix ajax error when form feed char is present in partial response

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/5036


BalusC:
All tests passed at localhost.

melloware (approved):

Issue #5059, opened 2022-03-02
FACELETS_REFRESH_PERIOD > 0 triggers incorrect rendering


rmartinc:
**Describe the bug**

If the option `javax.faces.FACELETS_REFRESH_PERIOD` is configured in the `web.xml` to something greater than 0 it makes that some attributes are not rendered OK for the components. In the attached application the HtmlForm loses the `style` attribute.

**To Reproduce**

Compile and deploy the attached app:

```bash
unzip test-jsf.zip
cd test-jsf
mvn clean install
```
Go to the test page, http://127.0.0.1:8080/test-jsf/test.xhtml and click the button two times. The style for the text is different. Check the form in the html page and it has no style attribute.

**Expected behavior**

The form should maintain the style attribute.

**Desktop (please complete the following information):**

 - OS: All
 - Browser All
 - Version All

**Additional context**

Looking the code the problem is that the `DefaultFacelet.java` adds a timestamp when using the refresh time to remove old elements ([here](https://github.com/eclipse-ee4j/mojarra/blob/4.0.0-M5-RELEASE/impl/src/main/java/com/sun/faces/facelets/impl/DefaultFacelet.java#L188-L202)). It happens that the form uses an optimized rendering that renders the attrs in [attributesThatAreSet](https://github.com/eclipse-ee4j/mojarra/blob/master/impl/src/main/java/jakarta/faces/component/UIComponentBase.java#L1893). Those seem to be removed when adding the timestamp.

@arjantijms @BalusC Don't we need to initiate the list of the delta like this?

```diff
diff --git a/impl/src/main/java/jakarta/faces/component/ComponentStateHelper.java b/impl/src/main/java/jakarta/faces/component/ComponentStateHelper.java
index 96d4cd88..e7de3068 100644
--- a/impl/src/main/java/jakarta/faces/component/ComponentStateHelper.java
+++ b/impl/src/main/java/jakarta/faces/component/ComponentStateHelper.java
@@ -210,14 +210,14 @@ class ComponentStateHelper implements StateHelper, TransientStateHelper {
     }
 
     private void initList(Serializable key) {
-        if (component.initialStateMarked()) {
-            deltaMap.computeIfAbsent(key, e -> new ArrayList<>(4));
-        }
-
         if (get(key) == null) {
             List<Object> items = new ArrayList<>(4);
             defaultMap.put(key, items);
         }
+
+        if (component.initialStateMarked()) {
[...]

PR #5062, opened 2022-03-07
Issue #5059: FACELETS_REFRESH_PERIOD > 0 triggers incorrect rendering

rmartinc:
Issue: #5059 

Now the `initList` initializes the `deltaMap` with the values in the `defaultMap`. That way, when restoring, attributes are not lost. Little test added to check that `attributesThatAreSet` are restored OK when there were initial values before marking initial state.


mnriem:
LGTM

PR #5089, opened 2022-05-04
Fix #5077: UIRepeat#visitTree caches/restores wrong index-value

### LINKED ISSUE #5077: UIRepeat: Dynamically added UIRepeat generates invalid clientID
**Describe the bug**

In our application we have a dynamicalle added `UIRepeat`. It is added in `PreRenderViewEvent`. The `UIRepeat` itself is contained in a column of a Primefaces DataTable.

When a not-null list is set as the value of the `UIRepeat`, its DataModel is normally intialized with `index = -1` and then overwritten by `setRowIndex(0)`. _(This could also be a 'bug' in my opinion. Why is the index not always `-1` for a newly created DataModel?)_
During Mojarras implementation of `reapplyDynamicActions`, the `UIRepeat` is visited in the method `locateComponentByClientId` for another component, that is removed from the view (just as an info: as the `ComponentStruct` has `parentClientId=null`, the whole tree is visited, and therefore also the `UIRepeat`).

In `UIRepeat#visitTree` [it will cache the old **row index**](https://github.com/eclipse-ee4j/mojarra/blob/02ea0f946cd354767fdd09b60ecb279ca1631490/impl/src/main/java/com/sun/faces/facelets/component/UIRepeat.java#L650) of the DataModel. Because of the not-null list, the `rowIndex` is `0` and not `-1`, while the `index` of the `UIRepeat` itself is `-1`.

When `UIRepeat#visitTree` completes, the `finally`-block will [restore the `index` of the `UIRepeat`](https://github.com/eclipse-ee4j/mojarra/blob/02ea0f946cd354767fdd09b60ecb279ca1631490/impl/src/main/java/com/sun/faces/facelets/component/UIRepeat.java#L691) by calling `setIndex` but with the previsouly cached `rowIndex` of the DataModel.

So, the state of the `UIRepeat` before `visitTree` is `index=-1, rowIndex=0`, and after `index=0, rowIndex=0`.

This is incorrect as `UIRepeat#getClientId` will now include the `index` if it is `>= 0` - e.g. `dataTable:uirepeat:0` instead of `dataTable:uirepeat`.

As soon as the child components of the `UIRepeat` should be restored by `reapplyDynamicActions`, it uses `invokeOnComponent` which does not expect the iteration index in the clientId and therefore is unable to resolve the `UIRepeat`. (This ends up in a fallback logik of the PrimeFaces DataTable, which thinks, that the ID of the `UIRepeat` must be an interation index of the DataTable which then throws a `NumberFormatException`)

**Expected behavior**

Compared to the MyFaces implementation of `UIRepeat`, the `index` of `UIRepeat` should be cached and not the `rowIndex` of the DataModel.

**Stacktrace off the wrong cached state:**
image

**Stacktrace of the wrong restored state:**
image

**Stacktrace on restore of a child of UIRepeat:**
image

image

image

Bug occurred with Mojarra 2.2.19 / API 2.2.19, but compared to the newer sources of 3.x, the bug should still be there.  I cannot test it, as I do not have any 3.x or newer infrastructure that I can use for that.

fanste:


fanste:
I cannot test the changes, as my ENV is unable to load all dependecies for compilation

melloware (approved):

arjantijms:
> as my ENV is unable to load all dependecies for compilation

Which dependencies is your ENV unable to load?

Did you try to activate the staging repo? E.g. `-Pstaging`?

arjantijms:
In order to pass the ECA check, you need to create an Eclipse account with the email you used for committing and then okay the ECA. You then have an "Eclipse Contributor Agreement" checkbox in your profile, like e.g. here:

https://accounts.eclipse.org/users/atijmse7y (see top right)

fanste:
> > as my ENV is unable to load all dependecies for compilation
> 
> Which dependencies is your ENV unable to load?
> 
> Did you try to activate the staging repo? E.g. `-Pstaging`?

 - Proxy of my internet connection - had to disable SSL checking
 - I had to skip tests as one test produces some strange output which kills the process 
image
 - But besides that I can build the project

BalusC:
I ran the entire 2.3 test set against https://github.com/eclipse-ee4j/mojarra/pull/5091 and all tests passed, so I'm going to merge the PRs.

PR #5095, opened 2022-05-17
Fix #5059: Backport for 2.3 FACELETS_REFRESH_PERIOD > 0

### LINKED ISSUE #5059: FACELETS_REFRESH_PERIOD > 0 triggers incorrect rendering

melloware:
Signed-off-by: melloware <mellowaredev@gmail.com>

### DIFF
--- a/impl/src/test/java/javax/faces/component/UIComponentBaseTestCase.java
+++ b/impl/src/test/java/javax/faces/component/UIComponentBaseTestCase.java
@@ -473,6 +473,21 @@ public class UIComponentBaseTestCase extends UIComponentTestCase {
 
     }
 
+    public void testAttributesThatAreSetStateHolder() throws Exception {
+        ComponentTestImpl c = new ComponentTestImpl();
+        c.getAttributes().put("attr1", "value1");
+        c.markInitialState();
+        c.getAttributes().put("attr2", "value2");
+        assertEquals(Arrays.asList("attr1", "attr2"), c.getAttributes().get("jakarta.faces.component.UIComponentBase.attributesThatAreSet"));
+
+        Object state = c.saveState(facesContext);
+        c = new ComponentTestImpl();
+        c.pushComponentToEL(facesContext, c);
+        c.restoreState(facesContext, state);
+        c.popComponentFromEL(facesContext);
+        assertEquals(Arrays.asList("attr1", "attr2"), c.getAttributes().get("jakarta.faces.component.UIComponentBase.attributesThatAreSet"));
+    }
+
     public void testValueExpressions() throws Exception {
 
         UIComponentBase test = (UIComponentBase) component;
--- a/impl/src/test/java/javax/faces/component/UIComponentBaseTestCase.java
+++ b/impl/src/test/java/javax/faces/component/UIComponentBaseTestCase.java
@@ -478,14 +478,14 @@ public class UIComponentBaseTestCase extends UIComponentTestCase {
         c.getAttributes().put("attr1", "value1");
         c.markInitialState();
         c.getAttributes().put("attr2", "value2");
-        assertEquals(Arrays.asList("attr1", "attr2"), c.getAttributes().get("jakarta.faces.component.UIComponentBase.attributesThatAreSet"));
+        assertEquals(Arrays.asList("attr1", "attr2"), c.getAttributes().get("javax.faces.component.UIComponentBase.attributesThatAreSet"));
 
         Object state = c.saveState(facesContext);
         c = new ComponentTestImpl();
         c.pushComponentToEL(facesContext, c);
         c.restoreState(facesContext, state);
         c.popComponentFromEL(facesContext);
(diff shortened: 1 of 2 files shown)


impl/src/test/java/javax/faces/component/UIComponentBaseTestCase.java
```diff
@@ -473,6 +473,21 @@ public void testValueBindings() {
...
+        c.markInitialState();
+        c.getAttributes().put("attr2", "value2");
+        assertEquals(Arrays.asList("attr1", "attr2"), c.getAttributes().get("jakarta.faces.component.UIComponentBase.attributesThatAreSet"));
+
+        Object state = c.saveState(facesContext);
+        c = new ComponentTestImpl();
+        c.pushComponentToEL(facesContext, c);
+        c.restoreState(facesContext, state);
+        c.popComponentFromEL(facesContext);
+        assertEquals(Arrays.asList("attr1", "attr2"), c.getAttributes().get("jakarta.faces.component.UIComponentBase.attributesThatAreSet"));
```
BalusC, line 488: Is this valid for 2.3 at all? (javax.faces)

impl/src/test/java/javax/faces/component/UIComponentBaseTestCase.java
```diff
@@ -473,6 +473,21 @@ public void testValueBindings() {
 
     }
 
+    public void testAttributesThatAreSetStateHolder() throws Exception {
+        ComponentTestImpl c = new ComponentTestImpl();
+        c.getAttributes().put("attr1", "value1");
+        c.markInitialState();
+        c.getAttributes().put("attr2", "value2");
+        assertEquals(Arrays.asList("attr1", "attr2"), c.getAttributes().get("jakarta.faces.component.UIComponentBase.attributesThatAreSet"));
```
BalusC, line 481: Is this valid for 2.3 at all? (javax.faces)

impl/src/test/java/javax/faces/component/UIComponentBaseTestCase.java
melloware: Sorry updated.  Cherry pickde the commit from the higher branch whoops!

BalusC:
FYI: all 2.3 tests passed

melloware:
nice!

PR #5102, opened 2022-05-24
Fix #4260: AttachedObjectListHolder IndexOutOfBounds fix 2.3

### LINKED ISSUE #4260: java.lang.IndexOutOfBoundsException: Index: 0, Size: 0 still occurs
For a @ViewScoped page we are catching an exception in phase RESTORE_VIEW:
  javax.faces.FacesException: Unexpected error restoring state for component with id gF:dgr_tabset:dgrStatus.  Cause: java.lang.IndexOutOfBoundsException: Index: 0, Size: 0.

The issue should have been fixed with #4011 in [2.1.29-05] which ports #3245 (JAVASERVERFACES-3241). 

Our stacktrace looks very much the same as those mentioned in #3245: [Stacktrace-ACME-2017-07-12.txt](https://github.com/javaserverfaces/mojarra/files/1141653/Stacktrace-ACME-2017-07-12.txt)

The issue still occurs very often with [2.1.29-08] in our performance tests. And occasionally in production, in fact just a few minutes ago. Apparently the exception does not happen deterministically.

The cause is the following code in AttachedObjectListHolder.restoreState

```
        } else {
            // assume 1:1 relation between existing attachedObjects and state
            for (int i = 0, len = attachedObjects.length; i < len; i++) {
                T l = this.attachedObjects.get(i);
                if (l instanceof StateHolder) {
                    ((StateHolder) l).restoreState(context, attachedObjects[i]);
                }
            }
        }
```
this.attachedObjects is an empty list and thus throws the IndexOutOfBoundsException. This is the same code in all versions up to [2.3.2]. (Except for formatting of generics).

First of all I would like to point on the assumption that is documented in the comment. This assumption of an 1:1 relation is not fulfilled, so in my opinion the code should throw an IllegalStateException.

Neither could we find the root cause of the problem, nor do we have a simple test to reproduce the problem. Strong load seems to facilitate the problem.

melloware:
Signed-off-by: melloware <mellowaredev@gmail.com>

### DIFF
--- a/impl/src/main/java/javax/faces/component/AttachedObjectListHolder.java
+++ b/impl/src/main/java/javax/faces/component/AttachedObjectListHolder.java
@@ -140,7 +140,7 @@ class AttachedObjectListHolder<T> implements PartialStateHolder {
                     add(restored);
                 }
             }
-        } else if (this.attachedObjects != null) {
+        } else if (this.attachedObjects != null && this.attachedObjects.length == attachedObjects.length) {
             // Assume 1:1 relation between existing attachedObjects and state
             for (int i = 0, len = attachedObjects.length; i < len; i++) {
                 T attachedObject = this.attachedObjects.get(i);
--- a/impl/src/main/java/javax/faces/component/AttachedObjectListHolder.java
+++ b/impl/src/main/java/javax/faces/component/AttachedObjectListHolder.java
@@ -140,7 +140,7 @@ class AttachedObjectListHolder<T> implements PartialStateHolder {
                     add(restored);
                 }
             }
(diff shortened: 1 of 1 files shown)


BalusC:
This PR fixes a potential NullPointerException but the associated ticket reports an IndexOutOfBoundsException.

wodarski:
Hello guys,

Any plans when Mojarra 2.3.18 is going to be released with AttachedObjectListHolder IndexOutOfBounds fix?

Regards,
Mauricio

arjantijms:
@wodarski Thanks for asking, 2.3.18 was scheduled for this week. I'll run the TCK on the 2.3 branch now.

PR #5115, opened 2022-06-16
Various Javadoc fixes

Emkas:
* Removed empty @return
* Fixed broken links
* Removed broken and badly placed Javadoc
* Corrected @throws declarations
* @param before @throws

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/view/ViewScopeContextObject.java
+++ b/impl/src/main/java/com/sun/faces/application/view/ViewScopeContextObject.java
@@ -35,7 +35,7 @@ class ViewScopeContextObject implements Serializable {
      * Constructor.
      *
      * @param passivationCapableId the return from PassivationCapable.getId().
-     * @param creationalContext the creational context.
+     * @param name name of context.
      */
     public ViewScopeContextObject(String passivationCapableId, String name) {
         this.passivationCapableId = passivationCapableId;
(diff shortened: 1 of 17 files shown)


melloware (approved):

Emkas:
> Thank you for all these PR's cleaning up!

You're all welcome :-) I've got many more of it.

When I've imported this project to eclipse I saw a sea of various warnings (more than 2k). And I don't like that :-) Mostly they are about raw types. Lack of them creates a space where there are test that are basically checking if raw types in Java works (like `fail(Should have thrown ClassCastException)`). I suspect that this is because the roots of the project remembers Java 4.

But I need more time on this, since I need to figure out which changes concert api and which are internal. I want to touch only the second ones. (And I hope that tests coverage in this project is good enough :-) Since I base mostly on them.)

arjantijms:
> When I've imported this project to eclipse I saw a sea of various warnings (more than 2k).

That's true. We've been addressing a number of them throughout the years, but there's more to be done as you noticed. If you really want to see a lot of those, just take a look at the 2.2 version of the sources :O

arjantijms:
> And I hope that tests coverage in this project is good enough

There's a large number of tests indeed, but we moved them to the [jakarta/faces](https://github.com/jakartaee/faces/tree/master/tck) project. It are nearly 6000, and they take some 2 hours to run.

Emkas:
> There's a large number of tests indeed, but we moved them to the [jakarta/faces](https://github.com/jakartaee/faces/tree/master/tck) project. It are nearly 6000, and they take some 2 hours to run.

Great! Lesser chance that I'll break something.

For now I get rid of about half of the warnings, but not all of them are as trivial as I thought in the first glance :-)

Issue #5045, opened 2022-02-06
Remove FrameTransport and use FormData


mnriem:
See issue https://github.com/eclipse-ee4j/mojarra/issues/5040 for background

BalusC:
First: https://github.com/eclipse-ee4j/mojarra/issues/5044#issuecomment-1046246568

mnriem:
I am closing this one as I have not found anyone to pick it up or found time to work on it myself.

BalusC:
Will be fixed via https://github.com/eclipse-ee4j/mojarra/pull/5141

Issue #5040, opened 2022-01-24
jsf.js | Remove FrameTransport + Remove IE6-7-8-9 support


pizzi80:
Please remove FrameTransport and use FormData for ajax

I think it's also time to remove check for IE9 and all the relative hacks in the code...

https://caniuse.com/mdn-api_formdata

https://github.com/omnifaces/omnifaces/issues/656

there is already a PR that implement FormData 

https://github.com/eclipse-ee4j/mojarra/pull/4626

mnriem:
Because this is asking for removal it needs to be the next cycle around as we will add a deprecation notice for it as part of the upcoming 4.0 release

pizzi80:
deprecate it on v3.0.3 and remove it on 4.0 that is not yet released ...

BalusC:
In big favor of this but I'd rather await for ALL existing JSF 2.x tests to be migrated for Faces 4 before I make serious changes to faces.js.

pizzi80:
Thanks!
In the meantime it could be possible to deprecate all the old functions?

This way it will be possible to implement it on v4.x after tests' migration...

arjantijms:
The spec process would only allow deprecation in a major new version, which is 4.0.

pizzi80:
Ok ...

it's only a matter of performance and size of jsf.js (or faces.js) ... 

The main point is to use FrameTransport only when 

` isIE() && context.form.enctype === "multipart/form-data" `

and then implement the FormData logic inside the sendRequest function ...

so there is nothing to deprecate or remove, it's an implementation that 
can be made even in v2.3 

Then in Faces v5 we can raise the level of the JS to standard ES6 
and remove half the lines of code of faces.js

mnriem:
Closing as this issue has been separated into 2 separate issues, one for deprecation for 4.0 and one for using FormData for 5.0.

BalusC:
Will be fixed via https://github.com/eclipse-ee4j/mojarra/pull/5141

PR #5125, opened 2022-06-21
Maven JS compressor + Faces.js + Faces.js partial submit + Java optimizations

pizzi80:
This is my first contribution to Mojarra, 
sorry if I made some mistake with the PR, I'm not used to contribute to open source project on github ... :smiley: 

### Maven
 - Removed yuicompressor (dead project with no support for ES6+ js)
 + Added Google Closure maven plugin for minification

### Faces JS
 + tested and minified with ECMASCRIPT5_STRICT profile of Google Closure
 + use strict
 + const and let instead of var
 - Removed IE-5-6-7-8-9-10 support
 - Removed IE 5-6-7 hacks and memory hacks
 - Removed FrameTransport
 - Removed / Commented out unused and very old functions and code
 + FormData for File upload
 + Many minor optimizations
 + for..of instead of for with index
 + 10 kB Minified and Gzipped

### Java
 + Java Generics 
 + some minor optimizations to old style code (probably Java 4/5/6)
 + ConcurrentHashMap instead of synchronized block 
 + CDI.current instead of manual jndi lookup

#5045 
#5040


pizzi80:
talking with @BalusC we came to the same conclusion that inside faces.js 
the method `faces.getViewState` should include the partial submit logic and, to do that, 
the best solution is to change the firm of the method like this:

`faces.getViewState = function(form,execute) `

I'll work on it and add the commit to this PR

pizzi80:
fixed https://github.com/eclipse-ee4j/mojarra/issues/5126

pizzi80:
> talking with @BalusC we came to the same conclusion that inside faces.js the method `faces.getViewState` should include the partial submit logic and, to do that, the best solution is to change the firm of the method like this:
> 
> `faces.getViewState = function(form,execute) `
> 
> I'll work on it and add the commit to this PR
> 
> [omnifaces/omnifaces#699](https://github.com/omnifaces/omnifaces/issues/699)

implemented partial submit, at the moment is always active :)

pizzi80:
> Please revert all of the Java part and split into separate PRs per kind of optimization so that the entire PR is fully focused on faces.js

ok, I have to play a bit with github ...
 
> The JNDI way of retrieving bean manager should not be removed. There is no guarantee that CDI.current is available at the moment the bean manager is needed (e.g. Tomcat startup).

no problem but:

I'm on Tomcat 10 latest with Java 17 latest and Weld 4 or 5 

I use only CDI.current and friends with no problem since 1 year now

Do you know if it is something related with multiple applications deploy?
... because I usually prefer to deploy only one app per tomcat and never had this kind of problem

PR #5128, opened 2022-06-24
MockServletContext - clean-up

Emkas:
* Add @Override
* Remove unused functions/methods
* Add raw types where they can be used


impl/src/test/java/com/sun/faces/mock/MockServletContext.java
```diff
@@ -39,111 +40,116 @@
...
 
+    @Override
     public String getContextPath() {
         return "test";
     }
 
+    @Override
     public String getInitParameter(String name) {
-        return ((String) parameters.get(name));
+        return (parameters.get(name));
```
BalusC, line 74: Unnecessary parenthesis should also be removed over all place.

Emkas: Sure. My bad.

PR #5129, opened 2022-06-24
jakarta.faces.component.html: remove OPTIMIZED_PACKAGE

Emkas:
Method `#handleAttribute` was probably added to all `jakarta.faces.component.html` package's classes. But not in all cases this method is used.

### DIFF
--- a/impl/src/main/java/jakarta/faces/component/html/HtmlComponentUtils.java
+++ b/impl/src/main/java/jakarta/faces/component/html/HtmlComponentUtils.java
@@ -1,0 +1,41 @@
+package jakarta.faces.component.html;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+import jakarta.el.ValueExpression;
+import jakarta.faces.component.UIComponent;
+
+class HtmlComponentUtils {
+
+    private static final String ATTRIBUTES_THAT_ARE_SET = "jakarta.faces.component.UIComponentBase.attributesThatAreSet";
+    private static final String OPTIMIZED_PACKAGE = "jakarta.faces.component.";
+
+    static void handleAttribute(UIComponent component, String name, Object value) {
+        final Map<String, Object> attributes = component.getAttributes();
+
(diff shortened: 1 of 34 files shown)


BalusC:
This is a leftover of the JSF 1.x era when those classes in `javax.faces.component.html` were autogenerated based on some XML configuration file. The generator was shut down and the last generated classes were plainly copypasted into the package.

I'd like to see this PR to be changed to simply deduplicate all those `handleAttribute()` methods into a package private static helper class/method so that you can ultimately use e.g. `HtmlComponentUtils.handleAttribute(this, "foo", foo);`.

Emkas:
@BalusC I wanted to do this "the smart way" but to be honest I was defeated since I tried to use interface for this. There is no place to hide the method, so I can be only public.
I thought such you would not like such a util class. But if you do, no problem! I'd update my PR.

Emkas:
OK. I added `HtmlComponentUtils.handleAttribute(this, "foo", foo);`. There's the same code in `com.sun.faces.component.PassthroughElement` but there I only added `@SuppressWarnings("unchecked")` and a `TODO`.

impl/src/main/java/jakarta/faces/component/html/HtmlComponentUtils.java
```diff
@@ -0,0 +1,41 @@
...
+
+    static void handleAttribute(UIComponent component, String name, Object value) {
+        final Map<String, Object> attributes = component.getAttributes();
+
+        @SuppressWarnings("unchecked")
+        List<String> setAttributes = (List<String>) attributes.get(ATTRIBUTES_THAT_ARE_SET);
+
+        if (setAttributes == null) {
+            String cname = component.getClass().getName();
+            if (cname != null && cname.startsWith(OPTIMIZED_PACKAGE)) {
```
BalusC, line 23: I realize this is original code but cname can really never be null.

Emkas: Good catch!

PR #5133, opened 2022-06-24
Remove raw types warnings with Class

Emkas:


Emkas:
Hmm... this is odd. Build failed. I use `mvn clean install` and all looks fine. Now I even did what fails in a build `mvn --no-transfer-progress install -Pstaging` and it ends fine as well. I there any command I should look for?

BTW. Shouldn't this be `mvn clean --no-transfer-progress install -Pstaging`?

BalusC:
This PR is too big to quickly glance over, but I see that this change also affected the public API at a handful places (i.e. these changes basically also affect the generated javadocs). I recommend to not do that in the context of this PR.

Emkas:
OK. PR is still big. If you want to to split, please tell me.

I removed everything what is touching public methods/fields etc.

impl/src/main/java/com/sun/faces/cdi/CdiProducer.java
```diff
@@ -49,7 +49,7 @@
 
     private String id = this.getClass().getName();
     private String name;
-    private Class<?> beanClass = Object.class;
+	private Class<?> beanClass = Object.class;
```
BalusC, line 52: Tab needs to be spaces. Also in rest of PR.

Emkas: There were only 2. Nevertheless, sorry for those remainings.

PR #5145, opened 2022-06-29
Using raw types in various classes

Emkas:


BalusC:
LGTM but please resolve conflict.

PR #5149, opened 2022-06-30
Various fixes of warnings in tests

Emkas:
* Fixed warnings in mocks - collective change for all warnings without changes in public interfaces
* Fixed warnings in component tests - as above
* HtmlResponseWriterTest: close writer - I know those are tests, but still, warnings easy to fix. I think this way is better than try-with-resources because any exception will break tests (and this is good)
* Remove unused code
* Remove empty UIComponent test overrides -  they don't do anything. Tests are run anyway, because of the base class. This looks like generated code waiting to be filled.

### DIFF
--- a/impl/src/test/java/com/sun/faces/mock/MockApplication.java
+++ b/impl/src/test/java/com/sun/faces/mock/MockApplication.java
@@ -258,7 +258,7 @@ public class MockApplication extends Application {
     }
 
     @Override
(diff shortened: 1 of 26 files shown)


melloware (approved):

Emkas:
ping :-)

arjantijms (approved):
Thanks for the ping and the pr ;)

PR #5158, opened 2022-10-20
Make sure the beanClass attribute of all synth beans points to the CDI extension declaring them

### LINKED ISSUE #5157: CDI bean attribute `beanClass` is not used correctly in CDI extension and producers
**Describe the bug**

Mojarra registers a lot of synthetic CDI beans via a CDI extension and for all of them it sets a `beanClass` attribute.
Setting this is valid operation but the values used don't make much sense. Currently, the extension [uses `Object` as a basis](https://github.com/eclipse-ee4j/mojarra/blob/master/impl/src/main/java/com/sun/faces/cdi/CdiProducer.java#L52) and each producer then sets their own class - [here is an example](https://github.com/eclipse-ee4j/mojarra/blob/master/impl/src/main/java/com/sun/faces/cdi/RequestParameterMapProducer.java#L45).

Now, CDI spec is not very helpful in telling users what `beanClass` stands for and [its javadoc](https://github.com/jakartaee/cdi/blob/master/api/src/main/java/jakarta/enterprise/inject/spi/Bean.java#L37-L43) only says what it is in case of class-based beans and producer methods.
In reality, the `beanClass` attribute is used (by Weld, the reference/compatible impl) to [determine where to instantiate a proxy](https://github.com/weld/core/blob/5.1.0.Final/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java#L393-L401) for any of these beans. This, based on environment, means by which class loader or in which Java module.
For this reason, Weld defaults `beanClass` of all synthetic beans to the extension class that registered them.

**_EDIT:_** _After some more digging in the CDI spec, i should correct myself in that it is clearer what `beanClass` should stand for in case of synth beans. There is a [section](https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#inter_module_injection_full) that mentions this and points out the importance of bean class in determining inter-module injection. In other words, this controls if the bean is available for injection in certain module or not._

This was recently noticed in [WildFly](https://github.com/wildfly/wildfly/pull/16183), where synth beans from Mojarra were registered under a different module.

**Expected behavior**

Synth beans should have a `beanClass` attribute of the CDI extension that registers them or at least the class of the producer bean that it uses.

Just for the sake of clarity I will also mention that `beanClass` does not anyhow affect bean types and the resolution of the bean.

manovotn:
Fixes #5157

### DIFF
--- a/impl/src/main/java/com/sun/faces/cdi/ApplicationMapProducer.java
+++ b/impl/src/main/java/com/sun/faces/cdi/ApplicationMapProducer.java
@@ -42,7 +42,7 @@ public class ApplicationMapProducer extends CdiProducer<Map<String, Object>> {
 
     public ApplicationMapProducer() {
         super.name("applicationScope").scope(ApplicationScoped.class).qualifiers(ApplicationMap.Literal.INSTANCE)
-                .types(new ParameterizedTypeImpl(Map.class, new Type[] { String.class, Object.class }), Map.class, Object.class).beanClass(Map.class)
+                .types(new ParameterizedTypeImpl(Map.class, new Type[] { String.class, Object.class }), Map.class, Object.class)
                 .create(e -> FacesContext.getCurrentInstance().getExternalContext().getApplicationMap());
     }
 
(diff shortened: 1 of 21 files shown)


BalusC:
LGTM but I want to run TCK on this first.

BalusC:
@manovotn can you please merge the PRs on your fork?

manovotn:
I've rebased this PR onto current `4.0` branch which is what I filed it against originally.
@BalusC I assume this should also go into `master` branch?

BalusC:
TCK passed.

PR #5186, opened 2022-12-13
Add back id attribute to VDL of h:head and h:body

BalusC:
https://github.com/jakartaee/faces/issues/1760
- Add back id attribute to VDL of h:head and h:body
- fix HeadRenderer to only output it when actually specified
- fix BodyRenderer to only output it when HTML5 doctype is used""

### DIFF
--- a/impl/src/main/java/com/sun/faces/renderkit/html_basic/BodyRenderer.java
+++ b/impl/src/main/java/com/sun/faces/renderkit/html_basic/BodyRenderer.java
@@ -54,7 +54,9 @@ public class BodyRenderer extends HtmlBasicRenderer {
     public void encodeBegin(FacesContext context, UIComponent component) throws IOException {
         ResponseWriter writer = context.getResponseWriter();
         writer.startElement("body", component);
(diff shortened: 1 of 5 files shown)


BalusC:
Ran entire tck localhost as well, all green.

melloware (approved):

BalusC:
Thomas can you review at least the spec change in taglib? https://github.com/eclipse-ee4j/mojarra/pull/5186/files#diff-a977313af2132f4b6dc25adc6bd8719a6822c5ae62ef547d1d7d82bbdc0baab3

BalusC:
@tandraschko can you please review?

melloware:
I know he is on holiday but i think he is back sometime next week.

tandraschko:
@BalusC i wonder why this was removed in general?

In MyFaces it looks like we always render the head/body id if specified by the user, doent matter if html5 or not

BalusC:
It was accidentally removed, see also https://github.com/jakartaee/faces/issues/1760#issuecomment-1349635281

Issue #5190, opened 2022-12-16
f:selectItem String null value


pizzi80:
**Describe the bug**

It is not possible to send the null value with f:selectItem 
when the backing bean property type is String

**To Reproduce**

Using the example at page 123 of "The definitive guide to Faces in Jakarta EE 10"

```
<h:form id="select">
    <h:selectOneMenu id="value" value="#{testViewAction.value}" >
        <f:selectItem itemValue="#{null}" itemLabel="-- select one --" />
        <f:selectItem itemValue="one" itemLabel="First item" />
        <f:selectItem itemValue="two" itemLabel="Second item" />
        <f:selectItem itemValue="three" itemLabel="Third item" />
    </h:selectOneMenu>

    <h:commandButton id="submit" value="Submit" action="#{testViewAction.submit}" />
</h:form>
```

```
@Named
@RequestScoped 
public class TestViewAction { 

    private static final Logger log = Logger.getLogger(MethodHandles.lookup().getClass().getName());

    private String value;
    public String getValue() {return value;}
    public void setValue(String value) {this.value = value;}

    public void submit() {
       log.info("["+value+"] --> value is null? " + (value==null) + ", value is empty? " + ("".equals(value)) + ", value is whitespace? " + (" ".equals(value) ) );
    }
}
```

**Expected behavior**

When you submit the form choosing the first element of the select
the backing bean value should be null and not an empty String 

**Desktop (please complete the following information):**

 - OS: Fedora 37  / Win 11 
 - Browser:  Chrome / FF
 - Version latest

**Stack**

Faces 4.0
Tomcat 10.1.4
Jdk: 17.0.5

I also tried to replace the EL implementation with the Eclispe's one 
but the result is the same (empty String instead of null value)
https://github.com/eclipse-ee4j/expressly

**Additional context**

There is a long story behind this (10+ years) on StackOverflow and on @BalusC blog

I tried all the solutions provided, and the problem is still on EL 
because when you submit the null value: 

1) the StringConverter is totaly bypassed 
2) the Faces init param jakarta.faces.INTERPRET_EMPTY_STRING_SUBMITTED_VALUES_AS_NULL
is totally useless

I don't know which is the best place to solve this, I've implemented a naive solution using 
a custom EL resolver like the one proposed by @BalusC but with a special String value which 
I use as null, eg.  'faces.string.null'

```
public class EmptyToNullStringELResolver extends ELResolver {

[...]

PR #5193, opened 2022-12-17
Mojarra issue 5190 merged into 4 0

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/5190


pizzi80:
I had to use the following check inside EmptyStringToNullELResolver
```
if (    value == null &&
                String.class == targetType &&
                context instanceof org.apache.el.lang.EvaluationContext
    ) {
```

otherwise EL produce NPE during String concat, 
for example:

#{'hello'.concat(null)}

PR #5203, opened 2023-01-19
The service file parsing is incomplete

JD-CSTx:
Issue https://github.com/eclipse-ee4j/mojarra/issues/5204.

According to the official documentation of Javas Service Loader it may contain comments. Everything which comes after the '#' is considered a comment. Reference: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/ServiceLoader.html.
But Mojarras own Service Loader implementation doesn't support comments. It even doesn't skip blank lines. This leads to unexpected behaviour when using Mojarras Service Loader. When files under META-INF/services are loaded from Mojarras Service Loader it may lead to exceptions, while these service files under any other META-INF/services directory may perform perfect when they're loaded by the JDK Service Loader.

Therefore I implemented a Pull-Request which fixes that, so comments and blank lines don't lead to crashes when loading service files.

I also added unit tests to test the new functionality. To an area which had previously no test coverage at all.

I also replaced the magic string `"UTF-8"` with `StandardCharsets.UTF_8` in the method I updated. It was introduced with Java 7 and can therefore also be used in a project where downwards compability is important.

P.S.: I read, understood and accepted the ECA.

### DIFF
--- a/impl/src/main/java/com/sun/faces/spi/ServiceFactoryUtils.java
+++ b/impl/src/main/java/com/sun/faces/spi/ServiceFactoryUtils.java
@@ -24,6 +24,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.net.URL;
 import java.net.URLConnection;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Enumeration;
@@ -100,7 +101,7 @@ final class ServiceFactoryUtils {
                     input = conn.getInputStream();
                     if (input != null) {
                         try {
-                            reader = new BufferedReader(new InputStreamReader(input, "UTF-8"));
+                            reader = new BufferedReader(new InputStreamReader(input, StandardCharsets.UTF_8));
                         } catch (Exception e) {
                             // The DM_DEFAULT_ENCODING warning is acceptable here
                             // because we explicitly *want* to use the Java runtime's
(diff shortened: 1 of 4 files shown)


melloware:
Nice work.  Can you open a Ticket please so they can link the issue with this PR?

impl/src/main/java/com/sun/faces/spi/ServiceFactoryUtils.java
```diff
@@ -143,4 +149,29 @@ static String[] getServiceEntries(String key) {
...
+        } else if (commentIndex > 0) {
+            // Cut everything of after '#'.
+            line = line.substring(0, commentIndex);
+        }
+        line = line.trim();
+        if (line.isEmpty()) {
+            return null;
+        } else {
+            return line;
+        }
```
melloware, line 175: Here is what MyFaces does it might look a little cleaner...

```java
                                int idx = serviceImplName.indexOf('#');
                                if (idx >= 0)
                                {
                                    serviceImplName = serviceImplName.substring(0, idx);
                                }
                                serviceImplName = serviceImplName.trim();

                                if (serviceImplName.length() != 0)
                                {
                                    results.add(serviceImplName);
                                }
```

JD-CSTx:
> Nice work.
Thanks.

> Can you open a Ticket please so they can link the issue with this PR?
Done. Link: https://github.com/eclipse-ee4j/mojarra/issues/5204

melloware:
Thank you!

impl/src/main/java/com/sun/faces/spi/ServiceFactoryUtils.java
JD-CSTx: I appreciate your effort, but MyFaces is under the Apache 2 License, which is to my understanding not really compatible with the EPL. I don't want to copy their code (even if they probably don't care).

melloware: You don't need to quite copy it but you can see its a much simpler version.  And Apache 2 is basically the most friendly open source license there is...

JD-CSTx:
Tried to fix the Eclipse verification, so I recreated the branch.

Issue #5204, opened 2023-01-19
Mojarras Service Loader can't parse comments (#)


JD-CSTx:
**Describe the bug**

According to the official documentation of Javas Service Loader it may contain comments. Everything which comes after the '#' is considered a comment. Reference: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/ServiceLoader.html.
But Mojarras own Service Loader implementation doesn't support comments. It even doesn't skip blank lines. This leads to unexpected behaviour when using Mojarras Service Loader. When files under META-INF/services are loaded from Mojarras Service Loader it may lead to exceptions, while these service files under any other META-INF/services directory may perform perfect when they're loaded by the JDK Service Loader.

Therefore I implemented a Pull-Request which fixes that, so comments and blank lines don't lead to crashes when loading service files.

I also added unit tests to test the new functionality. To an area which had previously no test coverage at all.

I also replaced the magic string `"UTF-8"` with `StandardCharsets.UTF_8` in the method I updated. It was introduced with Java 7 and can therefore also be used in a project where downwards compability is important.

Pull-Request: https://github.com/eclipse-ee4j/mojarra/pull/5205

**To Reproduce**

Add a service file under META-INF/services, with a comment line, to a JAR which gets loaded and see how Mojarra fails to load a _class_ named, for example `# Copyright 2017, 2021 Eclipse Foundation. Eclipse License 2.`.

**Expected behavior**

Comments and blank lines are ignored.

P.S.: I read, understood and accepted the ECA.

PR #5205, opened 2023-01-19
Completed service file parsing.

JD-CSTx:
Issue https://github.com/eclipse-ee4j/mojarra/issues/5204.

According to the official documentation of Javas Service Loader it may contain comments. Everything which comes after the '#' is considered a comment. Reference: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/ServiceLoader.html.
But Mojarras own Service Loader implementation doesn't support comments. It even doesn't skip blank lines. This leads to unexpected behaviour when using Mojarras Service Loader. When files under META-INF/services are loaded from Mojarras Service Loader it may lead to exceptions, while these service files under any other META-INF/services directory may perform perfect when they're loaded by the JDK Service Loader.

Therefore I implemented a Pull-Request which fixes that, so comments and blank lines don't lead to crashes when loading service files.

I also added unit tests to test the new functionality. To an area which had previously no test coverage at all.

I also replaced the magic string `"UTF-8"` with `StandardCharsets.UTF_8` in the method I updated. It was introduced with Java 7 and can therefore also be used in a project where downwards compability is important.

P.S.: I read, understood and accepted the ECA.

### DIFF
--- a/impl/src/main/java/com/sun/faces/spi/ServiceFactoryUtils.java
+++ b/impl/src/main/java/com/sun/faces/spi/ServiceFactoryUtils.java
@@ -24,6 +24,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.net.URL;
 import java.net.URLConnection;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Enumeration;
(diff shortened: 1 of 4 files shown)


JD-CSTx:
Old PR (wasn't working, used another company e-mail): https://github.com/eclipse-ee4j/mojarra/pull/5203

BalusC (approved):

Issue #5183, opened 2022-12-08
FacesContext EL variable incorrectly references previous instance in web.xml error page


BalusC:
https://github.com/omnifaces/showcase/issues/19 with Mojarra 4.0.0.

Basically:
1. have a Facelet as general error page registered in web.xml
2. in a test Facelet, let the view reference #{facesContext} during postback and let the bean action method throw an exception on a non-ajax request (this will internally abort the request, release the faces context, and the server will trigger response.sendError() which in turn creates new faces context if the error page happens to be a Facelet)
3. in the error page source, try to reference `#{facesContext.something}`

Expected behavior: it works fine.
Actual behavior: it throws ISE in assertNotReleased() because `#{facesContext}` EL variable is still referencing the previous instance from before response.sendError()

It worked fine in Mojarra 2.3 / 3.0.

BalusC:
It worked fine as `#{facesContext}` EL in previous Mojarra versions because ImplicitObjectELResolver took care of it. However that still gave the wrong instance via `@Inject FacesContext` in beans. After removal of ImplicitObjectELResolver and relying on `@Produces FacesContext` it went wrong in `#{facesContext}` EL as well.

I implemented sort of work around by manually destroying the CDI-produced instance in `FacesContext#release()`.

PR #5207, opened 2023-01-26
FacesContext EL variable incorrectly references previous instance after sendError() to a Facelet

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/5183 Quick work around for now. Cleanest solution would have been a custom scope like `@FacesContextScoped`. I'll look at it next.


tandraschko:
@BalusC we have something similar in MyFaces: FacesScoped but a better name is FacesContextScoped, yup.

PR #5210, opened 2023-01-28
FacesConfig#version() cannot return null

BalusC:
https://github.com/jakartaee/faces/issues/1784

### DIFF
--- a/impl/src/main/java/jakarta/faces/annotation/FacesConfig.java
+++ b/impl/src/main/java/jakarta/faces/annotation/FacesConfig.java
@@ -57,7 +57,7 @@ public @interface FacesConfig {
 
         @Override
         public Version version() {
-            return null; // non binding, so not used
+            return Version.JSF_2_3;
         }
     }
 


BalusC:
@arjantijms and @tandraschko WDYT? Wondering between returning `JSF_2_3` or adding something like `NULL` or `LATEST` as this isn't intended to be ever used anywhere.

tandraschko:
then lets return JSF_2_3 for now, we will remove it in 4.1 anyway?

BalusC:
Right. 

The only issue is that it's an API change and hence this PR targeted 4.1 instead of 4.0. But indeed we're going to remove it nonetheless.

tandraschko:
Cant we just add it to 4.0?

BalusC:
I'm all for it because it's clearly a bug, but technically speaking, it's still an API change and these should go in a minor/major version.

@arjantijms can't we make an exception for this one?

tandraschko:
for me its ok and already commited to MF
as long as we dont change method signatures, its not a problem IMO

BalusC:
Fair point. Ok, so be it.

arjantijms:
I gave this some extra thought, and I now think it's an implementation change/bug fix, and not an API one. 

The body of the `version()` method is implementation. By definition, it should return whatever the main annotation has as a default value:

See https://github.com/eclipse-ee4j/mojarra/blob/8b5e83b8016a4130b08ea22005ce4966c732e119/impl/src/main/java/jakarta/faces/annotation/FacesConfig.java#L94

BalusC:
Bon. Let's approve it then.

arjantijms (approved):

PR #5221, opened 2023-03-27
POC for Mojarra specific API split.

arjantijms:
* This minimizes the amount of code in API classes
* A delegate in the Mojarra namespace is used that contains all code
* For now as POC only UIViewRoot and its two base classes have been done
* Base classes now contain an if/else and have some things made public
* If/else and new public things to be removed later

See: https://github.com/eclipse-ee4j/mojarra/blob/5268967de26ea890b8f6eca121f04607a89374d3/impl/src/main/java/jakarta/faces/component/UIViewRoot.java


impl/src/main/java/com/sun/faces/api/component/UIColumnImpl.java
```diff
@@ -0,0 +1,131 @@
...
+     * </p>
+     */
+    public static final String COMPONENT_TYPE = "jakarta.faces.Column";
+
+    /**
+     * <p>
+     * The standard component family for this component.
+     * </p>
+     */
+    public static final String COMPONENT_FAMILY = "jakarta.faces.Column";
```
BalusC, line 45: These two constants belong in UIColumn.

impl/src/main/java/com/sun/faces/util/Util.java
```diff
@@ -456,6 +456,12 @@ public static void notNegative(String varname, long number) {
         }
     }
 
+    public static void notNull(Object var) {
+        if (var == null) {
+            throw new NullPointerException();
+        }
+    }
```
BalusC, line 463: Objects.requireNonNull() does the same.

impl/src/main/java/jakarta/faces/component/UIColumn.java
```diff
@@ -16,6 +16,8 @@
 
 package jakarta.faces.component;
 
+import com.sun.faces.api.component.UIColumnImpl;
```
BalusC, line 19: Is this intentional or temporary?

arjantijms: Definitely temporary. I'd like to have the components working first, and then in a second pass hide those *impl classes.

BalusC:
Superseded by https://github.com/eclipse-ee4j/mojarra/pull/5319

Issue #5222, opened 2023-04-06
JSF pages containing `#{session.id}` fail with NotSerializableException when sessions are distributable


pferraro:
**Describe the bug**

JSF pages containing `#{session.id}` fail with:
java.io.NotSerializableException: com.sun.faces.cdi.SessionProducer$$Lambda$1586/0x0000000100cc6478
... when deployed as a <distributable/> web application.

SessionProducer creates a `SessionScoped` bean using a non-serializable function, which naturally fails to serialize.
I see nothing about this bean that should warrant it being `SessionScoped` rather than `RequestScoped`.

**To Reproduce**

A reproducer for this issue is available at: https://issues.redhat.com/browse/WFLY-17840

**Expected behavior**

`#{session.foo}` should behave the same as `#{request.session.foo}`.

**Additional context**

Changing `#{session.foo}` expressions to `#{request.session.foo}` sufficiently works around the issue.  This resolves the expression via the RequestProducer, which, of course, is RequestScoped.

PR #5223, opened 2023-04-06
[5222] JSF pages containing #{session.id} fail with NotSerializableException when sessions are distributable

pferraro:
Fixes https://github.com/eclipse-ee4j/mojarra/issues/5222


pferraro:
Question - if I wanted to get this into a future 4.0.3 release, to which branch would I backport this fix?

Issue #5226, opened 2023-04-11
EmptyStringToNullELResolver


pizzi80:
**Describe the bug**

with the `EmptyStringToNullELResolver` active

EL string concatenation has problems

**To Reproduce**

`#{'hello'.concat(null)}`

```
Stack trace: java.lang.NullPointerException: Cannot invoke "String.isEmpty()" because "str" is null 
at java.base/java.lang.String.concat(String.java:2769) 
at java.base/jdk.internal.reflect.GeneratedMethodAccessor598.invoke(Unknown Source) 
at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) 
at java.base/java.lang.reflect.Method.invoke(Method.java:568) 
at jakarta.el.BeanELResolver.invoke(BeanELResolver.java:161) 
at jakarta.el.CompositeELResolver.invoke(CompositeELResolver.java:78) 
at org.apache.el.parser.AstValue.getValue(AstValue.java:159) 
at org.apache.el.parser.AstBracketSuffix.getValue(AstBracketSuffix.java:36) 
at org.apache.el.parser.AstValue.getValue(AstValue.java:143) 
at org.apache.el.parser.AstMethodParameters.getParameters(AstMethodParameters.java:33) 
at org.apache.el.parser.AstValue.getValue(AstValue.java:158) 
at org.apache.el.ValueExpressionImpl.getValue(ValueExpressionImpl.java:190) 
at org.jboss.weld.module.web.el.WeldValueExpression.getValue(WeldValueExpression.java:50) 
at com.sun.faces.facelets.el.ELText$ELTextVariable.writeText(ELText.java:208) 
at com.sun.faces.facelets.compiler.TextInstruction.write(TextInstruction.java:44) 
at com.sun.faces.facelets.compiler.UIInstructions.encodeBegin(UIInstructions.java:42) 
at com.sun.faces.renderkit.html_basic.HtmlBasicRenderer.encodeRecursive(HtmlBasicRenderer.java:255) 
at com.sun.faces.renderkit.html_basic.PassthroughRenderer.encodeChildren(PassthroughRenderer.java:75) 
at jakarta.faces.component.UIComponentBase.encodeChildren(UIComponentBase.java:557) 
at com.sun.faces.renderkit.html_basic.HtmlBasicRenderer.encodeRecursive(HtmlBasicRenderer.java:257) 
[...]

PR #5228, opened 2023-04-12
FIX EmptyStringToNullELResolver NPE (issue 5226)

pizzi80:
- FIX [https://github.com/eclipse-ee4j/mojarra/issues/5226](https://github.com/eclipse-ee4j/mojarra/issues/5226)

### DIFF
--- a/impl/src/main/java/com/sun/faces/el/EmptyStringToNullELResolver.java
+++ b/impl/src/main/java/com/sun/faces/el/EmptyStringToNullELResolver.java
@@ -16,13 +16,15 @@
 
 package com.sun.faces.el;
 
-import java.beans.FeatureDescriptor;
-import java.util.Iterator;
-
 import jakarta.el.ELContext;
 import jakarta.el.ELResolver;
 
+import java.beans.FeatureDescriptor;
+import java.util.Iterator;
+
 public class EmptyStringToNullELResolver extends ELResolver {
+
+    public static final String EVALUATION_CONTEXT_CLASS_SIMPLE_NAME = "EvaluationContext";
 
     @Override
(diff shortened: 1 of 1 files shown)


impl/src/main/java/com/sun/faces/el/EmptyStringToNullELResolver.java
```diff
@@ -32,7 +34,7 @@ public Class<?> getCommonPropertyType(ELContext context, Object base) {
     @Override
     @SuppressWarnings("unchecked")
     public <T> T convertToType(ELContext context, Object value, Class<T> targetType) {
-        if (value == null && targetType == String.class) {
+        if ( value == null && targetType == String.class && EVALUATION_CONTEXT_CLASS_SIMPLE_NAME.equals(context.getClass().getSimpleName()) ) {
```
BalusC, line 37: This is a hack not a fix. This is prone to fail when a different EL implementation is used which doesn't use `EvaluationContext` as class name in this specific situation.

The correct fix is most probably to just return an empty string.

pizzi80: Yes I know :smiley:  :smiley: 
but at the moment I had to fix my fork, and it works at least on Tomcat  ...
I'm not able to fix it on a long term perspective :smiling_face_with_tear:

BalusC: This hack goes wrong on `#{bean.setValue(null)}`. It will unexpectedly set empty string there.

BalusC:
Declined for reason as explained in https://github.com/eclipse-ee4j/mojarra/issues/5226#issuecomment-1546695840

PR #5231, opened 2023-04-14
faces.js Faces TCK test issues fixed.

pizzi80:
the only test that's failing is 

`Issue2906IT.testCommandLinksInRepeat`

I've looked at the test but it's quite obscure
and I'm not sure if it's well written.

I tried the view in a separate playground app
and the UI works as expected...


arjantijms (approved):

arjantijms:
Thanks @pizzi80 

I guess we should take a separate look `Issue2906IT.testCommandLinksInRepeat` then.

PR #5242, opened 2023-06-16
Generics + Some code optimization (at least for readability)

pizzi80:
- Generics
- StringBuffer -> StringBuilder
- Minor modernizations

I made half of the TCK tests with no errors

Also I'm testing this release in production on a side (but complete) crud project 
and everything works as expected

I have problem to execute tests that uses selenium because chrome version is now 114 and not 86...
I tried to change the release of selenium inside /tck/util/pom.xml to the latest but the error remains...

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/ApplicationImpl.java
+++ b/impl/src/main/java/com/sun/faces/application/ApplicationImpl.java
@@ -479,7 +479,7 @@ public class ApplicationImpl extends Application {
      * @see jakarta.faces.application.Application#createConverter(Class)
      */
     @Override
-    public Converter<?> createConverter(Class<?> targetClass) {
+    public <T> Converter<T> createConverter(Class<T> targetClass) {
         return instanceFactory.createConverter(targetClass);
     }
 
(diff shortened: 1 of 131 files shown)


pizzi80:
When will be possible to change the api?

Jsf 2.3 has Generics in converters and validators

Jsf 2.2 is in EOL since maybe 2019 

Generics is Java 5 technology,
Jakarta EE11 (few months) will be Java 21+...

pizzi80:
Regarding the splitted PRs it's my fault but these are basic changes with an IDE ...I'm using intellij and those changes are considered quite mandatory 

An interesting point is the memory consumption of the JVM ... After those changes I see a 25% drop in Code cache ...  It's probably the effects of code normalization using static  methods  of java lang instead of custom repeated code inside non static methods ... ?

Looking at the code what I see is a lacking of common utils ...(this is a problem of all Jakarta sub projects) for example in Faces there are 5 different cache implementations totally custom and some misterious data structure probably due to copy paste of code from other utils

BalusC:
> When will be possible to change the api?

Next Faces version.

> Jsf 2.3 has Generics in converters and validators
> Jsf 2.2 is in EOL since maybe 2019
> Generics is Java 5 technology,
> Jakarta EE11 (few months) will be Java 21+...

True. We can change it for next Faces version but not for current version. You're basically breaking the API contract. You can basically only do that for a next major version. I.e. Faces 5.0.

PR #5246, opened 2023-06-19
StringBuffer -> StringBuilder

pizzi80:
+ StringBuffer -> StringBuilder
+ fixed: `static trimRight(String s)` function inside `TextUnit` didn't trim anything

can I move this and other String util functions inside the Utils class?

for example inside the MessageFactory classes there is a function 
that should format a String with the passed parameters and it's copy/pasted
all over the place

also it returns `null` if the `locale` parameter is null... is it expected?

### DIFF
--- a/impl/src/main/java/com/sun/faces/facelets/compiler/TextUnit.java
+++ b/impl/src/main/java/com/sun/faces/facelets/compiler/TextUnit.java
@@ -141,7 +141,7 @@ final class TextUnit extends CompilationUnit {
             addInstruction(new CommentInstruction(el));
         }
 
-        buffer.append("<!--" + text + "-->");
+        buffer.append("<!--").append(text).append("-->");
     }
 
     public void startTag(Tag tag) {
@@ -182,7 +182,7 @@ final class TextUnit extends CompilationUnit {
 
     private void finishStartTag() {
         if (tags.size() > 0 && startTagOpen) {
-            buffer.append(">");
+            buffer.append('>');
             startTagOpen = false;
         }
     }
@@ -285,7 +285,7 @@ final class TextUnit extends CompilationUnit {
             if (Character.isWhitespace(s.charAt(i))) {
                 i--;
             } else {
-                return s;
+                return s.substring(0,i+1);
             }
         }
         return "";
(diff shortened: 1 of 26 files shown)


impl/src/main/java/com/sun/faces/facelets/compiler/TextUnit.java
```diff
@@ -285,7 +285,7 @@ private static String trimRight(String s) {
             if (Character.isWhitespace(s.charAt(i))) {
                 i--;
             } else {
-                return s;
+                return s.substring(0,i+1);
```
BalusC, line 288: Nice catch! This is indeed an existing bug.

impl/src/main/java/com/sun/faces/facelets/util/Classpath.java
```diff
@@ -93,18 +94,18 @@ public static URL[] search(ClassLoader cl, String prefix, String suffix, SearchA
...
-                    boolean searchDone = searchDir(all, new File(URLDecoder.decode(url.getFile(), "UTF-8")), suffix);
+                    boolean searchDone = searchDir(all, new File(URLDecoder.decode(url.getFile(), StandardCharsets.UTF_8)), suffix);
                     if (!searchDone) {
                         searchFromURL(all, prefix, suffix, url);
                     }
                 }
             }
         }
-        URL[] urlArray = (URL[]) all.toArray(new URL[all.size()]);
+        URL[] urlArray = all.toArray(new URL[0]);
```
BalusC, line 104: Why was `all.size()` changed by `0`? This will now unnecessarily create an empty URL array.

impl/src/main/java/com/sun/faces/facelets/compiler/TextUnit.java
pizzi80: thanks ;)

pizzi80:
fixed :+1:

impl/src/main/java/com/sun/faces/facelets/compiler/TextUnit.java
BalusC, line 288: This regressed as https://github.com/eclipse-ee4j/mojarra/issues/5989, so not really a bug in hindsight :o

PR #5251, opened 2023-06-20
UrlBuilder: Charset instead of String

pizzi80:
In this way we don't have to convert from String to Charset inside a for loop
and we can use the "new" `URLEncoder.encode(String,Charset)`

### DIFF
--- a/impl/src/main/java/com/sun/faces/context/UrlBuilder.java
+++ b/impl/src/main/java/com/sun/faces/context/UrlBuilder.java
@@ -16,17 +16,20 @@
 
 package com.sun.faces.context;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
 import java.net.URLEncoder;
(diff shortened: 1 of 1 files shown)


impl/src/main/java/com/sun/faces/context/UrlBuilder.java
```diff
@@ -16,21 +16,18 @@
...
-import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.nio.charset.Charset;
+import java.util.*;
```
BalusC, line 21: Can you please reconfigure your IDE to not do that anymore? Wildcard imports are highly discouraged.

PR #5253, opened 2023-06-20
Utils: improvements and new functions

pizzi80:


impl/src/main/java/com/sun/faces/util/CollectionsUtils.java
```diff
@@ -19,10 +19,7 @@
 import static java.util.Arrays.asList;
 import static java.util.Collections.unmodifiableMap;
 
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
BalusC, line 22: Can you please reconfigure your IDE to not do that anymore?

PR #5255, opened 2023-06-20
PhaseId optimization

pizzi80:


impl/src/main/java/jakarta/faces/event/PhaseId.java
```diff
@@ -16,19 +16,24 @@
...
+import static java.util.stream.Collectors.toMap;
 
 /**
  * <p>
  * <span class="changed_modified_2_2">Typesafe</span> enumeration of the legal values that may be returned by the
  * <code>getPhaseId()</code> method of the {@link FacesEvent} interface.
  */
 
-public class PhaseId implements Comparable {
+public class PhaseId implements Comparable<PhaseId> {
```
BalusC, line 36: This is a public API change. You can't do that without a ticket in https://github.com/jakartaee/faces/issues and referencing the PR to it.

impl/src/main/java/jakarta/faces/event/PhaseId.java
```diff
@@ -66,12 +76,12 @@ private PhaseId(String newPhaseName) {
      * if this object is less than, equal to, or greater than the specified object.
      * </p>
      *
-     * @param other The other object to be compared to
+     * @param phaseId The other {@link PhaseId} to be compared to
      */
     @Override
-    public int compareTo(Object other) {
+    public int compareTo(PhaseId phaseId) {
```
BalusC, line 82: Idem.

impl/src/main/java/jakarta/faces/event/PhaseId.java
```diff
@@ -248,6 +239,13 @@ public static PhaseId phaseIdValueOf(String phase) {
...
      */
-    public static final List<PhaseId> VALUES = Collections.unmodifiableList(Arrays.asList(values));
+    public static final List<PhaseId> VALUES = List.of(values);
+
+    /**
+     * <p>
+     * Valid {@link PhaseId} instances, mapped by their uppercase name
+     * </p>
+     */
+    public static final Map<String,PhaseId> VALUES_BY_NAME = unmodifiableMap(Stream.of(values).collect(toMap( phase -> phase.getName().toUpperCase() , identity())));
```
BalusC, line 249: This is a new field in the API. Please create a ticket in https://github.com/jakartaee/faces/issues referencing this PR. And you also you need to make 1st paragraph of javadoc a `<p class="changed_added_4_1">` and add a `@since 4.1` to the end. See also e.g. https://github.com/eclipse-ee4j/mojarra/pull/5187/commits/5e61f0d9b419e419d413c4509c8797809f54aac6 for an example.

BalusC: Or change to `private` if you didn't intend to expose this in the public API.

pizzi80: changed to private for now...

It's better to transform `PhaseId` to an `enum`

PR #5259, opened 2023-06-21
Basic Converters improved

pizzi80:
+ Basic Converters with Generics
+ Basic Converters optimized
+ Enum Converter is now stateless


BalusC:
This is a public API change. This needs a ticket in https://github.com/jakartaee/faces/issues

PR #5263, opened 2023-06-22
[5262] Do not set the response buffer size if set to -1.

### LINKED ISSUE #5262: Not honoring value of -1 for the jakarta.faces.FACELETS_BUFFER_SIZE application parameter
In https://github.com/eclipse-ee4j/mojarra/issues/5164 a change was made to honor the default value of 1024. However, the spec also [states](https://jakarta.ee/specifications/faces/4.0/jakarta-faces-4.0.html#a6088):

> A value of -1 will not assign a buffer size on the response. 

This does not seem to be implemented and is just passed to the `HttpServletResponse`.

**To Reproduce**

1. Set the context parameter of `jakarta.faces.FACELETS_BUFFER_SIZE` to `-1`.

**Expected behavior**

The value should not be set on the `jakarta.servlet.http.HttpServletResponse`.

**Additional context**

This fails on WildFly as it tries to allocate a buffer size of less than 0.

jamezp:
resolves #5262

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/view/FaceletViewHandlingStrategy.java
+++ b/impl/src/main/java/com/sun/faces/application/view/FaceletViewHandlingStrategy.java
@@ -404,7 +404,9 @@ public class FaceletViewHandlingStrategy extends ViewHandlingStrategy {
                 getSession(ctx);
             }
 
-            stateWriter = new WriteBehindStateWriter(extContext.getResponseOutputWriter(), ctx, responseBufferSize);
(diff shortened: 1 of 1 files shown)


jamezp:
I've still got one failing TCK with this change in WildFly. I'll make it a PR once I figure that out.

BalusC:
Hmm this was already added to 4.0.x and should just be upmerged to master. I think 5.0 label is not correct here, @arjantijms ?

PR #5266, opened 2023-06-22
LRUCache + ELUtils

pizzi80:
+ LRUCache implementation adapted to the Util codebase of Faces
+ ELUtils use the new LRUCache instead of a custom local implementation

The LRUCache can also be used in other part of the code that now use
a `Collection.synchronizeMap` over the `LRUMap`


impl/src/main/java/com/sun/faces/el/ELUtils.java
```diff
@@ -252,31 +244,29 @@ public static Object coerce(Object value, Class<?> toType) {
...
+
+
+
+
+
+
+
+
+
+
```
BalusC, line 269: What? :)

BalusC:
Looks great :)

impl/src/main/java/com/sun/faces/el/ELUtils.java
pizzi80: :rofl:

pizzi80:
I've further improved and unified with the `Cache.Factory` inside `util.Cache`

what do you think?

BalusC:
This is a great improvement.

PR #5267, opened 2023-06-22
faces.js: isAutoExec test is not needed anymore

pizzi80:
Replacing DOM element's innerHTML does not execute the (eventually) injected javascript.
This is standard, and we don't need to do this test anymore.

I'll leave the function only as a placeholder to identify the places where it is used in the code.

In the future a new and unique DOM replacement algorithm will replace the current functions, 
and it will be possible to remove the isAutoExec() function as well


BalusC:
+1

Issue #5269, opened 2023-06-28
ArrayIndexOutOfBoundsException during restore state


pizzi80:
**Describe the bug**

If `ViewExpiredException` is thrown inside an action 
the restoreState throws an `ArrayIndexOutOfBoundsException`

**To Reproduce**

```
<h:form id="test" >
    <h:commandButton id="expired" action="#{testErrors.throwViewExpiredException}" value="JSF Expired" />
</h:form>
```

```
@Named
@RequestScoped
public class TestErrors {

    public void throwViewExpiredException() {
		throw new ViewExpiredException();
	}

}
```

```
Caused by: java.lang.ArrayIndexOutOfBoundsException: Index 2 out of bounds for length 2
	at jakarta.faces.component.UIComponentBase.restoreState(UIComponentBase.java:1223)
	at com.sun.faces.application.view.FaceletPartialStateManagementStrategy.lambda$restoreView$1(FaceletPartialStateManagementStrategy.java:360)
	at com.sun.faces.component.visit.FullVisitContext.invokeVisitCallback(FullVisitContext.java:125)
	at jakarta.faces.component.UIComponent.visitTree(UIComponent.java:1266)
	at jakarta.faces.component.UIComponent.visitTree(UIComponent.java:1278)
	at jakarta.faces.component.UIComponent.visitTree(UIComponent.java:1278)
	at com.sun.faces.application.view.FaceletPartialStateManagementStrategy.restoreView(FaceletPartialStateManagementStrategy.java:349)
	at com.sun.faces.application.view.ViewHandlingStrategy.restoreView(ViewHandlingStrategy.java:90)
	at com.sun.faces.application.view.FaceletViewHandlingStrategy.restoreView(FaceletViewHandlingStrategy.java:256)
	at com.sun.faces.application.view.MultiViewHandler.restoreView(MultiViewHandler.java:129)
	at jakarta.faces.application.ViewHandlerWrapper.restoreView(ViewHandlerWrapper.java:96)
	at jakarta.faces.application.ViewHandlerWrapper.restoreView(ViewHandlerWrapper.java:96)
	at org.omnifaces.viewhandler.OmniViewHandler.restoreView(OmniViewHandler.java:128)
	at jakarta.faces.application.ViewHandlerWrapper.restoreView(ViewHandlerWrapper.java:96)
	at com.sun.faces.lifecycle.RestoreViewPhase.execute(RestoreViewPhase.java:161)
[...]

PR #5270, opened 2023-06-28
Fix for ArrayIndexOutOfBoundsException during restore state

pizzi80:
+ Fix for ArrayIndexOutOfBoundsException during restore state
+ Minor modernizations and code simplifications

https://github.com/eclipse-ee4j/mojarra/issues/5269


BalusC:
Declined, see ticket. This represents actually a real problem in app code logic and should not be hidden/suppressed.

pizzi80:
ok, I'll remove that part and resubmit the java "optimization" parts

PR #5274, opened 2023-07-07
Fixed: Form id always included in ajax payload

pizzi80:
Talking with @BalusC 
I've found that the new faces.js was not including the form.id 
in the xhr payload.

https://github.com/omnifaces/omnifaces/issues/767

+ fixed: form.id always included in the ajax payload
+ various optimizations

I've made some minor optimizations and other clarifications
which I need for a future big refactoring of the "render" phase algorithms I'm working on

### DIFF
--- a/impl/src/main/resources/META-INF/resources/jakarta.faces/faces-uncompressed.js
+++ b/impl/src/main/resources/META-INF/resources/jakarta.faces/faces-uncompressed.js
@@ -25,8 +25,8 @@
 "use strict";
 
 // Detect if this is already loaded, and if loaded, if it's a higher version
-if ( !( (faces && faces.specversion && faces.specversion >= 40000 )
-    && (faces.implversion && faces.implversion >= 4)) ) {
+if ( !( (window.faces && window.faces.specversion && window.faces.specversion >= 40000 )
+    && (window.faces.implversion && window.faces.implversion >= 4)) ) {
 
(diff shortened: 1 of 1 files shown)


BalusC:
Can you PR the form ID fix against 4.0 please?

pizzi80:
> Can you PR the form ID fix against 4.0 please?

I think that 4.0 doesn't need to be fixed because the "partial submit" part is totally new 

You should fix it on OmniFaces side in Form.ts simply by adding the form.id in the query string if form.id it's not included in the array of id that need to be processed...

pizzi80:
```
// if the partialExecuteIds does not include the form.id,
// then add it because it's required by the spec to be always included!
if ( partialExecuteIds && !partialExecuteIds.includes(form.id) ) {
       qString = appendToQueryString(qString, form.id, form.id);
}
```

BalusC:
Right.

BalusC:
> *You should fix it on OmniFaces side in Form.ts*

Not needed. It already does that and it's also covered by [`FormIT`](https://github.com/omnifaces/omnifaces/blob/b36a580c04722bb459e02176fbb3386152947787/src/test/java/org/omnifaces/test/component/form/FormIT.java#L53) since the beginning. If the Mojarra fix didn't work at all while using `<o:form>` then it would indeed have been a bug in OmniFaces as well.

PR #5283, opened 2023-07-22
3.0.5 has been released

BalusC:


impl/pom.xml
```diff
@@ -33,7 +33,7 @@
     
     <groupId>org.glassfish</groupId>
     <artifactId>jakarta.faces</artifactId>
-    <version>3.0.4</version>
+    <version>3.0.5</version>
```
BalusC, line 36: This is not correct?

Issue #5213, opened 2023-02-15
DataTable: Navigation from header facet causes the value EL to be re-evaluated


mihaiDumea:
Navigation from header facet causes the value EL to be re-evaluated, recreating the bean referenced by the EL in a new scope. This recreation is unnecessary as is for no use and may cause problems. This doesn't happen for navigation from table rows.

**To Reproduce**

Also reported here [https://github.com/primefaces/primefaces/issues/9793](https://github.com/primefaces/primefaces/issues/9793)

mnriem:
@BalusC What timeframe are you looking for cutting the last 2.3 release?

BalusC:
I wanted to release it as soon as PRs are reviewed+merged.

PR #5287, opened 2023-07-23
RowIndex/DataModel of UIData doesn't need to be reset when broadcasted row index is same as current row index; noticed same thing on UIRepeat as well so also improved over there

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/5213


BalusC:
Don't merge yet I want to run TCK on this.

BalusC:
@arjantijms can you please review this?

arjantijms (approved):

PR #5288, opened 2023-07-23
Fix incorrect reference to non-existent 'context' argument in Application#unsubscribeFromEvent methods

BalusC:
https://github.com/jakartaee/faces/issues/1712


BalusC:
cc: @tandraschko

tandraschko:
+1

PR #5292, opened 2023-07-29
Remove unused and deprecated @PostConstructCustomScopeEvent and @PreDestroyCustomScopeEvent

BalusC:
https://github.com/jakartaee/faces/issues/1830

cc: @tandraschko (note: for 5.0)


pnicolucci:
We need to hold off on this for 5.0 unless an exception to the process is granted. These classes were deprecated in 4.1 which is not part of any Jakarta Platform (EE10/EE11 etc). See the email thread: [https://www.eclipse.org/lists/faces-dev/msg00321.html](https://www.eclipse.org/lists/faces-dev/msg00321.html)

BalusC:
It should be noted that these classes have no effect at all because they are since 4.0 not anymore invoked because the associated `@CustomScoped` scope implementation has been removed in 4.0.

BalusC:
Declined. We can apparently not deprecate things in a minor version.

Issue #5296, opened 2023-08-04
UIData: return static empty data model if value is null


Rapster:
Macro optimization:

In `UIData#getDataModel`, if value is null, a new instance of an empty datamodel will be returned, instead reuse a static immutable instance.

For example:

```java
private static final DataModel EMPTY_DATA_MODEL = new ListDataModel(Collections.emptyList());

@Override
protected DataModel getDataModel() {
  if (model != null) {
    return (model);
  }

  Object current = getValue();
  if (current == null) {
    setDataModel(EMPTY_DATA_MODEL);
  }
  // etc.
}
```

melloware:
Original Discussion in PrimeFaces: https://github.com/primefaces/primefaces/pull/10426

pizzi80:
Why you didn't create a PR ?

If you can't I'll create one but it's basically a copy/paste of your code ...

pizzi80:
:smiley:  :+1: :+1:

BalusC:
2.3.21 and 4.0.4 have been released with the improvement.

PR #5299, opened 2023-08-11
UIData: return static empty data model if value is null

pizzi80:
https://github.com/eclipse-ee4j/mojarra/issues/5296


melloware (approved):

melloware:
Shouldn't this be done on the 2.3 branch so it can be promoted to 3.x an d4.x by BalusC?

melloware (changes requested):
See comment

BalusC:
Already done via https://github.com/eclipse-ee4j/mojarra/pull/5304

Issue #5245, opened 2023-06-19
Migrate messages properties files to UTF-8


BalusC:
Since Java9 properties files are (finally) by default read using UTF-8 instead of ISO-8859-1.

So, since Faces 4 supports Java11 instead of Java8, we should be able to migrate them in order to get rid of illegible Unicode escape sequences in for example the French and Chinese message bundles such as seen in this PR: https://github.com/eclipse-ee4j/mojarra/pull/5243/files#diff-259fe9f794427cd476baa91adc4048d19cffb09922888ba234afcc30eff4cff5

BalusC:
This is in hindsight less trivial than anticipated. When I migrate them then the tooling (both Eclipse and IntelliJ -- no Netbeans here) stubbornly still attempt to read them using ISO-8859-1 even though the project is configured with Java 9+ and therefore they present UTF-8 encoded properties files as mojibake in their editor and I cannot seem to find any configuration setting to reinstruct the IDE to interpret them as UTF-8. This is very inconvenient when you want to edit them.

melloware:
I have run into this issue too and I end up editing these files in Notepad++ because Eclipse always messes them up.

mnriem:
@BalusC Were you able to complete this?

pzygielo:
> This is in hindsight less trivial than anticipated. When I migrate them then the tooling (both Eclipse and IntelliJ -- no Netbeans here) stubbornly still attempt to read them using ISO-8859-1

Please report to JetBrains if that setting
![image](https://github.com/eclipse-ee4j/mojarra/assets/11896137/62f27917-98bc-493c-9569-50e13177b135)
does not work for you. 
It works fine for me.

BalusC:
> > This is in hindsight less trivial than anticipated. When I migrate them then the tooling (both Eclipse and IntelliJ -- no Netbeans here) stubbornly still attempt to read them using ISO-8859-1
> 
> Please report to JetBrains if that setting ![image](https://user-images.githubusercontent.com/11896137/268837770-62f27917-98bc-493c-9569-50e13177b135.png) does not work for you. It works fine for me.

Fabulous, that indeed worked! I couldn't find that.

BalusC:
> @BalusC Were you able to complete this?

On second thought I'll prep the PR nonetheless.

BalusC:
@manfredriem I picked 4.1 instead of 4.0 because I feel like this change warrants another minor version. Feel free to retarget 4.0 instead.

PR #5318, opened 2023-09-23
Migrate properties files to UTF-8

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/5245


mnriem:
@BalusC I think it would be fine to be in a 4.0 release. @arjantijms What do you think?

arjantijms (approved):

arjantijms:
LGTM, with the comment that I think we should not apply i18n to logging anymore in the future, but today it is what it is.

Issue #4480, opened 2018-09-29
Remove Mojarra specific classes from API jar


ggam:
The Java EE 8 API dependency we all use has the Mojarra version of the JSF API at the moment JSF 2.3 was released. It's very likely that your runtime will contain a more up-to-date Mojarra version, with modified method bodies (and lines). That makes debugging with the platform sources impossible, since the class you are putting breakpoints on isn't the same that is actually on the classpath.

This has to be known to MyFaces users, as it has a completely different implementation of all the classes. They already need to attach their version of the API in order to debug. Mojarra users should do the same.

In addition, the final API jar contains some package-private classes that are Mojarra specific. Tt's misleading to have them as part of the API jar and under the `javax.faces` package.

So I propose to:
- Stop reusing Mojarra API as the "official" JAR.
- Create the needed classes on the API module, with no-op methods unless for code that's truly standard.
- Put comments in the methods explaining behavior and pointing to the relevant sections of the spec when the JavaDoc is not enough.

That way, when people try to dig into JSF API, they will clearly see that it's implementation specific and that they need to get the specific version in order to debug.

This is related to #4457.

ggam:
I will provide a PR to showcase this

arjantijms:
This will need a wee bit more of work though. Due to all the historical luggage it's not that simple I'm afraid.

arjantijms:
@ggam 

Perhaps the best way forward with this would be to create a new project first; https://github.com/eclipse-ee4j/jsf-api

I believe that under Eclipse rules anyone can propose and incubate such a project. Would be great if you could kick that off.

deki:
Any update on that? A new project for the API sounds like a good idea...

bshannon:
The API jar file is only expected to be good enough for compiling against.
If you want an API jar file that's a reusable component that all implementations
can build on and use at runtime, that requires a bit of re-engineering at best.

tandraschko:
Exactly. I think it's impossible (or at lest very bad decision and ineffective) to create a API jar, that can be used for Mojarra and MyFaces, as it contains very much implementation details.

But it would be a big beneift would be create a minimal API jar, without any logic (just like mentioned above), which would just throw java.lang.UnsupportedOperationException.

Then we could also merge API and IMPL in MyFaces, which reduces a lot of duplicate code and reflection ****.

arjantijms:
[...]

PR #5319, opened 2023-09-24
Interim commit for #4480

mnriem:


BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/4480

BalusC:
Looks good!

Only why are unit tests hard removed? Shouldn't these be moved to faces/tck?

These conflicts highlight a painful backwards incompatibility with earlier versions tho. Perhaps best to wait until 4.1 is released before sorting out conflicts and then merging.

mnriem:
@arjantijms  @BalusC They were not conflicting when I submitted the PR. So I will resolve these and ask for a quick review again and then merge it to 5.0 if you are both OK with that?

BalusC:
> They were not conflicting when I submitted the PR

Indeed. These files were changed after you submitted the PR. It's just that this type of conflict shows that we cannot anymore simply merge 4.x into 5.x for future changes to 4.x.

mnriem:
@BalusC Which is why an umbrella issue is good to have that shows which PRs went where. I also always start with the most recent version first and then execute backports.

mnriem:
@BalusC @arjantijms I realize this is not the end of splitting the API out. Next is to make sure Glassfish is fine with the split. Which means we need someone to add the new API jar to that project. @arjantijms Is that something you can do?

BalusC:
How about these deleted tests? Shouldn't we move them to the Faces project?

mnriem:
@BalusC I have restored the tests, but as they are very tied to the Mojarra implementation I left all but one test in the implementation project. Can we now go ahead and merge this in? Thanks!

BalusC:
Yes I understood that they've some com.sun.faces deps but all these tests against API classes could simply be moved into the Faces project as part of TCK?

BalusC (approved):
Approved nonetheless. We need to move forward.

mnriem:
@BalusC Next step now would be to move the module to the API project :)

PR #5322, opened 2023-09-29
Fixes #5316 - ClientWindow: Custom PrimeClientWindowFactory not being respected

### LINKED ISSUE #5316: ClientWindow: Custom PrimeClientWindowFactory not being respected
## Description

**Mojarra Version:** 4.0.4

Original Issue reported for 2.3.21: https://github.com/eclipse-ee4j/mojarra/issues/5297

PrimeFaces has its own PrimeClientWindow and by overriding the client window factory like this:

```xml
<client-window-factory>org.primefaces.clientwindow.PrimeClientWindowFactory</client-window-factory>
```

This works fine in MyFaces but in Mojarra it does not pick up.  To turn it on we have to enable in `web.xml`

```xml
    <context-param>
        <param-name>jakarta.faces.CLIENT_WINDOW_MODE</param-name>
        <param-value>url</param-value>
    </context-param>
```

**Reproducer:**
Attached is a reproducer simply test it yourself with.
[pf-client-window-factory.zip](https://github.com/eclipse-ee4j/mojarra/files/12642763/pf-client-window-factory.zip)

`mvn clean jetty:run -Pmojarra40` **FAILS**
`mvn clean jetty:run -Pmyfaces40` **WORKS**

When you navigate to `http://localhost:8080/primefaces-test/` this is the result:

Mojarra 4.0.4: `http://localhost:8080/primefaces-test/`
MyFaces 4.0.1: `http://localhost:8080/primefaces-test/test.xhtml?jfwid=4a6b4`

You can see the proper Window ID is added to the URL.

We do NOT have to add `jakarta.faces.CLIENT_WINDOW_MODE`  for MyFaces as it overrides the default ClientWindowFactory with our own.  It seems like in Mojarra it is still loading the default ClientWindowFactory first and then overriding with our custom one thus the need to for the `web.xml` param.

PrimeFaces Issue: https://github.com/primefaces/primefaces/issues/10444

PrimeFaces current PR: https://github.com/primefaces/primefaces/pull/10449

**Expected behavior**

When I enable `PrimeClientWindow` it activates WITHOUT the need to set `jakarta.faces.CLIENT_WINDOW_MODE`

mnriem:


melloware (approved):
Tested and our use case now works!

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/5316

BalusC (approved):

PR #5326, opened 2023-10-01
Merge ActionSource2 back into ActionSource and deprecate ActionSource2 related classes

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/5327


impl/src/main/java/jakarta/faces/component/ActionSource2.java
```diff
@@ -26,7 +26,9 @@
  * </p>
  *
  * @since 1.2
+ * @deprecated Use {@link ActionSource} instead.
  */
+@Deprecated(since = "4.1", forRemoval = true)
```
mnriem, line 31: I doubt a 4.1 release will be done. Can you retarget it for 5.0?

BalusC: Will do when 4.1 will be merged into 5.0 because it's easier to find&replace 4.1 by 5.0 in one go.

mnriem (approved):
@BalusC Can you make sure you also file an implementation issue for this as well so we can tie the 2 PRs you will need. One targeting the 4.1 branch (this PR) 2. Another targeting the 5.0 branch? And link the implementation issue back to the spec issue? And set the 4.1 and 5.0 label on the implementation issue. Thanks!

BalusC:
ok but not sure I understand why .. github is capable of linking to issue in another repo as you can see in all previous tickets/prs/merges.

mnriem:
> ok but not sure I understand why .. github is capable of linking to issue in another repo as you can see in all previous tickets/prs/merges.

Ease of generating release notes and later on ease of finding for the issue in which branches it was fixed by having a common ancestor. Similary API issues do not show up as release notes issues in Mojarra.

PR #5349, opened 2023-11-04
Add rowStatePreserved property to UIRepeat, exactly the same as UIData

BalusC:
https://github.com/jakartaee/faces/issues/1263

cc: @tandraschko

### DIFF
--- a/impl/src/main/java/jakarta/faces/component/UIData.java
+++ b/impl/src/main/java/jakarta/faces/component/UIData.java
@@ -16,6 +16,11 @@
 
 package jakarta.faces.component;
 
+import static com.sun.faces.facelets.tag.faces.ComponentSupport.restoreFullDescendantComponentDeltaStates;
+import static com.sun.faces.facelets.tag.faces.ComponentSupport.restoreFullDescendantComponentStates;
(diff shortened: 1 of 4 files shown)


impl/src/main/java/jakarta/faces/component/UIData.java
```diff
@@ -16,6 +16,11 @@
 
 package jakarta.faces.component;
 
+import static com.sun.faces.facelets.tag.faces.ComponentSupport.restoreFullDescendantComponentDeltaStates;
```
mnriem, line 19: This won't work for 5.0 as UIData won't have access to com.sun.faces packages

mnriem: I suggest you move the necessary code into a private Utils class (or reuse one of the existing ones if possible) in the jakarta.faces.component package.

mnriem (changes requested):
Make is so the UIData and UIRepeat both use a class from the jakarta.faces.component package (I suggest PackageUtils

BalusC:
Already took into account this while approving your PR :)

mnriem (approved):

BalusC:
@mnriem there you go: https://github.com/eclipse-ee4j/mojarra/pull/5353

PR #5354, opened 2023-11-12
Use CDI as injection provider and fixed InitFacesContext

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/4480

Since API split there were a couple "RESOLVE THIS" comments especially wrt InjectionProvider and InitFacesContext. The InjectionProvider has become CDI and the overcomplicated Thread-InitFacesContext mapping has been removed/simplified.


BalusC:
We should probably also get rid of Util#getCDIBeanManager() in Mojarra and just use CDI.current().getBeanManager() all over place. CDI has currently sufficiently matured that this always works, even in Tomcat. WDYT?

mnriem (approved):

BalusC:
Ok let's merge now, so master branch at least produces a working Mojarra impl.

PR #5360, opened 2023-11-19
Add missing rowStatePreserved attribute to dataTable vdl

BalusC:
https://github.com/jakartaee/faces/issues/1864

While at it also fixed the 'since' version for id attributes of head/body, it was actually added in 2.2 but only to renderkitdoc instead of vdldoc.


mnriem (approved):
LGTM

PR #5363, opened 2023-11-22
[2.3] Do not override application-specified "com.sun.faces.enableDistributable" value

sudeshnas93:
Mojarra defines a number of initialization parameters. While "com.sun.faces.enableDistributable" is one of them, Mojarra automatically overrides any existing setting provided by the application.
If a web application wanted explicitly disable this for a distributable web application, it would normally add a servlet context parameter within web.xml.
e.g.

```
<web-app>
  <distributable/>
  <context-param>
    <param-name>com.sun.faces.enableDistributable</param-name>
    <param-value>false</param-value>
  </context-param>
</web-app>
```
However, the ConfigureListener auto-enables the com.sun.faces.enableDistributable configuration property whenever it see that the tag in web.xml, overriding any custom setting.
I would expect Mojarra to respect the setting defined by the application, and only provide a default value if no value was specified.

### DIFF
--- a/impl/src/main/java/com/sun/faces/config/ConfigureListener.java
+++ b/impl/src/main/java/com/sun/faces/config/ConfigureListener.java
@@ -179,8 +179,11 @@ public class ConfigureListener implements ServletRequestListener, HttpSessionLis
             }
         }
         
-        if (webXmlProcessor.isDistributablePresent()) {
-            webConfig.setOptionEnabled(WebConfiguration.BooleanWebContextInitParameter.EnableDistributable, true);
+        // Do not override if already defined
+        if (!webConfig.isSet(WebConfiguration.BooleanWebContextInitParameter.EnableDistributable)) {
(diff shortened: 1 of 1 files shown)


soul2zimate (commented):
Does this require the same deletion as same as https://github.com/eclipse-ee4j/mojarra/pull/5347 ?

BalusC:
Indeed. It should be the exact same change.

sudeshnas93:
Code changes have been made accordingly. Please review.

mnriem:
@sudeshnas93 Please make sure you use the same email as your Eclipse account so the ECA check passes

sudeshnas93:
I have used the same email address as that of Eclipse account and updated the ECA. Can this test be rerun again?
Thanks.

mnriem (approved):

lvydra:
Hi @BalusC, could you please review the changes?

PR #5364, opened 2023-11-25
Make sure that ViewHandler#getViews() also returns programmatic facelets

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/5365

I only question the validity to add the missing `@Nonbinding` to the `@View` annotation value at this stage already, which was clearly an oversight. I checked spec and apidocs, this is *nowhere* explicitly mentioned for all other existing Faces annotations with a `@Nonbinding` field. So I felt it's kind of safe to fix it in Mojarra 4.0.6 already instead of in Faces 4.1 (which is not yet released!).

If you disagree, let me know then I'll adjust the approach to fetch all CDI beans and then manually filter them instead of fetching them by qualifier, though this might be less efficient. And in the meanwhile try to get that `@Nonbinding` into Faces 4.1 before it gets released.

cc: @tandraschko

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/resource/FaceletWebappResourceHelper.java
+++ b/impl/src/main/java/com/sun/faces/application/resource/FaceletWebappResourceHelper.java
@@ -38,6 +38,7 @@ import com.sun.faces.config.WebConfiguration;
 import com.sun.faces.util.Util;
 
 import jakarta.faces.FacesException;
+import jakarta.faces.annotation.View;
 import jakarta.faces.application.ResourceVisitOption;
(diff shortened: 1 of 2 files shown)


tandraschko:
+1

mnriem:
Unless there is major broken functionality I would take the side of caution and do it for 4.1.

BalusC:
> *I would take the side of caution and do it for 4.1.*

Okay. I already guessed.

BalusC:
@manfredriem https://github.com/eclipse-ee4j/mojarra/pull/5368

I'll adjust current PR.

Issue #5367, opened 2023-11-26
@jakarta.faces.annotation.View#value must be @Nonbinding


BalusC:
https://github.com/jakartaee/faces/issues/1865

BalusC:
Reverted.

PR #5368, opened 2023-11-26
@jakarta.faces.annotation.View#value must be @Nonbinding 

BalusC:
https://github.com/jakartaee/faces/issues/1865
https://github.com/eclipse-ee4j/mojarra/issues/5367

### DIFF
--- a/impl/src/main/java/jakarta/faces/annotation/View.java
+++ b/impl/src/main/java/jakarta/faces/annotation/View.java
@@ -26,6 +26,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
 import jakarta.enterprise.util.AnnotationLiteral;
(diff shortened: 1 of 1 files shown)


mnriem (approved):

volosied:
This causes some regressions in 5.0 / 4.1: 


WELD-001318: Cannot resolve an ambiguous dependency between: - Managed Bean [class com.example.HelloFacelet2] with qualifiers [@Any @View], - Managed Bean [class com.example.HelloFacelet] with qualifiers [@Any @View] 


 I applied a similar change to MyFaces and found the same problem.  I think ` View.Literal.of(someURI)` during the bean lookup doesn't work now because the annotation is not considered anymore?

BalusC:
Right. There's the why. I'll have to revert this for now.

PR #5373, opened 2023-12-11
Support for the Jakarta EE 11 M1 release

arjantijms:
Update dependencies to their M1 versions where applicable, and update code to support this.

Set release target to JDK 17, and removed some resulting warnings.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        java: [ '11' ]
+        java: [ '17' ]
         os: [ubuntu-latest]
     steps:
     - name: Checkout Sources
--- a/impl/pom.xml
+++ b/impl/pom.xml
@@ -223,6 +223,15 @@
             </plugins>
         </pluginManagement>
         <plugins>
+            <!-- Restricts the Java version to 17 - API remains at Java 17 for now. -->
+            <plugin>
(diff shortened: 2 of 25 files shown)


arjantijms:
PR will be updated with expressly M1 once it's available.

pzygielo:
https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#release - it's recognized since plugin 3.6, but apparently no explicit version of m-compiler-p is specified and it depends on maven used and falls-back to 3.1 here.

arjantijms:
> https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#release - it's recognized since plugin 3.6, but apparently no explicit version of m-compiler-p is specified and it depends on maven used and falls-back to 3.1 here.

I thought, but clearly wrongly, that the compiler version came from the EE4J pom. But seems I was mistaken.

pzygielo:
> I also thought this would set the correct version of the compiler of Maven 3.6.0, but also seemingly it does not actually do that. Probably need to upgrade the minimal Maven version a bit to that which has Maven compiler 3.6 at least.

That would be [maven 3.9.0](https://maven.apache.org/ref/3.9.0/maven-core/default-bindings.html), as all previous have m-c-p at 3.1.

Note: the [current GH runner ubuntu-latest](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md) provides maven 3.8.8 only.

Issue #5375, opened 2023-12-11
The jfwid contains the session ID


jasondlee:
The client window ID generated in ClientWindowImpl contains the session ID, which is not _needed_ for this functionality, and exposure of the session ID can be used to compromise security. This method should be modified, then, so as not to use the session.  PR incoming.

PR #5376, opened 2023-12-11
[MOJARRA_5375]  The jfwid contains the session ID

jasondlee:
Issue: https://github.com/eclipse-ee4j/mojarra/issues/5375

Changes:
- Modify ClientWindowImpl to generate window ID w/o using session id 
- Add TokenGenerator from MyFaces

### DIFF
--- a/impl/src/main/java/com/sun/faces/lifecycle/ClientWindowFactoryImpl.java
+++ b/impl/src/main/java/com/sun/faces/lifecycle/ClientWindowFactoryImpl.java
@@ -32,6 +32,7 @@ public class ClientWindowFactoryImpl extends ClientWindowFactory {
 
     private boolean isClientWindowEnabled = false;
     private WebConfiguration config = null;
+    private final TokenGenerator tokenGenerator = new TokenGenerator();
 
     public ClientWindowFactoryImpl() {
         super(null);
@@ -42,6 +43,16 @@ public class ClientWindowFactoryImpl extends ClientWindowFactory {
     public ClientWindowFactoryImpl(boolean ignored) {
         super(null);
         isClientWindowEnabled = false;
+    }
+
+
+    @Override
+    public ClientWindow getClientWindow(FacesContext context) {
+        if (!isClientWindowEnabled) {
+            return null;
+        }
+
+        return new ClientWindowImpl(tokenGenerator);
     }
 
     private class PostConstructApplicationListener implements SystemEventListener {
@@ -65,14 +76,5 @@ public class ClientWindowFactoryImpl extends ClientWindowFactory {
         String optionValue = config.getOptionValue(WebConfiguration.WebContextInitParameter.ClientWindowMode);
 
         isClientWindowEnabled = null != optionValue && "url".equals(optionValue);
-    }
-
-    @Override
-    public ClientWindow getClientWindow(FacesContext context) {
-        if (!isClientWindowEnabled) {
-            return null;
-        }
-
-        return new ClientWindowImpl();
     }
 }
--- a/impl/src/main/java/com/sun/faces/lifecycle/ClientWindowImpl.java
+++ b/impl/src/main/java/com/sun/faces/lifecycle/ClientWindowImpl.java
@@ -19,7 +19,6 @@ package com.sun.faces.lifecycle;
 import static com.sun.faces.renderkit.RenderKitUtils.PredefinedPostbackParameter.CLIENT_WINDOW_PARAM;
 
 import java.util.Map;
-
 import jakarta.faces.component.UINamingContainer;
 import jakarta.faces.context.ExternalContext;
 import jakarta.faces.context.FacesContext;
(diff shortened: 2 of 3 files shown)


impl/src/main/java/com/sun/faces/lifecycle/ClientWindowImpl.java
```diff
@@ -28,6 +28,7 @@
 
 public class ClientWindowImpl extends ClientWindow {
 
+    private final TokenGenerator tokenGenerator = new TokenGenerator();;
```
arjantijms, line 31: Tiny thing, but double semi colon here

impl/src/main/java/com/sun/faces/lifecycle/TokenGenerator.java
```diff
@@ -0,0 +1,80 @@
...
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ *
+ * @author lu4242
+ */
+class TokenGenerator
+{
```
arjantijms, line 31: Another small thing, but should probably be formatted like the rest of Mojarra (essentially Sun rules, with 4 spaces for tab)

impl/src/main/java/com/sun/faces/lifecycle/TokenGenerator.java
```diff
@@ -0,0 +1,80 @@
...
+        seed = new AtomicLong(generateSeed());
+    }
+
+    private long generateSeed()
+    {
+        SecureRandom rng;
+        try
+        {
+            // try SHA1 first
+            rng = SecureRandom.getInstance("SHA1PRNG");
```
codylerum, line 45: Would it make more sense here to just call getInstanceStrong so we use what is currently considered the strong instance based on whatever jvm is running?

impl/src/main/java/com/sun/faces/lifecycle/ClientWindowImpl.java
jasondlee: Fixed.

impl/src/main/java/com/sun/faces/lifecycle/TokenGenerator.java
jasondlee: Fixed. I should have noticed that, but I just copied and pasted. :)

impl/src/main/java/com/sun/faces/lifecycle/TokenGenerator.java
jasondlee: I can make that change. As I noted above, I just copied their code over but have not examined it for optimizations (assumptions ftw, I guess :). I've made the change.

impl/src/main/java/com/sun/faces/lifecycle/TokenGenerator.java
```diff
@@ -0,0 +1,66 @@
...
+        long nextToken = seed.incrementAndGet();
+
+        // convert using base 36 because it is a fast efficient subset of base-64
+        return Long.toString(nextToken, 36);
+    }
+
+    private long generateSeed() {
+        SecureRandom rng;
+        try {
+            rng = SecureRandom.getInstanceStrong();
```
bstansberry, line 52: AIUI this is going to be called per-request, so this is very expensive to recreate each request.

jasondlee: Good catch. Moving this up to the factory. Commit incoming.

impl/src/main/java/com/sun/faces/lifecycle/ClientWindowFactoryImpl.java
```diff
@@ -32,6 +32,7 @@ public class ClientWindowFactoryImpl extends ClientWindowFactory {
 
     private boolean isClientWindowEnabled = false;
     private WebConfiguration config = null;
+    private final TokenGenerator tokenGenerator= new TokenGenerator();
```
codylerum, line 35: missing a space here between the equals

impl/src/main/java/com/sun/faces/lifecycle/ClientWindowImpl.java
```diff
@@ -64,9 +65,9 @@ private String calculateClientWindow(FacesContext context) {
                 counter = Integer.valueOf(0);
             }
             char sep = UINamingContainer.getSeparatorChar(context);
-            id = extContext.getSessionId(true) + sep + +counter;
+            id = tokenGenerator.getNextToken() + sep + ++counter;
```
codylerum, line 68: Not super familiar with the client window, if we are doing a token + counter stored with the session shouldn't the token be stored with the session as well so that we have a consistent token per session?

Do we even need a token? or can it just increment like was done for conversation id? The client window id is only valid with the users session anyway right?

jasondlee: I'm not expert on this either, but the value can't be stored in the session, as there can be multiple client windows at any given time. The value is retrieved from the request in the decode() method, with this method only being called when the ID can't be retrieved from the request.

mnriem (changes requested):
Please make sure this can also be applied to 5.0 branch

jasondlee:
> Please make sure this can also be applied to 5.0 branch

I don't see any reason why this wouldn't apply cleanly to any 4.x or 5.x branch. Am I missing something?

arjantijms (approved):

BalusC:
This doesn't seem to be in 4.0?

Please keep in mind that master branch is the future 5.0 version and that the currently active version is 4.0. It's therefore much easier to just target the fix at 4.0, it can then be safely merged into 4.1 and then 5.0.

Issue #5380, opened 2023-12-16
Synchronizing on the session object could be ineffectual due to the servlet implentation not providing reference stability


VsevolodGolovanov:
Mojarra synchronizes on the session object in some places, but at least Undertow 2.0.15.Final returns a different instance (wrapper) from `HttpServletRequest#getSession` for every request in the same session.
A simple text search of "synchronized (ses" in mojarra-2.3.21-RELEASE sources yields:
```
\impl\src\main\java\com\sun\faces\application\view\ViewScopeContextManager.java
\impl\src\main\java\com\sun\faces\renderkit\ServerSideStateHelper.java
\test\servlet30-isolated\cactus\src\main\java\com\sun\faces\application\DeprStateManagerImpl.java
```

This is issue is a follow up to [this](https://github.com/eclipse-ee4j/mojarra/issues/3743#issuecomment-1843748056).

mnriem:
Can you create a PR for the suggested changes?

BalusC:
I prepped a PR for 4.0. 

Feel free to backport for 2.3.

PR #5393, opened 2024-02-10
Add session mutex

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/5380


jasondlee (approved):

mnriem (approved):
Approving, but would have loved to see a test. It is a bit weird that the HttpSession object changes and it feels like a hack.

BalusC:
Mojarra doesn't anymore have IT env (everything has been moved to Faces/TCK) and current UT env is super outdated with JUnit4/Easymock/Powermock, should become JUnit5/Mockito.

Issue #5396, opened 2024-02-10
PartialViewContextImpl should use same VisitContext to reset values as the one used to process the execute and render


BalusC:
https://github.com/jakartaee/faces/issues/1877

PartialViewContextImpl should use same VisitContext to reset values as the one used to process the execute and render

PR #5397, opened 2024-02-10
PartialViewContextImpl should use same VisitContext to reset values as the one used to process the execute and render

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/5396


BalusC:
cc: @tandraschko

tandraschko:
+1 but do we really need EXECUTE_LIFECYCLE? AFAICS its not used in our IMPLs but the MF javadoc says something like
```
     * Hint that indicates that the visit is being performed as part of
     * lifecycle phase execution and as such phase-specific actions
     * (initialization) may be taken.
```

if i reset something, i dont want to trigger "initialization" of something

tandraschko:
currently it's only used for partial processing the tree, i dont think we should use it for resetting components

PR #5395, opened 2024-02-10
[MOJARRA_5375]  The jfwid contains the session ID

BalusC:
Original PR for https://github.com/eclipse-ee4j/mojarra/issues/5375 was against master instead of 4.0 https://github.com/eclipse-ee4j/mojarra/pull/5376

So here's a cherrypick for 4.0.

PR #5402, opened 2024-02-18
Improved impl for https://github.com/eclipse-ee4j/mojarra/issues/5375

BalusC:
Original PR for https://github.com/eclipse-ee4j/mojarra/issues/5375 was against master instead of 4.0 https://github.com/eclipse-ee4j/mojarra/pull/5376. I cherrypicked it for 4.0 https://github.com/eclipse-ee4j/mojarra/pull/5395 but there was a license concern (and I accidentally upmerged with 4.1 instead of 4.0) so I declined it.

Here's a improved and license safe impl. I will update 4.0/5.0 accordingly once PR is approved.


mnriem (commented):
Is there a potential for clashing here?

BalusC:
Nope. It's now an UUID. Which is "too much" actually because it's session scoped. But that's OK for now and much easier to do.

PR #5405, opened 2024-02-20
No need to swallow exception at that point.

arjantijms:
There's an outer catch that will log it.


BalusC (approved):

pnicolucci (commented):
The outer catch is:

`            } catch (Throwable t) { // CNFE or ValidationException or any other
                LOGGER.fine("Unable to load Beans Validation");
            }`

While we're here in the code should we update that message to better reflect the current name of "Jakarta Validation"? Something like: "Unable to load Jakarta Validation"?

mnriem:
LGTM

PR #5406, opened 2024-02-22
Revert "Remove deprecated bindings field"

volosied:
This reverts commit d3aed2499c582c461676cece1a9f845fde77fe1b.

Spoke to Emily Jiang  and we cannot remove this field from 4.1 as it's a minor release.  If we do, it would need to be a 5.0 release and that's too much work at this point.

Linking to https://github.com/jakartaee/faces/issues/1725


BalusC (approved):

BalusC:
Don't forget to remove https://github.com/jakartaee/faces/blob/4.1/spec/src/main/asciidoc/ChangeLog.adoc?plain=1#L48

Issue #5417, opened 2024-03-15
Component default required attribute is evaluated differently with xhtml/java implementation


chaloma1:
**Describe the bug**

1) Component with required attribute and **default parameter specified in xhtml interface** definition ignores calling java methods in xhtml like #{true} or #{xxxController.isComponentRequired}. If the required attribute is set as **plain string "true", then it works**.

2) Component **with not defined default value** in xhtml works with both #{true}, #{xxxController.isComponentRequired} or plain string "true"

3) Component with overriden isRequired() method using "required" instead PropertyKeys.required string as in UIInput parent class (you can add default value here if attribute is not found or set). Works with #{true} or #{xxxController.isComponentRequired} **but not with plain string like "true"**.

**To Reproduce**

[primefaces-test-git.zip](https://github.com/eclipse-ee4j/mojarra/files/14615081/primefaces-test-git.zip)
See the attached sample. You can execute the sample with mvn jetty:run command and hit http://localhost:8080/primefaces-test to run the page.

**Expected behavior**

Implementation type should not matter (at least the required attribute within xhtml component interface) and all of those values should have the same required result.

In the sample project i also tried using non-component specific attribute for FirstComponent like "styleClass" which is working ok even with default value.

**Environment**

I used primefaces test project to reproduce the issue. Sample is using Mojarra 4.0.5 version.

melloware:
@chaloma1 does the same happen with MyFaces 4?  `mvn clean jetty:run -Pmyfaces40` ?

melloware:
Just to be clear its a bug in MyFaces too?

chaloma1:
Default set in xhtml is fixed in MyFaces -> so this seems like a **bug in Mojarra only**
Default set by overriding isRequired method in Component class works the same in MyFaces as in Mojarra (which means it does not recognize plain string value -> required="true" while inserting component to a page) -> **bug in Mojarra and MyFaces** as well

melloware:
OK can you open a MyFaces ticket too please: https://issues.apache.org/jira/projects/MYFACES/issues/MYFACES-4655?filter=allopenissues

chaloma1:
Sure, just need few days until new Jira account is created.

chaloma1:
Issue at MyFaces created https://issues.apache.org/jira/browse/MYFACES-4656.

BalusC:
Again the EL literal vs VE matter.

Work around for now, use `default="#{false}"` instead of `default="false"`.

[...]

PR #5418, opened 2024-03-16
#5417: remove "optimization" when composite attribute default is literal instead of VE, this will fail when the final value itself is VE

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/5417

~Hold on with proactively merging. I want to run TCK before merge just to ensure there are no corner cases which depend on this "optimization".~ TCK passed.

(sorry for imports being shuffled in the code diff, I reconfigured Eclipse to treat jakarta.* the same way as javax.*)


arjantijms (approved):

mnriem (commented):
Is there a performance penalty? Does it break backwards compatibility?

BalusC:
> Is there a performance penalty? Does it break backwards compatibility?

Theoretically not. We would otherwise have retrieved issue reports from people complaining that `default="#{foo}"` is significantly slower than `default="foo"`.

Wrt backwards compatiblity, the existing TCK has passed all tests on this branch. That's all I can say about this.

mnriem (approved):

PR #5435, opened 2024-04-07
Fix #5434: Inverse default config of autoCompleteOffOnViewState

### LINKED ISSUE #5434: hidden field autocomplete="off" is not anymore valid
Originally reported in Faces project: https://github.com/jakartaee/faces/issues/1912

As per latest HTML spec the autocomplete="off" is not anymore valid on hidden fields (e.g. view state, client window state).

It was originally introduced to work around Firefox specific behavior which was not observable on other browsers: https://github.com/eclipse-ee4j/mojarra/issues/1133

Reproducer:
1. Open Faces page with Ajax form (and inspect jakarta.faces.ViewState hidden input field and remember its value).
2. Perform Ajax submit.
3. Hard-Refresh page and inspect jakarta.faces.ViewState hidden input field. It should have a new value, not the one from the initially opened page.

This happened only in Firefox as per https://bugzilla.mozilla.org/show_bug.cgi?id=520561 and work around was to add `autocomplete="off"` to the hidden field.

Fast forward ~15 years, the specific attribute value of `off` is considered invalid as per Nu HTML validator https://validator.w3.org/nu/ as per recent changes in HTML spec https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill

We can turn off this work around via following context param:
```
<context-param>
    <param-name>com.sun.faces.autoCompleteOffOnViewState</param-name>
    <param-value>false</param-value>
</context-param>
```
The original problem is also not anymore reproducible in current Firefox versions (tested 124.0.2) so I think it's OK to make this config the default as per Mojarra 4.1. I.e. don't anymore render `autocomplete="off"` unless explicitly instructed as follows:
```
<context-param>
    <param-name>com.sun.faces.autoCompleteOffOnViewState</param-name>
    <param-value>true</param-value>
</context-param>
```

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/5434

### DIFF
--- a/impl/src/main/java/com/sun/faces/config/WebConfiguration.java
+++ b/impl/src/main/java/com/sun/faces/config/WebConfiguration.java
@@ -46,13 +46,6 @@ import javax.naming.Context;
 import javax.naming.InitialContext;
 import javax.naming.NamingException;
 
-import com.sun.faces.application.ApplicationAssociate;
-import com.sun.faces.application.view.FaceletViewHandlingStrategy;
-import com.sun.faces.facelets.util.Classpath;
-import com.sun.faces.lifecycle.HttpMethodRestrictionsPhaseListener;
-import com.sun.faces.util.FacesLogger;
-import com.sun.faces.util.Util;
-
 import jakarta.faces.FactoryFinder;
 import jakarta.faces.application.ProjectStage;
 import jakarta.faces.application.ResourceHandler;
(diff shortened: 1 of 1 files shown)


melloware:
I will make the same default change in MyFaces.

melloware (approved):

BalusC:
> Are you keeping a list of what is being changed?

Yes, https://github.com/eclipse-ee4j/mojarra/milestone/45?closed=1

PR #5439, opened 2024-04-10
4.0 Backport of some Faces Util optimizations

pizzi80:
Some optimizations from my fork ...

### DIFF
--- a/impl/src/main/java/com/sun/faces/util/Util.java
+++ b/impl/src/main/java/com/sun/faces/util/Util.java
@@ -61,6 +61,8 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.Spliterator;
+import java.util.Spliterators;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


impl/src/main/java/com/sun/faces/util/Util.java
```diff
@@ -1262,29 +1259,29 @@ public static String getNamingContainerPrefix(FacesContext context) {
...
         Integer counter = (Integer) contextAttrs.get(viewStateCounterKey);
         if (null == counter) {
-            counter = Integer.valueOf(0);
+            counter = 0;
         }
 
         char sep = UINamingContainer.getSeparatorChar(context);
         UIViewRoot root = context.getViewRoot();
-        result = root.getContainerClientId(context) + sep + ResponseStateManager.VIEW_STATE_PARAM + sep + +counter;
+        result = root.getContainerClientId(context) + sep + ResponseStateManager.VIEW_STATE_PARAM + sep + counter;
```
BalusC, line 1272: This changes the outcome.

impl/src/main/java/com/sun/faces/util/Util.java
```diff
@@ -760,44 +797,17 @@ public static StateManager getStateManager(FacesContext context) throws FacesExc
...
-            result = Void.TYPE;
-            break;
-        default:
-            if (type.indexOf('.') == -1) {
-                type = "java.lang." + type;
-            }
-            result = Util.loadClass(type, Void.TYPE);
-            break;
+    public static Class<?> getTypeFromString(String type) throws ClassNotFoundException {
+        if ( type == null ) throw new ClassNotFoundException("Type is required");
```
BalusC, line 801: This changes the previous behavior.

impl/src/main/java/com/sun/faces/util/Util.java
```diff
@@ -549,6 +559,31 @@ public static boolean isEmpty(String string) {
...
+        return string == null || string.isBlank();
+    }
+
+    /**
+     * @return null if the passed String is null or blank, s otherwise
+     */
+    public static String nullIfBlank(String s) {
+        return isBlank(s) ? null : s;
+    }
+
```
BalusC, line 586: This is not an optimization, this is an addition which is nowhere used.

impl/src/main/java/com/sun/faces/util/Util.java
pizzi80: switch over a null value raise a NPE before Java 21

so the previous result was `NullPointerException` which is imho less explicative...

impl/src/main/java/com/sun/faces/util/Util.java
pizzi80: ok

impl/src/main/java/com/sun/faces/util/Util.java
BalusC: Just change the message. You can use the utility method Objects.requireNonNull() for this.

impl/src/main/java/com/sun/faces/util/Util.java
BalusC: Just tested and confirmed, indeed you're right.

impl/src/main/java/com/sun/faces/util/Util.java
pizzi80: ok

PR #5446, opened 2024-05-18
Fix #5445 Dynamic actions not correctly processed

### LINKED ISSUE #5445: Dynamic actions not correctly processed
**Describe the bug**

We have an application that is havily based on programmatic changes to the view tree. In this case, a datatable of primefaces with a static column (== defined in xHTML) is extended with some programmatically created columns. The static column is additionally moved to another position. This is done by removing the column from the view and later on it is added again.

The static column looks like this:
```
<p:column id="col_static" ...>
    <p:commandLink ... />
</p:column>
```

On remove, the component itself **and ALL of it's children** are added to the list of dynamic actions as REMOVE. The very same components are added as ADD to the same list of dynamic actions.

The list looks like this:
```
// list of dynamic actions
[
    REMOVE: Column,
    REMOVE: CommandLink,
    ADD: Column
    ADD: CommandLink
]
```

On the next request (here AJAX) the view needs to be restored. All dynamic actions are processed. It first removes the column itself. No problem, the component could be found. But as the first child of the first component should be removed too, the logic failes in UIData (the one from faces).

```
java.lang.NumberFormatException: Trying to extract rowIndex from clientId 'frmDialogs:j_idt905:j_idt906' there is no numeric segment
	at jakarta.faces.impl@4.0.7//jakarta.faces.component.UIData.invokeOnComponent(UIData.java:953)
```
UIData tries a fallback to resolve the component. Therefore it tries to parse `j_idt906` as a row index, which is not possible. UIData is not prepared for this case. 

I tried to fix it by switching the order of how the elements are removed from view (bottom-up instead of top-down). This fixes the REMOVEs. But then it fails again for the ADDs on the same code. Reason is, that the logic tries to find the component on the view before it will be restored from the list `StateContext#getDynamicComponents()`. And the part "find the component on the view" uses again `UIData#invokeOnComponent` which then tries to parse e.g. `j_idt906` as a row index.

So the only possible solution seems to be to fix UIData to it ignores the case "no row index".
I compared it to MyFaces, they do it the same - they simply ignore that case. And according to the DOCs this is ok.
```
Returns false if no component with the given clientId is found.
```

**To Reproduce**

I will try to build a WAR for this, that is not related to our specific infrastructure.

**Expected behavior**

It should work without UIData throwing a `NumberFormatException`.

**Screenshots**

Stack on REMOVE (p:commandLink)
grafik

Stack on ADD
grafik

**Additional context**

Faces version 4.0.7 on WildFly 32

BalusC:
Fix #5445: problem was two-fold:
- UIData#invokeOnComponent shouldn't have skipped non-UIColumn kids
- A FacesException while trying to locate dynamic component should not have blocked the view build time

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/view/FaceletViewHandlingStrategy.java
+++ b/impl/src/main/java/com/sun/faces/application/view/FaceletViewHandlingStrategy.java
@@ -1685,11 +1685,21 @@ public class FaceletViewHandlingStrategy extends ViewHandlingStrategy {
      * @param context the Faces context.
      * @param clientId the client id of the component to find.
      */
-    private UIComponent locateComponentByClientId(final FacesContext context, final UIComponent parent, final String clientId) {
+    private UIComponent locateComponentByClientId(final FacesContext context, final UIComponent parent, final String clientId, final boolean dynamic) {
         final List<UIComponent> found = new ArrayList<>(1);
         UIComponent result = null;
 
-        parent.invokeOnComponent(context, clientId, (context1, target) -> found.add(target));
+        try {
+        	parent.invokeOnComponent(context, clientId, (context1, target) -> found.add(target));
+        } catch (FacesException e) {
+        	if (dynamic) {
+                LOGGER.log(FINE, e, () -> "Cannot find dynamic component " + clientId + " in " + parent.getClientId(context)
+		            	+ "; assuming it just doesn't exist anymore"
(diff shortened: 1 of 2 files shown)


BalusC:
I want to run TCK on this as well so wait with merge.

melloware (approved):
Nice!

BalusC:
TCK passed for me so it's OK to merge.

PR #5458, opened 2024-07-13
Fix 5457: revert removal of getBeanClass on Faces artifact producers

### LINKED ISSUE #5457: Revert #5157
**Describe the bug**

The issue #5157 changed a behaviour of Mojarra which was there since 10 years++ to a new version which breaks Apache OpenWebBeans and thus TomEE and other containers. Thus I'd kindly ask to revert to the old implementation.

See https://github.com/eclipse-ee4j/mojarra/issues/5157

**To Reproduce**

See https://issues.apache.org/jira/browse/TOMEE-4355 on how to reproduce

**Expected behavior**

The CDI spec is not very outspoken about what a custom Bean<T> should return for the ```getBeanClass()``` method. JBoss Weld and Apache OpenWebBeans use this information in subtly different ways. It also seems that something did change in behaviour in late 2022 in either Weld or Wildfly resulting in a visibility issue in Wildfly.

In section "3.5. Bean constructors" the _bean class_ is defined as the class the constructor is called for. If you e.g. have a custom ```Bean<UIViewRoot>``` then you want the constructor of UIViewRoot to be invoked, not the one of the Extension class registering that bean. In OWB this information is used to determine which proxy to be created for the contextual references of that bean. There are also other points in the spec which underline the OWB interpretation. Further note that it would be rather easy to store away the information about the CDI Extension which registers a bean with the various ```addBean()``` methods as the CDI container always exactly knows which CDI Extension.

3rd note: visibility in CDI was broken from the beginning for more complex cases like EAR deployments. This started with the famous CDI-18 and CDI-129 tickets and went on to this day. To fix this the spec would need to introduce a HierarchicBeanManager to correspond with the hierarchic ClassLoaders. I had a long discussion with Bill, Linda, Pete et al back then. Some findings made it to a blog post https://struberg.wordpress.com/2015/02/18/cdi-in-ears/
All this might not get enough traction as the importance of EARs mostly faded over the last decade.

Note that this is a well known topic in the CDI spec and has been discussed since the CDI-1.0 days with no decision in either direction.

BalusC:
Fix #5457: revert removal of getBeanClass on Faces artifact producers; adjust it to check if Weld is used and if so return CdiExtension.class

### DIFF
--- a/impl/src/main/java/com/sun/faces/cdi/ApplicationMapProducer.java
+++ b/impl/src/main/java/com/sun/faces/cdi/ApplicationMapProducer.java
@@ -20,6 +20,7 @@ import java.lang.reflect.Type;
 import java.util.Map;
 
 import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.inject.spi.BeanManager;
 import jakarta.faces.annotation.ApplicationMap;
 import jakarta.faces.context.ExternalContext;
 import jakarta.faces.context.FacesContext;
@@ -40,10 +41,13 @@ public class ApplicationMapProducer extends CdiProducer<Map<String, Object>> {
      */
     private static final long serialVersionUID = 1L;
 
-    public ApplicationMapProducer() {
-        super.name("applicationScope").scope(ApplicationScoped.class).qualifiers(ApplicationMap.Literal.INSTANCE)
-                .types(new ParameterizedTypeImpl(Map.class, new Type[] { String.class, Object.class }), Map.class, Object.class)
-                .create(e -> FacesContext.getCurrentInstance().getExternalContext().getApplicationMap());
+    public ApplicationMapProducer(BeanManager beanManager) {
+        super.name("applicationScope")
+    		.scope(ApplicationScoped.class)
+    		.qualifiers(ApplicationMap.Literal.INSTANCE)
+    		.beanClass(beanManager, Map.class)
+            .types(new ParameterizedTypeImpl(Map.class, new Type[] { String.class, Object.class }), Map.class, Object.class)
+            .create(e -> FacesContext.getCurrentInstance().getExternalContext().getApplicationMap());
     }
 
 }
--- a/impl/src/main/java/com/sun/faces/cdi/ApplicationMapProducer.java
+++ b/impl/src/main/java/com/sun/faces/cdi/ApplicationMapProducer.java
@@ -43,9 +43,9 @@ public class ApplicationMapProducer extends CdiProducer<Map<String, Object>> {
 
     public ApplicationMapProducer(BeanManager beanManager) {
         super.name("applicationScope")
-    		.scope(ApplicationScoped.class)
-    		.qualifiers(ApplicationMap.Literal.INSTANCE)
-    		.beanClass(beanManager, Map.class)
+            .scope(ApplicationScoped.class)
+            .qualifiers(ApplicationMap.Literal.INSTANCE)
+            .beanClass(beanManager, Map.class)
             .types(new ParameterizedTypeImpl(Map.class, new Type[] { String.class, Object.class }), Map.class, Object.class)
             .create(e -> FacesContext.getCurrentInstance().getExternalContext().getApplicationMap());
     }
--- a/impl/src/main/java/com/sun/faces/cdi/CdiExtension.java
+++ b/impl/src/main/java/com/sun/faces/cdi/CdiExtension.java
@@ -180,27 +180,27 @@ public class CdiExtension implements Extension {
         // but this is not detectable as ServletContext is not necessarily available at this moment.
 
         afterBeanDiscovery.addBean(new ApplicationProducer());
-        afterBeanDiscovery.addBean(new ApplicationMapProducer());
-        afterBeanDiscovery.addBean(new CompositeComponentProducer());
-        afterBeanDiscovery.addBean(new ComponentProducer());
-        afterBeanDiscovery.addBean(new FlashProducer());
-        afterBeanDiscovery.addBean(new FlowMapProducer());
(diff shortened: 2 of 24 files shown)


mnriem (approved):

arjantijms:
@manovotn just thinking out loud, but can't we do something at the spec level to replace this:

```java
  protected CdiProducer<T> beanClass(BeanManager beanManager, Class<?> beanClass) {
         if (CdiUtils.isWeld(beanManager)) {
             this.beanClass = CdiExtension.class; // See #5457 and #5157
         } else {
             this.beanClass = beanClass;
         }

         return this;
     }
```

Maybe something like

```
public interface Bean<T> ...

    public Class<?>[] getBeanClasses();
}
```

With [0] the bean class as you would intuitively think about (the class of the bean, literally the name)
And [1] the class of the extension in which Bean<T> is created.

I haven't given it much more though, but in that way we would be able to keep both Weld and other CDI implementations happy?

manovotn:
@Pandrex247 Weld only requires that for beans that use client proxy.
That does not seem to be the case from this PR - the class seems to be added as a `@Dependent` bean as no other scope is specified and hence w/o client proxy.
Otherwise the `SerializableClientProxy` call should not happen, see the proxy generation code [here](https://github.com/weld/core/blob/master/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java#L596-L607).

Any chance Paraya changes the scope of the producer bean class somehow?

Pandrex247:
I'd have to continue digging for that to confirm, but not ottomh.

The reproducer we've got is for a `@SessionScoped` bean injecting an `ExternalContext`

```
@Named
@SessionScoped
public class ExternalContextUser implements Serializable {

    private static final long serialVersionUID = 1L;

    @Inject
    private ExternalContext externalContext;

    public String getContextPath() {
        return "Request context path is: " + externalContext.getRequestContextPath();
   }

}
```

What I observed from my digging so far is that Payara will attempt persist/restore the session using Hazelcast, and it's during the restore action where the above error is hit. 
Snippet from the stack trace above:
```
...
2025-05-06 22:28:01 at org.apache.catalina.session.StandardSession.deserialize(StandardSession.java:1246)
2025-05-06 22:28:01 at org.apache.catalina.session.StoreBase.readSession(StoreBase.java:289)
2025-05-06 22:28:01 at org.glassfish.web.ha.session.management.ReplicationStore.getSession(ReplicationStore.java:566)
2025-05-06 22:28:01 at org.glassfish.web.ha.session.management.ReplicationStore.getSession(ReplicationStore.java:509)
2025-05-06 22:28:01 at org.glassfish.web.ha.session.management.ReplicationStore.loadFromBackingStore(ReplicationStore.java:415)
2025-05-06 22:28:01 at org.glassfish.web.ha.session.management.ReplicationStore.load(ReplicationStore.java:389)
2025-05-06 22:28:01 at org.apache.catalina.session.PersistentManagerBase.doSwapIn(PersistentManagerBase.java:1124)
...
```

Persisting to Hazelcast will always entail some serialisation. From the stack trace, I can't see anywhere obvious where Payara would be changing the scope of the bean.

Pandrex247:
Yes that's where I'm saying this PR breaks things - it replaced the no-args constructor for `ExternalContextProducer` (and many other producers) with constructors requiring a `BeanManager` object.

Unless I'm misunderstanding something, Payara is trying to deserialise the session from a high availability store (Hazelcast):
```
...
2025-05-06 22:28:01 at org.apache.catalina.session.StandardSession.deserialize(StandardSession.java:1246)
2025-05-06 22:28:01 at org.apache.catalina.session.StoreBase.readSession(StoreBase.java:289)
...
```
This fails as Weld attempts to reinstantiate the deserialised object with `SerializableClientProxy`, which requires a no-arg constructor:

```
2025-05-06 22:28:01 org.jboss.weld.exceptions.UnproxyableResolutionException: WELD-001435: Normal scoped bean class com.sun.faces.cdi.ExternalContextProducer is not proxyable because it has no no-args constructor - com.sun.faces.cdi.ExternalContextProducer@60658389.
2025-05-06 22:28:01 at org.jboss.weld.bean.proxy.DefaultProxyInstantiator.validateNoargConstructor(DefaultProxyInstantiator.java:50)
2025-05-06 22:28:01 at org.jboss.weld.util.Proxies.getUnproxyableClassException(Proxies.java:249)
2025-05-06 22:28:01 at org.jboss.weld.util.Proxies.getUnproxyableTypeException(Proxies.java:199)
2025-05-06 22:28:01 at org.jboss.weld.util.Proxies.getUnproxyableTypesExceptionInt(Proxies.java:211)
2025-05-06 22:28:01 at org.jboss.weld.util.Proxies.getUnproxyableTypesException(Proxies.java:190)
2025-05-06 22:28:01 at org.jboss.weld.bean.proxy.ClientProxyProvider.getClientProxy(ClientProxyProvider.java:213)
2025-05-06 22:28:01 at org.jboss.weld.bean.proxy.util.SerializableClientProxy.readResolve(SerializableClientProxy.java:62)
...
```

With this PR reverted, the issue is not encountered.

manovotn:
Ah, my initial comment was wrong, the bean is actually `@RequestScoped` ([code](https://github.com/eclipse-ee4j/mojarra/pull/5458/files#diff-a729c848622c284db50a56ae8f724f9bd263d6a8cda1eed4432982ea7c925ccfR42)) so the proxy requirement is indeed correct.
Right, it makes sense to fail here.

Note that this commit is just a revert of a different solution I originally proposed (https://github.com/eclipse-ee4j/mojarra/pull/5158) but that didn't fly with OWB impl if I recall correctly.
Short of going that way, the fix would be to add a no-args constructor for the purpose of proxy creation which is pretty standard CDI solution.

PR #5462, opened 2024-07-26
Move global EL resolvers from static variables to application variables

### LINKED ISSUE #5461: Static variables in ELUtils cause memory leaks and leaks of resolver cache between applications
**Describe the bug**

When Mojarra is used in an application server, its classes are loaded by the app server's classloader and are shared by all deployed applications. This can lead to at least to serious issues:

* Some of the static resolvers store references to the application classloader with soft references. If the application is undeployed from the server, these references can hang in the resolver cache for a long time, even until the heap is full. They don't generate OOM but may block the memory up to the heap size, even though the classloader is not used anymore. Details in this issue: https://github.com/jakartaee/expression-language/issues/214
* Since the resolvers are global and shared by all applications, different applications might clash, and one application could see data from another application. This is purely hypothetical, I didn't find any resolver that would allow that but still, sharing a global variable between multiple applications is not a good practice and is potentially risky

**To Reproduce**

The memory leak can be reproduced as discussed in https://github.com/jakartaee/expression-language/issues/214

**Expected behavior**

Each application that is deployed on an application server uses its own set of resolvers. Doesn't share them with other applications.

OndroMih:
Fixes #5461

This avoids sharing resolvers between multiple applications on the classpath and also avoids memory leaks in application servers after an application and its classloaders are undeploed.

Resolvers are tied to the Faces Application instance.

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/ApplicationAssociate.java
+++ b/impl/src/main/java/com/sun/faces/application/ApplicationAssociate.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Contributors to Eclipse Foundation.
+ * Copyright (c) 2021, 2024 Contributors to Eclipse Foundation.
  * Copyright (c) 1997, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -79,8 +79,8 @@ import com.sun.faces.facelets.util.DevTools;
 import com.sun.faces.facelets.util.FunctionLibrary;
 import com.sun.faces.spi.InjectionProvider;
 import com.sun.faces.util.FacesLogger;
-
 import jakarta.el.CompositeELResolver;
+
 import jakarta.el.ELResolver;
 import jakarta.el.ExpressionFactory;
 import jakarta.faces.FacesException;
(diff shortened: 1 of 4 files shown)


melloware (approved):

lprimak (approved):

BalusC:
Does this also happen in 4.0?

lprimak:
Yea this happens in 4.0

OndroMih:
Yes, it would be great to backport this PR also to 4.0

BalusC:
It's easier to create PR for 4.0. Once merged into 4.0 then it's just a matter of merging 4.0 into 4.1. The other way round is not possible without manually backporting.

OndroMih:
OK, I didn’t know. I’ll create a PR for 4.0 then

lprimak:
No need to recreate the PR. Just edit the title and it’ll have a picker to move your target branch. You can just change that to 4.0

pzygielo:
> No need to recreate the PR. Just edit the title and it’ll have a picker to move your target branch. You can just change that to 4.0

Wouldn't this create PR for 256 commits then: https://github.com/eclipse-ee4j/mojarra/compare/4.0...OmniFish-EE:mojarra:4.1-ondromih-remove-static-vars? (af97a354e8fbe35e6dd748f84d3ad5ddaa3aa2e2)

OndroMih:
I created a new PR as I didn't see your comment. I'll close this PR in favor of https://github.com/eclipse-ee4j/mojarra/pull/5479.

BalusC:
> No need to recreate the PR. Just edit the title and it’ll have a picker to move your target branch. You can just change that to 4.0

Yeah but the current PR was created based off 4.1 branch. So effectively you're also going to commit the diff between 4.0 and 4.1 back into 4.0. This is not permitted.

Issue #5464, opened 2024-07-26
HtmlUtils#writeAttribute must follow same rules as HtmlUtils#writeUnescapedTextForXML when writing for partial response


BalusC:
Reported by https://stackoverflow.com/q/78770904

Related issue in past https://github.com/eclipse-ee4j/mojarra/issues/4516

PR #5465, opened 2024-07-26
Fix XML escaping when writing attributes

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/5464

### DIFF
--- a/impl/src/main/java/com/sun/faces/util/HtmlUtils.java
+++ b/impl/src/main/java/com/sun/faces/util/HtmlUtils.java
@@ -324,7 +324,7 @@ public class HtmlUtils {
 
     }
 
-    static private boolean isAllowedXmlCharacter(int ch) {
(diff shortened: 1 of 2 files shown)


melloware (approved):

mnriem (approved):
Do we have any specific tests testing this change?

BalusC:
> Do we have any specific tests testing this change?

Integration test isn't anymore possible in Mojarra project since the split of TCK. Everything has been moved to Faces TCK project where adding new tests for the **current** version isn't anymore possible. You can only add tests for the next Faces version but any tests which would benefit current Faces version as well thus won't be launched. Unit test is however possible for this specific use case and we can fortunately still put them in Mojarra project. I'll look at it.

In long term I really wish to be able to run integration tests from Mojarra project again. We can always simply migrate these to the TCK when a new major/minor version is to be released.

BalusC:
> Do we have any specific tests testing this change?

I added an unit test.

Issue #5460, opened 2024-07-15
Revert or fix #5417 and release a new 4.0.x version


filipeferraz:
**Describe the bug**

The change implemented in jakarta.faces-4.0.7.jar broke reading the default attribute value in composite components with default value in xhtml template.

Reading UIComponentBase.getAttributes().get("attribute") returns null if the attribute is not redefined at the xhtml page with componente usage (always need to redefine, ignores the default value defined in custom componente template).

**To Reproduce**

1. Create a composite component with a basic property:
		<cc:attribute name="customRole" required="false" default="user" type="java.lang.String" />
2. Use in a xhtml page without configuring customRole.
3. Try to read in java class of the component this.getAttributes().get("customRole").

**Expected behavior**

It's expected to get the default value for the attribute.

**Screenshots**

**Desktop (please complete the following information):**

 - OS: Manjaro Linux
 - Browser: Chrome
 - Version: 126.0.6478.126

**Additional context**

Bug found using Wildfly 32.0.1.Final. Rolling back  to jakarta.faces-4.0.6.jar keeps the component working.

customComponent.xhtml:
```
<ui:component
	    xmlns="http://www.w3.org/1999/xhtml"
		xmlns:h="jakarta.faces.html"
		xmlns:ui="jakarta.faces.facelets"
	    xmlns:cc="jakarta.faces.composite"
		xmlns:f="jakarta.faces.core">
    <cc:interface componentType="customComponent">
		<cc:attribute name="customRole" required="false" default="user" type="java.lang.String" />
    </cc:interface>
    <cc:implementation>
        <h:outputText value="Role"/>
		<h:commandButton action="#{cc.action}" value="Action"/>
    </cc:implementation>
</ui:component>
```

CustomComponent.java:
```
import jakarta.faces.component.FacesComponent;
import jakarta.faces.component.UINamingContainer;

import java.util.ArrayList;
import java.util.List;

@FacesComponent("customComponent")
public class CustomComponent extends UINamingContainer {	

	public List<String> action() {
		System.out.println(getAttributes().get("customRole"));
		return new ArrayList<>();
	}
}
```

filipeferraz:
Tested with a compiled version of the branch **mojarra_issue_5460** and the fix works (replaced the Widlfly 32.0.1.Final module jar).

It's possible to include in 4.0.8 milestone?

Thanks @BalusC

PR #5469, opened 2024-08-03
Add minimal configuration for integration tests

BalusC:
Also added IT tests for #5460 and #5464 and spotted one more place where #5464 should be fixed.

NOTE: the technical reason for adding back the IT config to Mojarra project is because we CANNOT add new tests to the TCK of an already-released Faces version. We can only add them for the next Faces version. But this also means that these won't be executed when the current version is being patched/bugfixed. Hence we can only add them to the Mojarra project. Once a next TCK version is to be prepared we can simply move them from Mojarra into Faces project for TCK.

### DIFF
--- a/impl/src/main/java/com/sun/faces/util/HtmlUtils.java
+++ b/impl/src/main/java/com/sun/faces/util/HtmlUtils.java
@@ -128,6 +128,10 @@ public class HtmlUtils {
                 // UNICODE entities: encode as needed
                 nextIndex = _writeDecRef(out, buff, buffIndex, buffLength, ch);
             } else {
+                if (forXml && !isAllowedXmlCharacter(ch)) {
+                    return buffIndex;
+                }
+
                 nextIndex = addToBuffer(out, buff, buffIndex, buffLength, ch);
             }
         }
--- a/pom.xml
+++ b/pom.xml
@@ -51,6 +51,7 @@
         <module>impl</module>
         <module>action</module>
         <module>rest</module>
+        <module>test</module>
     </modules>
 
    <properties>
--- a/test/base/pom.xml
+++ b/test/base/pom.xml
@@ -1,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) Contributors to Eclipse Foundation.
+
+    This program and the accompanying materials are made available under the
+    terms of the Eclipse Public License v. 2.0, which is available at
+    http://www.eclipse.org/legal/epl-2.0.
+
+    This Source Code may also be made available under the following Secondary
+    Licenses when the conditions for such availability set forth in the
+    Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+    version 2 with the GNU Classpath Exception, which is available at
+    https://www.gnu.org/software/classpath/license.html.
+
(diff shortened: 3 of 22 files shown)


mnriem (commented):
Would you consider using Playwright instead of Selenium?

BalusC:
I wasn't ware of this possibility. I'm familiar with Playwright itself, but only in Node.js projects. I'll therefore need more time to figure out how to use it in Java projects.

pizzi80:
I tried Playwright with Java yesterday after reading this conversation... prepare 2 Gbyte of free space... it needs a whole Ubuntu 20.x + many other libraries 😲

BalusC:
Can we please move forward and decide whether to use Selenium or Playwright? I don't want being unable to add tests for Faces for a long time again. I've already been waiting almost one year for a working 5.0 TCK https://github.com/jakartaee/faces/pull/1852

mnriem:
I mentioned Playwright merely because I like the Trace Viewer. If Selenium has something as capable then sure go with Selenium

BalusC:
Selenium uses OpenTelemetry.

I'll move forward and merge nonetheless. The point is that any new tests should be able to run in TCK project as well without further modification. If we want to use PlayWright then TCK has to be migrated first.

mnriem:
Does OpenTelemetry with Selenium have the ability to step through the request like Trace Viewer?

BalusC:
I've never used it. I normally just use the IDE debugger.

pizzi80:
My 2 cents: 
"Keep it simple" ... Selenium it's a simple Java dependency and it works out of the box ... 
PlayWright seems like a microsoft product... don't know what will be in the next 10 years

mnriem:
And so is Playwright:

```
         <dependency> 
            <groupId>com.microsoft.playwright</groupId>
            <artifactId>playwright</artifactId>
            <version>${playwright.version}</version>
            <scope>test</scope>
        </dependency>
```

It is open-source just like Selenium

arjantijms:
I personally don't care much about using either Selenium or Playwright. At the time someone just started using Selenium, and it does the job AFAIK. 

We did had issues with Chrome/Chromium on the (Eclipse) CI, which were eventually resolved by using a custom pipeline based on the `eclipsecbijenkins/basic-ubuntu-chrome:latest` image. See https://ci.eclipse.org/jakartaee-tck/view/EFTL-Certification-Jobs-11/job/11/job/standalone-tck/job/jakarta-faces-tck-glassfish/configure

For Playwright to be used instead we have to see if it will run on that particular image and it doesn't use another image. We also have to rewrite the existing tests that use Selenium to use Playwright then.

I'm not deep enough in the actual frontend tests to be able to say how much work that will be, but obviously someone needs to do that. It's not a simple change of a dependency, is it?

pizzi80:
This PR is not only about test, 
I found that the original ticket was about special char problems
during ajax request 

https://stackoverflow.com/questions/78770904/browsers-behavior-utf-8-page-with-extra-character-like-ufffe-error-by-ajax-x


but, this fix has the downside that is no more possible to save 
a String with emoji because the following function is removing it from xml response:

`HtmlUtils.isAllowedXmlCharacter` <-- blocks emoji

pizzi80:
I found a fix on my fork

but it fails the following sub-test, 
I don't know what kind of special chars are they...

```
for (int c = 0xD800; c <= 0xDFFF; c++) {
  assertFalse(isAllowedXmlCharacter(c));
}
```

PR #5481, opened 2024-08-12
Add expression language 6.0 Optional and Record resolvers

### LINKED ISSUE #5480: Support new resolvers from Expression Language 6
Expression Language 6 added two new resolvers; the OptionalELResolver for Optional, and the RecordELResolver. These are not automatically used in Faces/Mojarra, since we build our own chain explicitly.

We should support them for 4.1.x, and for 5.0 probably add them as requirement in the spec.

See also https://github.com/eclipse-ee4j/glassfish/issues/25052

arjantijms:
Fixes #5480


arjantijms:
cc @tandraschko WDYT?

tandraschko:
+1
will also add it to MyFaces now

arjantijms:
Thanks @tandraschko

PR #5526, opened 2024-11-05
Handle empty mapping in normalizeRequestURI

### LINKED ISSUE #5525: Endless loop in MultiViewHandler.normalizeRequestURI
**Describe the bug**

When FacesServlet is mapped to `/*` or invoked via forwarding where other servlet has this mapping, the view processing never terminates as it is stuck in `while` loop in `MultiViewHandler.normalizeRequestURI`. `mapping` is `""` in this case and therefore loop condition is always true and the `viewId` never changes.

**To Reproduce**

1. Set mapping of `facesServlet` to `/*`
2. Make a request to a view
3. Observe infinite loop

**Expected behavior**

Request processing should terminate. Given that `/*` is rare but valid entry, it shouldn't try any normalization

**Additional context**

Will provide PR for version 4, master is not buildable at this time.

pdudits:
Fixes #5525.

I assume `mapping` is not `null`, but the assumption was in the previous version as well by means of calling `.startsWith(mapping)`.

Project builds, and the fix can be tested by trivial example where `facesServlet` would be mapped to `/*`.

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/view/MultiViewHandler.java
+++ b/impl/src/main/java/com/sun/faces/application/view/MultiViewHandler.java
@@ -458,6 +458,10 @@ public class MultiViewHandler extends ViewHandler {
      * @since 1.2
      */
     protected String normalizeRequestURI(String viewId, String mapping) {
+        if (mapping.isEmpty()) {
+            return viewId;
+        }
+        
         boolean logged = false;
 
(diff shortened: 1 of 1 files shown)


BalusC:
Excellent, but please recreate PR for 4.0 branch instead of master (5.0) branch.

Issue #5528, opened 2024-11-14
isErrorPage() check fails in Tomcat when thrown exception has null message


BalusC:
https://github.com/omnifaces/omnifaces/issues/865

Unsure about other servers.

Solution: check exception itself as well.

Spec: https://jakarta.ee/specifications/faces/4.0/jakarta-faces-4.0#a404

> Determine if this request is a postback or initial request by executing the following algorithm. Find the render-kit-id for the current request by calling calculateRenderKitId() on the Application’s ViewHandler. Get that RenderKit’s ResponseStateManager and call its isPostback() method, passing the current FacesContext. **If the current request is an attempt by the servlet container to display a servlet error page**, do not interpret the request as a postback, even if it is indeed a postback.

However neither Servlet nor Faces spec is sufficiently specific as to how exactly to detect that. The current impl only checks for the error message, which is in case of exceptions (status=500) the raw exception message. However, if the exception message itself is null, e.g. `throw new RuntimeException()` instead of `throw new RuntimeException("peek-a-boo")`, this check fails.

pizzi80:
Confirmed, also opened a ticket for Tomcat

https://bz.apache.org/bugzilla/show_bug.cgi?id=69444

pizzi80:
I can submit a PR if you want

pizzi80:
Tomcat will fix this in 

- 11.0.x for 11.0.2 onwards
- 10.1.x for 10.1.34 onwards
-  9.0.x for  9.0.98 onwards

PR #5529, opened 2024-11-16
isErrorPage() check in RestoreViewPhase fails when thrown exception has a null message

BalusC:
#5528

### DIFF
--- a/impl/src/main/java/com/sun/faces/lifecycle/RestoreViewPhase.java
+++ b/impl/src/main/java/com/sun/faces/lifecycle/RestoreViewPhase.java
@@ -42,10 +42,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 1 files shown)


BalusC:
Note: RequestMap#containsKey() doesn't work for 2 reasons:

1. Tomcat doesn't set request attribute at all.
2. Mojarra's RequestMap impl doesn't check for attribute name but for attribute value (and this bug is apparently present in many other internal Map implementations such as ApplicaitonMap, InitParameterMap, RequestHeaderMap, RequestHeaderValuesMap, RequestParameterMap, RequestParameterValuesMap and SessionMap -- this needs to be addressed via separate issue https://github.com/eclipse-ee4j/mojarra/issues/5530

mnriem (approved):
Are there any tests that validate all the (new) code paths?

BalusC:
Nope not yet.

Issue #5164, opened 2022-11-04
Not honoring default value of  jakarta.faces.FACELETS_BUFFER_SIZE (1024) 


volosied:
**Describe the bug**

The default buffer size is not set to 1024, as [specified by the specification. ](https://jakarta.ee/specifications/faces/4.0/jakarta-faces-4.0.html#a6088)
```
jakarta.faces.FACELETS_BUFFER_SIZE —The buffer size to set on the response when the ResponseWriter is generated. By default the value is 1024.
```

Tested with jakarta.faces-4.0.0.jar

**To Reproduce**

Test a page with: 
```
<h:form>
    jakarta.faces.FACELETS_BUFFER_SIZE : #{externalContext.getResponse().getBufferSize()}
</h:form>
```

**Expected behavior**

1024 should be set. 

**Screenshots**

**Desktop (please complete the following information):**

 - OS: Mac
 - Browser FireFox

**Additional context**

Tomcat 10.1.0:
jakarta.faces.FACELETS_BUFFER_SIZE : 8192  

Open Liberty (via facesContainer-4.0)
jakarta.faces.FACELETS_BUFFER_SIZE : 4096 

Glassfish 7:
jakarta.faces.FACELETS_BUFFER_SIZE : 8192 
(Interestingly, I had set FACELETS_BUFFER_SIZE to 1024 and 2048 here, but these values looks to be ignored...? Could there be a bug glassfish here...?)

volosied:
https://github.com/eclipse-ee4j/mojarra/blob/master/impl/src/main/java/com/sun/faces/application/view/FaceletViewHandlingStrategy.java#L893-L896

It only calls setResponseBufferSize when the buffer size is set via web.xml.  I think the `if (responseBufferSizeSet) {` should be removed. 


```
        try {
            responseBufferSizeSet = webConfig.isSet(FaceletsBufferSize);
            responseBufferSize = Integer.parseInt(webConfig.getOptionValue(FaceletsBufferSize));
        } catch (NumberFormatException nfe) {
            responseBufferSize = Integer.parseInt(FaceletsBufferSize.getDefaultValue());
        }
        
... later on... 

        if (responseBufferSizeSet) {
            // set the buffer for content
            extContext.setResponseBufferSize(responseBufferSize);
        }
```

BalusC:
Merged in 4.0 now.

BalusC:
[...]

PR #5532, opened 2024-11-20
Don't attempt to set response buffer size on a committed response

Pandrex247:
Changes made to fix https://github.com/eclipse-ee4j/mojarra/issues/5164 and https://github.com/eclipse-ee4j/mojarra/issues/5262 causes breakages in the Jakarta EE 10 Security TCK (specifically in the "old" bit).

```
[INFO]      [exec] [javatest.batch] FAILED........com/sun/ts/tests/securityapi/ham/customform/base/Client.java#testCustomFormHAMHasCorrectQualifier
[INFO]      [exec] [javatest.batch] FAILED........com/sun/ts/tests/securityapi/ham/customform/base/Client.java#testCustomFormHAMValidateRequest
[INFO]      [exec] [javatest.batch] FAILED........com/sun/ts/tests/securityapi/ham/customform/base/Client.java#testCustomFormLoginToContinueErrorPage
```

This is reproducible in Payara 6 and GlassFish 7 when using a Mojarra version higher than 4.0.0 (the first change was introduced in 4.0.1).

I'm not 100% certain if this is just a workaround for a dodgy flow, but it fixes the TCK.

The TCK fails because Mojarra is attempting to redirect from a login page after authentication and is attempting to set the response buffer size on the redirected request (which has been committed), causing the redirection to fail and the TCK to not find the correct content. If you refresh the page it _has_ been authenticated, so that's working, it just specifically fails at the point of redirection because Mojarra appears to be too optimistically trying to set the response buffer size.

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/view/FaceletViewHandlingStrategy.java
+++ b/impl/src/main/java/com/sun/faces/application/view/FaceletViewHandlingStrategy.java
@@ -898,7 +898,11 @@ public class FaceletViewHandlingStrategy extends ViewHandlingStrategy {
 
         // set the buffer for content, -1 indicates nothing should be set.
         if (responseBufferSize != -1) {
-            extContext.setResponseBufferSize(responseBufferSize);
+            if (!extContext.isResponseCommitted()) {
+                extContext.setResponseBufferSize(responseBufferSize);
+            } else {
+                LOGGER.warning("Skipping attempt to set buffer size on a committed response");
+            }
         }
 
         // get our content type


Pandrex247:
I am very open to changing the log level of the "skip" message, it may be better to have it at something like FINE.

pizzi80:
I've causal exceptions like this on my webapps, thanks

BalusC:
The existing checks all log WARN and this is OK as it signals something dodgy is going on and therefore makes the developer aware that the flow the developer was using should be fixed/adjusted/improved. E.g. sending redirect instead of forward and calling responseComplete on FacesContext.

BalusC (approved):

PR #5533, opened 2024-11-29
Handle empty mapping in normalizeRequestURI

pdudits:
Fixes https://github.com/eclipse-ee4j/mojarra/issues/5525.

I assume mapping is not null, but the assumption was in the previous version as well by means of calling .startsWith(mapping).

Project builds, and the fix can be tested by trivial example where facesServlet would be mapped to /*.


BalusC:
Problem reproduced and confirmed fixed with this.

PR #5540, opened 2024-12-26
Fix identity of BindingFacesMessage

BalusC:
Fix identity of BindingFacesMessage; it broke since introduction of equals/hashCode for FacesMessage in
https://github.com/jakartaee/faces/issues/1823

Reproducer: trigger 2 different bean validation errors on the same property, it should show 2 separate messages instead of only the last one.

Discovered while analyzing o:validateBean IT failure since switching from Mojarra 4.0.x to 4.1.x.


melloware (approved):

arjantijms (approved):

mnriem:
As you are fixing a bug I would have expected to see a test. Did I miss it?

BalusC:
Nope, there is no test.

I would also have added a test but the TCK situation was super frustrating at that moment.

PR #5546, opened 2025-01-23
Code cleanup, Patterns instead of operations direct on String

kamilkrzywanski:
Code cleanup, Patterns instead of operations direct on String
enhanced loops, breaks in loops, use primitives, toArray with [0] initializer

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/NavigationHandlerImpl.java
+++ b/impl/src/main/java/com/sun/faces/application/NavigationHandlerImpl.java
@@ -472,9 +472,7 @@ public class NavigationHandlerImpl extends NavigationHandler {
                 NavigationInfo info = new NavigationInfo();
                 if (!switches.isEmpty()) {
                     info.switches = new ConcurrentHashMap<>();
-                    for (Map.Entry<String, SwitchNode> cur : switches.entrySet()) {
-                        info.switches.put(cur.getKey(), cur.getValue());
-                    }
+                    info.switches.putAll(switches);
                 }
                 if (!navRules.isEmpty()) {
                     info.ruleSet = new NavigationMap();
--- a/impl/src/main/java/com/sun/faces/application/applicationimpl/ExpressionLanguage.java
+++ b/impl/src/main/java/com/sun/faces/application/applicationimpl/ExpressionLanguage.java
@@ -72,7 +72,7 @@ public class ExpressionLanguage {
      */
     public ELContextListener[] getELContextListeners() {
         if (!elContextListeners.isEmpty()) {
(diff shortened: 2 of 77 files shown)


pizzi80:
good job!

the toArray with [0] initializer to me it's not an optimization... 
After some investigation I've disable that suggestion on IntelliJ 

If you inspect the code of the `toArray(T[])` method you'll see that 
if you pass a correct sized Array it will use directly the `System.arraycopy` method
instead of the `Arrays.copyOf` method which is optimized by JVM but it's still 
less efficient of the native `System.arraycopy` call.

Probably at runtime the C2 compiler will optimize that but it requires some work,
so at best they will be on par but there is no advantage in first place if you know the size of the array

kamilkrzywanski:
@pizzi80 This change is indeed debatable in terms of its actual benefit. If you prefer, I can revert it.

BalusC:
Thank you, but frankly it's better to break down each different kind of mass-change in a separate PR for a quicker mind processing.

Right now this is basically a big PR with a lot of unrelated cosmetic changes which do not fix real world problems.

On such a big code base it's generally better to rewrite/cleanup/modernize the part when you *actually* need to work on that part.

BalusC:
And indeed, the `[0]` initializer is a bad idea. These days the `list.toArray(Type[]::new)` is the most efficient way.

pizzi80:
Sorry for the late reply, 
I've ported all the changes in my fork ( except the `toArray(...)` changes ) 
and **everythings is working fine**

During this porting I had the chance to improve a lot of other things,
but at the moment I can't create all the Pull requests 
(probably something like 100 / 150 PRs)

@BalusC which could be the best moment to create all those PR?


> On such a big code base it's generally better to rewrite/cleanup/modernize only the part when you actually need to work on that part.

true but questionable: a lot of these changes are very simple to do with IntelliJ, 
and imho are a long term investement to create a better and cleaner code base ...

(some classes are unreadable with a modern IDE, also the programming style 
is very error prone, probably derived from C programming style and Java 1.4 with no IDE support)

kamilkrzywanski:
+1 For me, the most important change was replacing String.replaceAll with precompiled patterns because it was annoying during debugging of my project. The rest of the changes I made along the way.

kamilkrzywanski:
@BalusC
The (Type[]::new) approach is not necessarily the best choice. It creates a new method reference that pollutes the metaspace, increases the class file size, and prolongs the warm-up phase, leading to worse performance while the code is still being interpreted.

BalusC:
Please break down in isolated PRs.

Or if it absolutely needs to be one giant PR, please post a screenshot wherein you show that the [entire TCK](https://github.com/jakartaee/faces/tree/5.0/tck) has successfully passed.

PR #5548, opened 2025-01-25
Impl jakartaee/faces#1972: log dev stage warning when prependId is used

BalusC:
https://github.com/jakartaee/faces/issues/1972


mnriem (approved):
Maybe even stronger to also include that it will be removed in Faces 6?

Issue #5562, opened 2025-02-13
Unicode characters are filtered out even though allowed by the spec


florian-dahlmanns-inform:
**Describe the bug**

When using a h:inputText component and entering a text with some unicode characters between 0x010000 and #x10FFFF that is transferred to the server, then this character does not get rendered anymore when updating the h:inputText component with the value from the backend.

**To Reproduce**

1. Create a view scoped backing bean with a property "text":
```java
@ViewScoped
@Named
public class InputBean implements Serializable {
  private static final long serialVersionUID = 1L;

  private String            text;

  public String getText() {
    return text;
  }

  public void setText(String text) {
    this.text = text;
  }
}
```
2. Create a XHTML file that defines the text component with a `valueChange` ajax event listener that updates the text component after changing the value:
```xml
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
   "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">

<html xmlns="http://www.w3.org/1999/xhtml"
      xmlns:c="jakarta.tags.core"
      xmlns:h="jakarta.faces.html"
      xmlns:f="jakarta.faces.core">
      
   <h:head>
      <title>Mojarra Test</title>
   </h:head>

   <h:body>
      <h:form id="form">
        <h:inputText  id="text" value="#{inputBean.text}">
          <f:ajax event="valueChange" execute="@this" render="@this"/> 
        </h:inputText>
      </h:form>
   </h:body>
</html>
```
3. Open the xhtml page
4. Enter "Hello 𝄞" (the G-Clef sign, unicode character U+1D11E, see https://www.compart.com/en/unicode/U+1D11E) and press TAB so the `changeValue` event fires
5. The inputText component now just shows "Hallo" without the special sign instead of the original input with the special sign

**Desktop**

 - OS: Windows 11
 - Browser: Chrome 133

**Additional context**

This problem can be seen when using Mojarra 4.0.9. It does *not* happen when using an older version like Mojarra 4.0.4.

The relevant code change was introduced in *Mojarra 4.0.8* with ticket https://github.com/eclipse-ee4j/mojarra/issues/5464 and pull-request https://github.com/eclipse-ee4j/mojarra/pull/5465/files, where the following code was added to `HtmlUtils`:

```java
    public static boolean isAllowedXmlCharacter(int ch) {
        // See https://www.w3.org/TR/xml/#charsets Character Range
[...]

PR #5564, opened 2025-02-16
Add unicode range #x10000-#x10FFFF

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/5562

### DIFF
--- a/impl/src/main/java/com/sun/faces/util/HtmlUtils.java
+++ b/impl/src/main/java/com/sun/faces/util/HtmlUtils.java
@@ -330,7 +330,7 @@ public class HtmlUtils {
 
     public static boolean isAllowedXmlCharacter(int ch) {
         // See https://www.w3.org/TR/xml/#charsets Character Range
-        return ch < 0x20 ? isPrintableControlChar(ch, true) : ch <= 0xD7FF || ch >= 0xE000 && ch <= 0xFFFD; 
+        return ch < 0x20 ? isPrintableControlChar(ch, true) : ch <= 0xD7FF || ch >= 0xE000 && ch <= 0xFFFD || ch >= 0x10000 && ch <= 0x10FFFF;
     }
 
     /**
--- a/impl/src/test/java/com/sun/faces/util/HtmlUtilsTest.java
+++ b/impl/src/test/java/com/sun/faces/util/HtmlUtilsTest.java
@@ -11,6 +11,7 @@ class HtmlUtilsTest {
     /**
      * https://github.com/eclipse-ee4j/mojarra/issues/4516
      * https://github.com/eclipse-ee4j/mojarra/issues/5464
+     * https://github.com/eclipse-ee4j/mojarra/issues/5562
      */
     @Test
     void testAllowedXmlCharacter() {
@@ -44,5 +45,6 @@ class HtmlUtilsTest {
 
         assertFalse(isAllowedXmlCharacter(0xFFFE));
         assertFalse(isAllowedXmlCharacter(0xFFFF));
+        assertTrue(isAllowedXmlCharacter(0x1D11E));
     }
 }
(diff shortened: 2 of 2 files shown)


melloware (approved):

pizzi80:
please merge this PR, so I can create a new PR based on this one :)

melloware:
Just chiming in here is what MyFaces has which has a few more?

```java
private static boolean isInvalidChar(int codePoint)
    {
        if (codePoint == 1113088)
        {
            return true;
        }

        if (codePoint == 0x9 || codePoint == 0xA || codePoint == 0xD) 
        {
            return false;
        }
        if (codePoint >= 0x20 && codePoint <= 0xD7FF) 
        {
            return false;
        }
        if (codePoint >= 0xE000 && codePoint <= 0xFFFD) 
        {
            return false;
        }
        if (codePoint >= 0x10000 && codePoint <= 0x10FFFF) 
        {
            return false;
        }

        return true;
    }
```

BalusC:
Not sure on 1113088, that's 0x10FC00 and it's inside the allowed range 0x10000-0x10FFFF.

The 0x9, 0xA and 0xD are in Mojarra already checked by isPrintableControlChar().

melloware:
Yep I have no idea why that `1113088` is allowed through as TRUE!  So it must have been some use case that was breaking with that.

melloware:
The commit for that was done with this ticket: https://issues.apache.org/jira/browse/MYFACES-4294

pizzi80:
I was waiting the merge to create a PR to fix the emoji check, now it's late in Italy and I'm fron mobile but basically the fix is:


Character.isHighSurrogate(c) || Character.isLowSurrogate(c)

pizzi80:
https://stackoverflow.com/questions/65693683/unicode-code-points-array-is-converted-to-string-and-back-with-changes

they are the boundary values of surrogate pair chars

melloware:
@pizzi80 why not just post your fixed solution here in the ticket so we can see. I think MyFaces and Mojarra should align their checking method.

pizzi80:
> @pizzi80 why not just post your fixed solution here in the ticket so we can see. I think MyFaces and Mojarra should align their checking method.

Absolutely, this is my version, that check also for emoji, 
but I'm not sure that cover also the incriminated char 1113088

the isEmoji() call can be merge with the other conditions...



```
public static boolean isAllowedXmlCharacter(int ch) {

      // See https://www.w3.org/TR/xml/#charsets Character Range
      boolean isAllowedXmlChar = ch < 0x20 ? isPrintableControlChar(ch, true) : ch <= 0xD7FF || ch >= 0xE000 && ch <= 0xFFFD || ch >= 0x10000 && ch <= 0x10FFFF;

      // if it's not an allowed char check if it's an emoji
      if ( !isAllowedXmlChar ) {
          isAllowedXmlChar = isEmoji((char)ch);
      }

      return isAllowedXmlChar;
  }

  /** @return true if it's an emoji char. an emoji is a char with high/low surrogate pair */
  public static boolean isEmoji( char c ) {
      return isHighSurrogate(c) || isLowSurrogate(c);
  }
```

pizzi80:
inside `HtmlUtilsTest` I've added this 2 check:

```
// rocket emoji (🚀)
for (char c : "\uD83D\uDE80".toCharArray()) {
    assertTrue(isAllowedXmlCharacter(c));
}

// polar bear emoji (🐻‍❄️)
for (char c : "\uD83D\uDC3B\u200D\u2744\uFE0F".toCharArray()) {
    assertTrue(isAllowedXmlCharacter(c));
}
```

BalusC:
The `isAllowedXmlCharacter(int ch)` method is wrong. It should have taken a `char ch` argument from the beginning on instead of `int ch`. This because it's never invoked with a `int codepoint` but with a literal `char ch`. I will create a new PR.

BalusC:
Superseded by https://github.com/eclipse-ee4j/mojarra/pull/5572

mnriem (commented):
LGTM

Issue #5552, opened 2025-02-05
[Question] Jakarta Faces 4.0.9 on multiwar-deployment in tomcat, possible NPE when a stale cookie exists


DGuhr:
porting from [here](https://github.com/jakartaee/faces/issues/2007) by request:

Hey. So, I spent the last days wrapping my head around an error, which I hopefully solved by now, but wanted to ask if this is potentially a bug in the internals, too, so here's an Issue. 

First of all, a bit about my context and setup: 
- I have little to no prior JSF experience.
- I have a legacy app and migrating it now to use jakarta faces 4.0.9 with primefaces 14.0.6 and spring framework 6.1.16 (not spring boot, then I would've used joinfaces, but migrating to boot wasn't in scope) on tomcat 10.0.
- I also have a login process that uses a login.xhtml in one .war - lets call it login.war - to do a "local" login  that uses a viewAction. Also, and another oauth.xhtml using another viewaction in another .war - let's call it gui.war - and this led me to some problems. 

The Problem:
Sometimes (not deterministic) after restarting my dockerized setup, I was not able to log in. The flow was like this:
First, the login.xhtml shows the credential form
Then, when submitting, an `f:viewaction` redirects to loginsuccess.xhtml which on successful login in turn redirects (via submitting a form using JS after adding params to the form, before you ask: I didn't write this, I'm just here to try to navigate the mess ;) ) to another oauth.xhtml, bundled in gui.war - which again logs in (multiple loginmodules, chosen at runtime, think oauth, saml, depends) and _should_ redirect in the end to a dashboard.xhtml 

Now this worked 98% of times, but sometimes when restarting the dockerized env, I got an JSF1094 error, that showed up for me with the error message 
```
[http-nio-8080-exec-10] com.sun.faces.context.flash.ELFlash$PreviousNextFlashInfoManager.decode JSF1094: Could not decode flash data from incoming cookie value null.  Processing will continue, but the flash is unavailable for this request.
```

I looked up this error, but could only find JSF1094 errors with other error messages, and talking about using myFaces as a solution which i declined, as this was seemingly too big a task, adding to the effort and would possibly introduce other sideeffects. 

[...]

PR #5565, opened 2025-02-16
Add clue to error log when flash is unavailable after server restart without session persistence

BalusC:
#5552


BalusC:
Crap, branched off from wrong branch. I will recreate PR.

PR #5566, opened 2025-02-16
Add clue to error log when flash is unavailable after server restart without session persistence

BalusC:
#5552


DGuhr (commented):
LGTM, if I may give my 2c as thread starter. One optional thing: 
Would it make sense to add a guard-clause to 
https://github.com/eclipse-ee4j/mojarra/blob/b7bc22a97b2b8b540089a87bdcb6055b97b5b6b9/impl/src/main/java/com/sun/faces/context/flash/ELFlash.java#L812
so there's no NPE thrown? I'm not very sure if there are other situations where this could happen, for me it was the JSF1094 though, so maybe throw that one, then, instead?

BalusC:
I added an unit test and sorted out the NPE.

impl/src/main/java/com/sun/faces/context/flash/ELFlash.java
```diff
@@ -156,7 +156,7 @@ public class ELFlash extends Flash {
      */
     static final String FLASH_NOW_REQUEST_KEY = FLASH_ATTRIBUTE_NAME + "n";
 
-    private enum CONSTANTS {
+    enum CONSTANTS {
```
mnriem, line 159: Why is private being removed?

BalusC: So it's accessible from ELFlashTest in same package.

mnriem: As I don't remember so I am just going to ask. Would this impact the signature tests?

BalusC: No idea. How do I run them so I can verify?

Issue #5579, opened 2025-04-15
Follow up on #5445: Dynamic actions not correctly processed


fanste:
In #5445 I already reported some bugs regarding dynamic actions processing in Mojarra.

A remaining issue is [this use case](https://github.com/eclipse-ee4j/mojarra/issues/5445#issuecomment-2129453852):
```xhtml
<h:form>
	<ui:repeat value="#{[1]}" var="x">
		<ui:fragment rendered="#{x eq 2}">
			<ui:repeat value="#{testBean.nonExistingGetter}">
			
			</ui:repeat>
		</ui:fragment>
	</ui:repeat>
	
	<h:dataTable binding="#{dt}">
		<h:column id="col">
			<h:outputText id="txt" value="test" />
		</h:column>
	</h:dataTable>

	<h:commandButton value="Remove Column" actionListener="#{testBean.doRemoveColumn(dt, 'col')}" />
	<h:commandButton value="Remove Column Reverse" actionListener="#{testBean.doRemoveColumnBottomUp(dt, 'col')}" />
</h:form>
```

```java
@RequestScoped
@Named
public class TestBean {
	public void doRemoveColumn(UIData dataTable, String columnIdFragment) {
		final UIComponent column = dataTable.findComponent(columnIdFragment);
		dataTable.getChildren().remove(column);
	}

	public void doRemoveColumnBottomUp(UIData dataTable, String columnIdFragment) {
		final UIComponent column = dataTable.findComponent(columnIdFragment);
		_removeBottomUp(column);
	}

	private void _removeBottomUp(UIComponent component) {
		for(UIComponent child : component.getChildren()) {
			_removeBottomUp(child);
		}

		component.getParent().getChildren().remove(component);
	}
}
```

`Remove Column` throws an exception, as the tree traversal (top-down) ignores `rendered`.
`Remove Column Reverse` works, as the removal is done bottom-up (manually).

This could be fixed by switching some lines of code in `UIComponentBase`:
```java
private static void disconnectFromView(FacesContext context, Application application, UIComponent component) {

    [...]

    application.publishEvent(context, PreRemoveFromViewEvent.class, component);
    component.setInView(false);
    component.compositeParent = null;
 }
```

[This was already done](https://github.com/eclipse-ee4j/mojarra/commit/6fb231aff3a36288c4851bc69fda561e58bbb909) by @BalusC but caused some new bugs. The new bug came up, as his fix only moved two of those three lines.

[...]

PR #5580, opened 2025-04-15
Improve timing of firing PreRemoveFromViewEvent wrt children

fanste:
#5579


fanste:
The test failes even on the 4.0 branch. So this is not related to my changes.

BalusC (approved):

PR #5581, opened 2025-04-17
Fix #5577: use autocomplete="one-time-code" by default on view state hidden field

### LINKED ISSUE #5577: Revist hidden field autocomplete="off"
> @BalusC  You said in Step 3: ` It should have a new value, not the one from the initially opened page.` 
> 
> I found that the new value is used when the page is refreshed in FireFox. However, in Safari / Chrome, the original value is displayed: 
> 
> Pure HTML/JS code to replicate
> ```
>       <form><input type=hidden name="test" value="original"></form>
>       <button id=show>show hidden value</button><button
>          id=update>Update value</button>
>       <script>document.getElementById("show").addEventListener("click", () => { alert(document.querySelector("input").value) }); document.getElementById("update").addEventListener("click", () => { document.querySelector("input").value = "updated" })</script>
> ```
> 
> I was also able to reproduce the behavior above via JSF (client side state saving and an input within a c:forEach).  The scenario I tested likely isn't very common, but it still proves that hidden field change is still possible.
> 
> I'm wondering which browsers are correct? Is it Firefox or the others.  This is a behavior difference, and I could still see someone affected by this difference in functionality between browsers.
> 
> I'm leaning to revert this change if it can somehow break an application/users, as the HTML validation error is less severe.  

 _Originally posted by @volosied in [#5434](https://github.com/eclipse-ee4j/mojarra/issues/5434#issuecomment-2790421405)_

BalusC:
https://github.com/eclipse-ee4j/mojarra/issues/5577

### DIFF
--- a/impl/src/main/java/com/sun/faces/renderkit/ClientSideStateHelper.java
+++ b/impl/src/main/java/com/sun/faces/renderkit/ClientSideStateHelper.java
@@ -41,16 +41,16 @@ import java.util.logging.Logger;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.GZIPOutputStream;
 
+import jakarta.faces.FacesException;
+import jakarta.faces.context.FacesContext;
+import jakarta.faces.context.ResponseWriter;
+
 import com.sun.faces.RIConstants;
 import com.sun.faces.config.WebConfiguration.BooleanWebContextInitParameter;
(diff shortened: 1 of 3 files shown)


melloware (approved):

mnriem:
Looks like this is change in behavior so should it not be targeted to 5.0?

BalusC:
It's not part of spec.

PR #5601, opened 2025-08-10
First step for jakartaee/faces#1590

BalusC:
First step for jakartaee/faces#1590: convert all Mojarra-generated inline event handlers to attached event handlers so they are at least CSP compatible

### DIFF
--- a/impl/src/main/java/com/sun/faces/facelets/tag/ui/UIDebug.java
+++ b/impl/src/main/java/com/sun/faces/facelets/tag/ui/UIDebug.java
@@ -22,15 +22,14 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
-import com.sun.faces.facelets.util.DevTools;
-import com.sun.faces.facelets.util.FastWriter;
-import com.sun.faces.renderkit.RenderKitUtils;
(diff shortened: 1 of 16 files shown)


BalusC:
Entire TCK passed, so undrafting PR.

jasondlee:
What are the chances of backporting this behavior to 4.x? It seems like the spec changes are just to codify support for a CSP, right? Can we (and by that I probably mean me ;) backport these changes to the 4.x branches?

BalusC:
TCK would need update as well because Issue2439IT would fail on this change.

@arjantijms would it be possible to update the TCK for 4.1 or perhaps even 4.0 so we can backport this?

jasondlee:
> TCK would need update as well because Issue2439IT would fail on this change (fixed for TCK 5.0 here: [jakartaee/faces#2055](https://github.com/jakartaee/faces/pull/2055) )
> 
> @arjantijms would it be possible to fix this as well in the TCK for 4.1 or perhaps even 4.0 so we can backport this?

FWIW, I was feeling ambitious last night, so I attempted to backport the changes. Short of one broken integration test, which may very well be environmental, I think they all applied fairly cleanly, if requiring fair bit of manual work. You can see the PR here: https://github.com/eclipse-ee4j/mojarra/pull/5606

Issue #5609, opened 2025-09-20
potential regression ( or change in behaviour) arround jakarta validation (@NotNull) in 4.1.4 vs 4.1.3


christophs78:
We first saw https://github.com/primefaces/primefaces/pull/14174 failing. (pushes Faces from 4.1.3 to 4.1.4 for PrimeFaces integration tests)

`Messages001Test.basic:56 expected: but was:`

I´m able to reproduce this regression/change localy with 4.1.4. too.

Please look at

- https://github.com/primefaces/primefaces/blob/bd0ebce10bf8989a3ca7b10d12d9b425f06c0887/primefaces-integration-tests/src/test/java/org/primefaces/integrationtests/messages/Messages001Test.java#L56
- https://github.com/primefaces/primefaces/blob/bd0ebce10bf8989a3ca7b10d12d9b425f06c0887/primefaces-integration-tests/src/main/java/org/primefaces/integrationtests/messages/Messages001.java#L46
- https://github.com/primefaces/primefaces/blob/bd0ebce10bf8989a3ca7b10d12d9b425f06c0887/primefaces-integration-tests/src/main/webapp/messages/messages001.xhtml#L24

`@NotEmpty` still work´s as expected.

Is this change in behaviour regarding `@NotNull` intended.

PR #5612, opened 2025-10-18
Regression on @NotNull

BalusC:
#5609


christophs78:
Thanks for taking the time and fixing this one!

PR #5623, opened 2025-12-19
account for timezone with java.time types in DateTimeConverter

### LINKED ISSUE #5622: timeZone not honored by DateTimeConverter for java.time types
For the new java.time types like zonedDateTime the DateTimeConverter when used like

`<f:convertDateTime type="zonedDateTime" pattern="MM/dd/yyyy hh:mm a z" timeZone="MST7MDT"  />`

It appears that the defined timezone is only set on the `DateFormat` and not on the `DateTimeFormatter` which is created for ZonedDateTime

https://github.com/eclipse-ee4j/mojarra/blob/4.0/impl/src/main/java/jakarta/faces/convert/DateTimeConverter.java#L554

https://github.com/eclipse-ee4j/mojarra/blob/4.0/impl/src/main/java/jakarta/faces/convert/DateTimeConverter.java#L456

Mojarra 4.0.12

codylerum:
closes #5622

### DIFF
--- a/impl/src/test/java/jakarta/faces/convert/DateTimeConverterTest.java
+++ b/impl/src/test/java/jakarta/faces/convert/DateTimeConverterTest.java
@@ -3,24 +3,34 @@ package jakarta.faces.convert;
 import com.sun.faces.junit.JUnitFacesTestCaseBase;
 import jakarta.faces.component.UIViewRoot;
 import jakarta.faces.component.html.HtmlOutputText;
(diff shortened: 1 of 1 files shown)


codylerum:
Looks like this is by design.

https://stackoverflow.com/questions/34883270/how-to-use-java-time-zoneddatetime-localdatetime-in-pcalendar#:~:text=Dealing%20with%20ZonedDateTime%20in%20JSF,explained%20in%20%231%20here%20above.

Issue #5538, opened 2024-12-18
Add EXCEPTION_TYPES_TO_IGNORE_IN_LOGGING context param


tandraschko:
We have it myfaces, omnifaces and primefaces, i think mojarra should have the same :)

Also lets add it to 5.0 specs?

BalusC:
+1

PR #5630, opened 2026-01-05
#5538: add com.sun.faces.exceptionTypesToIgnoreInLogging

BalusC:
#5538


mnriem (approved):
Code looks fine, but wondering why it is needed? Any specific reason?

PR #5712, opened 2026-04-20
Fix #5711: FlowMapProducer must be @RequestScoped not @FlowScoped

### LINKED ISSUE #5711: FlowMapProducer must be @RequestScoped not @FlowScoped
### Summary

`org.glassfish.mojarra.cdi.FlowMapProducer` declares `.scope(FlowScoped.class)`, which makes the produced `Map` a `@FlowScoped` CDI bean. Weld requires the producer's declared scope context to be active *even to invoke the producer method*, so EL evaluation of `#{flowScope}` throws `WELD-001303: No active contexts for scope type jakarta.faces.flow.FlowScoped` in cases where the FlowScoped context is not currently active.

### Spec behavior

`jakarta.faces.flow.FlowHandler#getCurrentFlowScope` javadoc:

> Return the `Map` that backs the `#{flowScope}` Jakarta Expression Language implicit object or `null` if no flow is currently active.

So `#{flowScope}` must resolve to `null` outside an active flow — it must not throw.

### How to reproduce

Any view that renders `#{flowScope.value}` (or any `#{flowScope.something}` expression as Facelet text content) and that is rendered outside an active flow, e.g. a flow-return page. Minimal reproducer:

```xhtml
<!-- return1.xhtml (rendered after flow exits) -->
<p>flowScope value, should be empty: #{flowScope.value}</p>
```

Result: HTTP 500 with stack trace starting at `AbstractWeldELResolver.lookup` → `WELD-001303: No active contexts for scope type jakarta.faces.flow.FlowScoped`.

This is also observable in the Jakarta Faces TCK `flows/basicmethodcall`, `flows/basicmultipage` and `flows/multipagewebinf` tests, which fail for exactly this reason.

### Proposed fix

Change the producer scope from `FlowScoped` to `RequestScoped`:

```java
super.name("flowScope")
    .scope(RequestScoped.class)   // was: FlowScoped.class
    .qualifiers(FlowMap.Literal.INSTANCE)
    …
```

`@RequestScoped` matches the sibling `FlowProducer` (which exposes `#{flow}`) and every other implicit-object producer in the `org.glassfish.mojarra.cdi` package (`FacesContextProducer`, `ExternalContextProducer`, `FlashProducer`, `RequestMapProducer`, `SessionMapProducer`, `HeaderMapProducer`, `ViewMapProducer`, …). The producer simply delegates to `FlowHandler.getCurrentFlowScope()`, which returns `null` when no flow is active and the current flow's `Map` otherwise — correct behavior in both states.

BalusC:
#5711

### DIFF
--- a/impl/src/main/java/org/glassfish/mojarra/cdi/FlowMapProducer.java
+++ b/impl/src/main/java/org/glassfish/mojarra/cdi/FlowMapProducer.java
@@ -19,12 +19,12 @@ package org.glassfish.mojarra.cdi;
 import java.lang.reflect.Type;
 import java.util.Map;
 
+import jakarta.enterprise.context.RequestScoped;
 import jakarta.enterprise.inject.spi.BeanManager;
 import jakarta.faces.annotation.FlowMap;
 import jakarta.faces.application.Application;
 import jakarta.faces.context.FacesContext;
 import jakarta.faces.flow.FlowHandler;
-import jakarta.faces.flow.FlowScoped;
 
 /**
  * <p class="changed_added_2_3">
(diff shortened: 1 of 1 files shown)


BalusC:
Merging so we can move forward. The TCK PR which depends on current PR has been merged. But it would still fail without current PR.

PR #5722, opened 2026-05-02
Fix #5721 prefer extension fallback for exact mappings

### LINKED ISSUE #5721: Exact-mapped Faces requests can choose the wrong wildcard fallback when multiple Faces servlet mappings exist
**Describe the bug**

When a request reaches Faces through an exact servlet mapping, Mojarra falls back to a wildcard Faces servlet mapping when it needs to generate action URLs or resource URLs.

  Today that fallback effectively depends on the first wildcard mapping Mojarra finds. If an application has both an extension mapping and a path mapping, that can produce the wrong generated URL.

  Example setup:

  - exact mapping for the current request
  - `*.jsf`
  - `/faces/*`

  Observed behavior:

  - for an exact-mapped request, Mojarra can generate fallback URLs using the path mapping instead of the extension mapping
  - this affects both view URLs and resource URLs

  Why this is a problem:

  - generated URLs depend on mapping declaration order
  - behavior differs from what the same application gets when the active request is extension mapped
  - this showed up in exact-mapping TCK coverage for generated view/resource paths

**Expected behavior:**

  - when the current request is exact mapped and Mojarra needs a wildcard fallback, prefer an extension mapping if one exists
  - if no extension mapping exists, fall back to the path mapping
  - if no wildcard Faces servlet mapping exists at all, keep the current failure behavior

**Additional context**

This issue was observed on both the 4.1.2 and the latest 4.1.8-SNAPSHOT versions.

robertpatrick:
Fixes #5721.

  ## Problem

  When the current request is exact mapped, Mojarra falls back to a wildcard Faces servlet mapping when it needs to generate action URLs or resource URLs.

  Today that fallback effectively depends on the first wildcard mapping Mojarra finds. If an application defines both an extension mapping and a path mapping, that can generate the wrong fallback URL based on declaration order.

  ## Change

  When the current request is exact mapped and Mojarra needs a wildcard fallback:

  - prefer an extension mapping if one exists
  - otherwise keep the existing path-mapping fallback
  - keep the existing failure behavior when no wildcard fallback exists

  ## Why

  This keeps exact-mapped fallback behavior aligned with extension-mapped Faces URL generation instead of letting declaration order choose the wrong wildcard mapping.

  ## Files

  - `impl/src/main/java/com/sun/faces/util/Util.java`
  - `impl/src/test/java/com/sun/faces/application/view/MultiViewHandlerTest.java`
  - `impl/src/test/java/com/sun/faces/application/resource/ResourceImplTest.java`
  - `impl/src/test/java/com/sun/faces/mock/MockFacesMappingSupport.java`

  ## Validation

  Added unit coverage for:

  - exact request + extension fallback
  - exact request + path fallback
  - exact request + both wildcard mappings present
  - exact request + no wildcard fallback
  - resource URL query parameter stability

  This is also the behavior that unblocked the exact-mapping Faces TCK case we were debugging:

  - `ee.jakarta.tck.faces.test.servlet40.exactmapping_selenium.Spec1260IT`

  ## Scope

  This PR is intentionally limited to exact-mapping fallback selection.

### DIFF
--- a/impl/src/main/java/com/sun/faces/util/Util.java
+++ b/impl/src/main/java/com/sun/faces/util/Util.java
@@ -1081,35 +1081,52 @@ public class Util {
         // If needed, cache this after initialization of Faces
         Object context = externalContext.getContext();
         if (context instanceof ServletContext) {
-            return getFacesServletMappings((ServletContext) context).stream()
-                    .filter(mapping -> mapping.contains("*"))
-                    .map(mapping -> new HttpServletMapping() {
+            HttpServletMapping pathMapping = null;
 
-                        @Override
-                        public String getServletName() {
-                            return "";
-                        }
+            for (String mapping : getFacesServletMappings((ServletContext) context)) {
+                if (!mapping.contains("*")) {
+                    continue;
+                }
 
-                        @Override
-                        public String getPattern() {
-                            return mapping;
-                        }
+                HttpServletMapping wildcardMapping = toWildcardMapping(mapping);
+                if (wildcardMapping.getMappingMatch() == MappingMatch.EXTENSION) {
+                    return wildcardMapping;
+                }
 
-                        @Override
-                        public String getMatchValue() {
-                            return null;
-                        }
+                if (pathMapping == null) {
+                    pathMapping = wildcardMapping;
+                }
+            }
 
-                        @Override
-                        public MappingMatch getMappingMatch() {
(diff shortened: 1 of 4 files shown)


BalusC (approved):

BalusC:
Crap. 4.1.8 was the wrong target branch, it should have been 4.1.

No problem, but we need to record here that this is NOT part of the 4.1.8 release but of the upcoming 4.1.9 release.

See also https://github.com/eclipse-ee4j/mojarra/pull/5725

PR #5724, opened 2026-05-02
Fix #5723 restore inline behavior handlers when CSP is not active

### LINKED ISSUE #5723: Non-CSP command components lose inline onclick behavior after CSP listener changes
**Describe the bug**

A recent change in behavior listener rendering appears to regress non-CSP command component output.

When CSP nonce support is **not** active, Mojarra should still render the inline `onclick` hooks used by command components such as:

- `h:commandLink`
- `h:commandButton` with `f:ajax`

Instead, behavior attachment can move to the CSP-style listener path (`mojarra.ael(...)`) even when no CSP nonce is active, which suppresses the inline `onclick` markup.

## Observed behavior

Without CSP nonce support:

- `h:commandLink` can lose its inline `onclick`
- Ajax command buttons can lose their inline `onclick`
- rendered markup changes in a way older clients/tests do not expect
- click behavior can also regress

## Expected behavior

- when CSP nonce support is active, keep the CSP-safe behavior-script path
- when CSP nonce support is not active, preserve inline `onclick` rendering for these command components

## To Reproduce

I have a standalone reproducer in a local `test/issue5606` module that verifies:

- commandLink renders inline `onclick` without CSP
- commandLink click still works without CSP
- ajax command button renders inline `onclick` without CSP
- ajax command button click still works without CSP

The reproducer page is essentially:

- `h:commandLink`
- `h:commandButton` with `f:ajax`

with assertions against both:

- rendered markup
- actual click behavior

## Suspected area

The regression appears to be in the behavior-listener rendering path around:

- `RenderKitUtils`
- `CommandLinkRenderer`
- `ButtonRenderer`

Specifically, CSP-style listener rendering seems to be applied too broadly instead of being conditional on CSP nonce support being active.

## Notes

This is separate from the exact-mapping fallback issue fixed under `#5721`.

**Additional context**

This problem was discovered when upgrading from 4.1.2 to the latest 4.1.8-SNAPSHOT version while testing the Fix for Issue #5721 in PR #5722.

robertpatrick:
• Fixes #5723.

## Problem

Recent behavior-listener rendering changes regressed non-CSP command component output.

When CSP nonce support is not active, Mojarra should still render the inline `onclick` hooks used by command components such as:

- `h:commandLink`
- `h:commandButton` with `f:ajax`

Instead, behavior attachment can move to the CSP-style listener path even when no CSP nonce is active, which suppresses the inline `onclick` markup.

## Change

- keep CSP behavior-script rendering when a nonce is active
- restore inline `onclick` rendering when CSP nonce support is not active

## Why

This preserves the CSP-safe listener model for nonce-enabled pages without regressing the existing non-CSP command component behavior.

## Files

- `impl/src/main/java/com/sun/faces/renderkit/RenderKitUtils.java`
- `impl/src/main/java/com/sun/faces/renderkit/html_basic/ButtonRenderer.java`
- `impl/src/main/java/com/sun/faces/renderkit/html_basic/CommandLinkRenderer.java`
- `test/pom.xml`
- `test/issue5606/**`

## Validation

Added `test/issue5606` integration coverage for:

- commandLink renders inline `onclick` without CSP
- commandLink click still works without CSP
- ajax command button renders inline `onclick` without CSP
- ajax command button click still works without CSP

Existing CSP coverage remains separate in `test/csp`.

With #5722 and this fix, WebLogic Server was able to pass the latest Faces TCK 4.1.2

## Scope

This PR is intentionally limited to the non-CSP renderer regression.

It does not include the separate exact-mapping fallback fix from #5721.

### DIFF
--- a/impl/src/main/java/com/sun/faces/renderkit/RenderKitUtils.java
+++ b/impl/src/main/java/com/sun/faces/renderkit/RenderKitUtils.java
@@ -425,7 +425,8 @@ public class RenderKitUtils {
             ? Collections.singletonList(new ClientBehaviorContext.Parameter("incExec", true))
             : Collections.emptyList();
 
-        addBehaviorEventListener(context, component, clientId, params, handlerName, userHandler, behaviorEventName, domEventName, null, false, incExec, true);
+        addBehaviorEventListener(context, component, clientId, params, handlerName, userHandler, behaviorEventName, domEventName, null, false, incExec,
+                ResourceHandlerImpl.resolveCurrentNonce(context) != null, true);
     }
 
     // Renders the onclick event listener for command buttons. Handles
@@ -451,7 +452,8 @@ public class RenderKitUtils {
             }
         }
 
-        addBehaviorEventListener(context, component, null, params, handlerName, userHandler, behaviorEventName, domEventName, submitTarget, needsSubmit, false, true);
+        addBehaviorEventListener(context, component, null, params, handlerName, userHandler, behaviorEventName, domEventName, submitTarget, needsSubmit, false,
+                ResourceHandlerImpl.resolveCurrentNonce(context) != null, true);
 
     }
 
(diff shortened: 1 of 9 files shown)


BalusC:
Please rebase against the real 4.1 branch. See also bottom of https://github.com/eclipse-ee4j/mojarra/blob/master/DEVELOPERS.md for PR-acceptable branches.

Issue #5506, opened 2024-09-17
Faces 5.0: we need to split faces-uncompressed.js in interface and implementation


BalusC:
That can only be done if we migrate to TypeScript. Then we can put interface ts in Faces project and keep implementation ts here in Mojarra project. Then we can finally run jsdoc from Faces project.

Related: 
- https://github.com/jakartaee/faces/issues/1598
- https://github.com/eclipse-ee4j/mojarra/issues/5398

PR #5728, opened 2026-05-03
Migrate faces-uncompressed.js to TypeScript

BalusC:
#5506

Primarily in order to be able to wire [faces.d.ts](https://github.com/jakartaee/faces/blob/5.0/api/src/main/resources/META-INF/resources/jakarta.faces/faces.d.ts) typecheck (as the "interface") so any future (accidental) API-deviations are caught at compiletime.

Also better long-term maintainability than a monolithic JS file.

All 363 faces.js unit tests in src/test/ts are unchanged and pass.

NOTE: tsconfig.ts is temporarily non-strict. Will be addressed in separate PR: https://github.com/eclipse-ee4j/mojarra/pull/5729


BalusC:
no review?

merging anyway so we can move forward

comment in #5506 if you want to push back

PR #5729, opened 2026-05-03
Tighten types and remove IE-specific cruft in faces.js

BalusC:
#5506 and jakartaee/faces#1598

Goal: enforce strict tsconfig:
```
"strict": true,
"noImplicitAny": true,
"strictNullChecks": true,
```
And remove now unnecessary IE-specific fluff (4.x compiled for ES5, 5.0 compiles for ES2020 which won't run in any IE in first place).

Requires https://github.com/jakartaee/faces/pull/2155 in build path in order to build/test.

1 test has been dropped (IE fallback in error checking in faces.push) and all 362 original tests on faces.js pass; then 24 new tests have been added to tighten the test coverage.


BalusC:
merging without review so we can move forward

comment in #5506 if you want to push back

PR #5754, opened 2026-05-27
Reduce component tree allocations + cache resolved Renderer to speed up Restore View / Render Response

BalusC:
#5753

## Summary

MyFaces is measurably faster than Mojarra during Restore View and Render Response. This PR closes most of the gap by attacking six hot spots in `UIComponentBase`, `ComponentStateHelper`, `UIData`, `UIRepeat`, and `Util.checkIdUniqueness`. All changes are behavior-preserving for spec-compliant applications.

The work is split into five commits (plus the perf harness commit), each independently reviewable. Each commit contains an inline ns/op delta in its message so the contribution of each phase is auditable.

### Phase A — `UIComponentBase`: cache resolved Renderer; drop `isRendered()` String round-trip

- `UIComponentBase.isRendered()` previously did `Boolean.valueOf(getStateHelper().eval(rendered, TRUE).toString())` — one `String` allocation per call. Replaced with direct `(Boolean)` cast, matching MyFaces.
- Added a `transient Renderer cachedRenderer` field, populated lazily by `getRenderer(FacesContext)`. Invalidated on `setRendererType`, `setParent`, and `restoreState` (the last is defensive: state restore writes `rendererType` through `StateHelper` bypassing the setter).
- `UIComponent.encodeAll` calls `encodeBegin/encodeChildren/encodeEnd`, each of which independently resolved the Renderer in Mojarra. The cache collapses to one lookup + two field reads per node per render.

**Measured impact** (5 × 1000 iters, JDK 11 + MockFacesContext):

| Scenario | encodeAll baseline | encodeAll Phase A | Δ |
|---|---:|---:|---:|
| flat tree (100 UIOutput) | 77.9 μs | 17.9 μs | **−77%** |
| flat tree (500 UIOutput) | 366.3 μs | 72.8 μs | **−80%** |
| form (50 UIInput) | 39.5 μs | 9.1 μs | **−77%** |
| form (200 UIInput) | 158.5 μs | 63.8 μs | **−60%** |
| deep panels (5×10) | 42.2 μs | 11.4 μs | **−73%** |
| UIData (20×3) | 4.7 μs | 1.7 μs | **−64%** |
| UIData (100×5) | 6.6 μs | 2.3 μs | **−65%** |

Added `UIComponentBaseTestCase#testCachedRendererInvariants` verifying cache invalidation on `setRendererType`, `setParent`, and `restoreState`.

**Behavior change**: `isRendered()` now throws `ClassCastException` if the `rendered` attribute is bound to a ValueExpression that returns a non-Boolean value; the old code silently coerced anything to `false`. Spec-defensible (the property is declared `boolean`); matches MyFaces.

### Phase B — `ComponentStateHelper`: lazy `deltaMap`

The constructor previously allocated two 16-bucket HashMaps eagerly (`deltaMap` + `defaultMap`). `deltaMap` is meaningless until the component is `initialStateMarked()` AND a delta is written; for typical render-only views with no model changes, no delta ever accrues, so the eager allocation was pure overhead.

Now `deltaMap` stays null until first write via a private `deltaMap()` accessor (`new HashMap<>(2)`). Read sites treat null as empty. `saveState` short-circuits to null when the component is initialStateMarked and `deltaMap` is null.

Construction-time win — eliminates one HashMap per UIComponent instance. Not visible in per-operation harness measurements but real heap-pressure relief on view-build. The UIData setRowIndex cycle is a per-operation case where it does show: ~30% faster than Phase A.

### Phase C — `UIData`: skip per-row state save (only) on read-only tables; setRowIndex no-op short-circuit

Two optimizations to the per-row save/restore tree walk:

1. `setRowIndexWithoutRowStatePreserved`: if the new rowIndex equals `getRowIndex()`, return immediately. The rowStatePreserved path already had this; the more common non-preserved path didn't.
2. Lazy `hasStatefulDescendants` flag. When false (the subtree contains no `EditableValueHolder` or `UIForm`), `saveDescendantState()` short-circuits without walking the tree. `restoreDescendantState()` **always** recurses — in addition to restoring EVH/UIForm state, it calls `setId(getId())` on every descendant to invalidate cached clientIds so each row produces a row-specific clientId. Skipping the restore-side walk would leave stale clientIds and render duplicate `id="..."` attributes across rows.

| Scenario | baseline | Phase C |
|---|---:|---:|
| UIData setRowIndex cycle (100 rows, UIInput) | 294 μs | ~177 μs (within noise vs Phase B) |
| **UIData setRowIndex cycle (100 rows, read-only)** | 294 μs | **~68 μs (−77%)** |

### Phase D — `UIRepeat`: same treatment

Mirror of Phase C for `com.sun.faces.facelets.component.UIRepeat`:

1. `setIndex` no-op short-circuit (UIRepeat's `index` is already a primitive field, so this is a plain equality check).
2. Lazy `hasStatefulDescendants` flag (UIRepeat doesn't wrap UIForm the way UIData uses UIColumn, so the scan only checks for EVH). `saveChildState(FacesContext)` short-circuits when false; `restoreChildState(FacesContext)` always recurses to keep clientId invalidation.

| Scenario | post-Phase-D |
|---|---:|
| UIRepeat (100 rows × 5 UIInput) iterate | ~187 μs |
| **UIRepeat (100 rows × 5 UIOutput, read-only) iterate** | **~62 μs (~3×)** |

### Phase E — `Util.checkIdUniqueness`: skip `UntargetableComponent` descendants; size the working HashSet sensibly

`checkIdUniqueness` runs once per `saveView` (i.e. once per postback/render). It recurses every facet+child node in the view tree, calling `getClientId()` and `HashSet.add()` on each. Facelets-compiled views liberally distribute `UILeaf` wrappers for static text and whitespace — every line of whitespace and every literal text fragment between Faces tags becomes one — so a moderately-sized view easily carries 100–300 `UILeaf` nodes whose auto-generated ids cannot collide via user-authored templates.

Two changes:
1. Skip nodes implementing `jakarta.faces.component.search.UntargetableComponent` (the spec marker interface meaning "not a meaningful target"; Mojarra's `UILeaf` implements it). Don't recurse, don't `getClientId`, don't add to the seen-ids set.
2. Initial `HashSet` sizing bumped from `viewRoot.getChildCount() << 1` (often 2 buckets in practice) to a fixed `64` — avoids log₂(N) resizes during the walk on any non-trivial view.

| Scenario (50 user components + 200 UILeaf fillers) | ns/op |
|---|---:|
| Raw walk (recurses into UILeaf) | 15,950 |
| **Patched (skips UntargetableComponent)** | **6,633 (−58%)** |

~9 μs saved per `saveView` call. At 100 req/s, ~1 ms/s of CPU reclaimed.

## What's deliberately NOT in this PR

- **`rowIndex` as primitive field on UIData**: rowIndex is part of saved view state via StateHelper; storage swap would need careful save/restore handling for marginal additional gain.
- **`int _capabilities` bit-flag refactor**: large surgery touching every getter/setter on UIComponentBase.
- **Per-component `_facesContext` cache (MyFaces pattern)**: the comment on Mojarra `UIComponentBase:1021-1034` warns "we can't use the cache ivar because we don't always know when to clear it" — needs investigation. MyFaces avoids this by scoping cache lifetime to public method entry, which is invasive.
- **Indexed `for`-loop micro-opts** (over `getFacetsAndChildren()` iterator in `process*`/`encodeChildren`/`visitTree`) and **shared `StringBuilder` in `addParentId`** and **empty-stateMap fast-path in PSS `saveView`** — individually low impact, can land later if profiling justifies.
- **HashMap-keyed facet state** in `processSaveState`/`processRestoreState`: a wire-format change, not safe for backport.
- **View pooling** (MyFaces' `ViewPoolProcessor`): entirely new infrastructure, not a backport candidate.

## Files

| Path | Change |
|---|---|
| `impl/src/main/java/jakarta/faces/component/UIComponentBase.java` | `cachedRenderer` field + invalidation; drop `.toString()` in `isRendered()`. |
| `impl/src/main/java/jakarta/faces/component/ComponentStateHelper.java` | Lazy `deltaMap` + `deltaMap()` accessor; null-safe reads. |
| `impl/src/main/java/jakarta/faces/component/UIData.java` | `setRowIndex` no-op short-circuit; lazy `hasStatefulDescendants` flag; `saveDescendantState` skip-when-no-stateful. |
| `impl/src/main/java/com/sun/faces/facelets/component/UIRepeat.java` | Same pattern as UIData applied to UIRepeat. |
| `impl/src/main/java/com/sun/faces/util/Util.java` | `checkIdUniqueness` skips `UntargetableComponent` descendants. |
| `impl/src/main/java/com/sun/faces/application/view/FaceletPartialStateManagementStrategy.java` | Initial `HashSet` capacity bumped from `viewRoot.getChildCount() << 1` to `64`. |
| `impl/src/main/java/com/sun/faces/application/view/FaceletFullStateManagementStrategy.java` | Same `HashSet` capacity bump. |
| `impl/src/test/java/jakarta/faces/component/UIComponentBaseTestCase.java` | `testCachedRendererInvariants`. |
| `impl/src/test/java/com/sun/faces/perf/ComponentTreePerfHarness.java` | Manual perf harness, disabled by default. |

## How to reproduce the numbers

```
mvn -pl impl test -Dtest=ComponentTreePerfHarness -Dperf=true
```

The harness is `@EnabledIfSystemProperty(named = "perf", matches = "true")`, so default `mvn test` skips it. Output is a tabular dump on stdout. Each commit's message contains the relevant before/after numbers for that phase.

🤖 Generated with Claude Opus 4.7

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/view/FaceletFullStateManagementStrategy.java
+++ b/impl/src/main/java/com/sun/faces/application/view/FaceletFullStateManagementStrategy.java
@@ -652,7 +652,7 @@ public class FaceletFullStateManagementStrategy extends StateManagementStrategy 
         /*
          * Check uniqueness.
          */
-        Util.checkIdUniqueness(context, viewRoot, new HashSet<>(viewRoot.getChildCount() << 1));
+        Util.checkIdUniqueness(context, viewRoot, new HashSet<>(64));
 
         /**
          * Save the dynamic actions.
--- a/impl/src/main/java/com/sun/faces/application/view/FaceletPartialStateManagementStrategy.java
+++ b/impl/src/main/java/com/sun/faces/application/view/FaceletPartialStateManagementStrategy.java
@@ -427,7 +427,7 @@ public class FaceletPartialStateManagementStrategy extends StateManagementStrate
             return null;
         }
 
-        Util.checkIdUniqueness(context, viewRoot, new HashSet<>(viewRoot.getChildCount() << 1));
+        Util.checkIdUniqueness(context, viewRoot, new HashSet<>(64));
 
         final Map<String, Object> stateMap = new HashMap<>();
         final StateContext stateContext = StateContext.getStateContext(context);
--- a/impl/src/main/java/com/sun/faces/facelets/component/UIRepeat.java
+++ b/impl/src/main/java/com/sun/faces/facelets/component/UIRepeat.java
@@ -519,13 +519,19 @@ public class UIRepeat extends UINamingContainer {
 
     private void setIndex(FacesContext ctx, int index) {
 
+        // Force the DataModel to build before any short-circuit. For a value-less
+        // ui:repeat with begin/end attributes, the first getDataModel() call has the
+        // side effect of rewriting begin/end into the model's [0, size) form (see
+        // getDataModel() ArrayDataModel branch); the caller of setIndex(faces, -1) at
+        // the start of process() relies on that side effect to fire before process()
+        // reads getBegin()/getEnd() into its local iteration bounds.
+        DataModel localModel = getDataModel();
+
         // No-op short-circuit when the index is already current. Common when callers
         // restore the iterator to -1 at end-of-iteration on a repeat that was already at -1.
         if (this.index == index) {
             return;
         }
-
-        DataModel localModel = getDataModel();
 
         // save child state
         if (this.index != -1 && localModel.isRowAvailable()) {
--- a/impl/src/main/java/com/sun/faces/facelets/component/UIRepeat.java
+++ b/impl/src/main/java/com/sun/faces/facelets/component/UIRepeat.java
@@ -341,6 +341,19 @@ public class UIRepeat extends UINamingContainer {
 
     private Map<String, SavedState> childState;
 
+    /**
+     * Per-iteration cache: {@code true} if any descendant is an {@link EditableValueHolder} whose
+     * state needs to be saved/restored across {@link #setIndex} changes; {@code false} if the
+     * entire subtree is read-only. {@code null} means not yet scanned. Computed lazily by
+     * {@link #hasStatefulDescendants()} on first need; cleared when iteration ends
+     * ({@link #setIndex} to {@code -1}) so the next iteration re-evaluates the tree shape.
+     *
+     * <p>When this is {@code false}, {@link #saveChildState(FacesContext)} and
+     * {@link #restoreChildState(FacesContext)} become no-ops, eliminating the per-row tree walk
+     * for read-only repeats.
+     */
+    private transient Boolean hasStatefulDescendants;
+
     private Map<String, SavedState> getChildState() {
         if (childState == null) {
             childState = new HashMap<>();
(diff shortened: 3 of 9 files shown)


BalusC:
TCK passed at my machine.

melloware (approved):

arjantijms (approved):

PR #5755, opened 2026-05-27
Response-writer performance improvements (range-emit + buffer elimination)

BalusC:
#5753 

## Summary

Four-phase performance refactor of the response-writer hot path, plus a follow-up API cleanup. The dominant pattern across all phases is the same — eliminate the per-character buffer-store + flush copies that the original code did "for performance" but which in JIT-era JDKs are slower than handing the underlying Writer a single bulk write. Each phase is independently reviewable.

## Measured impact

Median of 5×1M iterations, JDK 11, `NullWriter` sink. Phase A only touches `writeText`/`writeAttribute`; Phase B touches `writeURL`/`encodeURIString`; Phase C eliminates the `attributesBuffer` intermediate; Phase D consolidates element-kind classification, swaps CHM→HashMap, LinkedList→ArrayDeque.

| Scenario | baseline | Phase D | Δ |
|---|---:|---:|---:|
| `writeText` -- short plain ASCII (6 chars) | 30 ns | 21 ns | **−30%** |
| `writeText` -- short with `<`, `>`, `&` (13 chars) | 75 ns | 30 ns | **−60%** |
| `writeText` -- long plain ASCII (54 chars) | 257 ns | 79 ns | **−69%** |
| `writeText` -- long with mixed escapes | 300 ns | 73 ns | **−76%** |
| `writeAttribute` -- plain id (no escapes) | ~95 ns | 34 ns | **−64%** |
| `writeAttribute` -- style with quotes (escapes) | ~240 ns | 75 ns | **−69%** |
| `writeURIAttribute` -- href with query and fragment | 583 ns | 65 ns | **−89%** |
| `startElement` + `endElement` (empty `<div>`) | 17 ns | 12 ns | **−29%** |
| **Simulated component render** (`div` + `input` + attrs + text) | 525 ns | 215 ns | **−59%** |
| `PartialResponseWriter` -- startUpdate + text + endUpdate | 171 ns | 80 ns | **−53%** |

The simulated component render — open tag, write a few attributes, write text, close tag — is the realistic end-to-end number, **−59%**. The performance gain is bigger when JDK 17 or newer is used thanks to more JIT improvements in `OutputStreamWriter` and `CharsetEncoder`.

## Cross-JDK comparison

Same harness, **median of 3 runs**, master baseline vs final Phase D (`PD`).

| Scenario | JDK 11 base | JDK 11 PD | JDK 17 base | JDK 17 PD | JDK 21 base | JDK 21 PD |
|---|---:|---:|---:|---:|---:|---:|
| `writeText` short plain | 30 | 21 | 24 | 13 | 25 | 13 |
| `writeText` short `<>&` | 75 | 30 | 52 | 18 | 64 | 20 |
| `writeText` long plain | 257 | 79 | 213 | 49 | 273 | 47 |
| `writeText` long mixed escapes | 300 | 73 | 230 | 50 | 279 | 52 |
| `writeAttribute` plain id | ~95 | 34 | 90 | 24 | 84 | 25 |
| `writeAttribute` escapes | ~233 | 75 | 210 | 54 | 207 | 46 |
| `writeURIAttribute` | 583 | 65 | 602 | 59 | 629 | 49 |
| `startElement` + `endElement` | 17 | 12 | 14 | 11 | 16 | 17 |
| **Simulated render** | **525** | **215** | **529** | **194** | **477** | **204** |
| `PartialResponseWriter` | 171 | 80 | 173 | 70 | 148 | 77 |

Reduction on the realistic simulated render:
- **JDK 11**: 525 → 215 = **−59%**
- **JDK 17**: 529 → 194 = **−63%**
- **JDK 21**: 477 → 204 = **−57%**

Notes:
- Master baseline runs ~10–25% faster on JDK 17 than JDK 11 for `writeText`/`writeAttribute` (JIT wins in `OutputStreamWriter`/`CharsetEncoder`); JDK 21 doesn't push further on these scenarios.
- The refactor's contribution stacks on top of any JDK upgrade win. JDK 21 + Phase D simulated render at 204 ns is ~61% faster than JDK 11 master baseline.
- `writeURIAttribute` collapses from ~600 ns to ~50–65 ns on every JDK — the full Phase A+B+C+D pipeline applies to that path.

## Per-phase summary

**Phase A** (939f28742) — Rewrite `HtmlUtils.writeText` and `writeAttribute` around a range-emit strategy. Track the start of the current safe run; on hitting an escape character, bulk-write the run to the underlying Writer in one call. The fast path for HTML5 + UTF-8 (the dominant production case) is a single bounds-checked branch per char. Plain-ASCII content collapses from 50+ per-char buffer-store ops + final flush down to one underlying write per call.

**Phase B** (10b9c86c4) — Same range-emit strategy applied to `writeURL` and `encodeURIString`. The `String` entry collapses to `getChars` + char[] delegation; the legacy `<16`/`>=16` split is gone. Shared `encodeCharPercentBytes` helper extracted from the duplicated "OSW.write + flush + %XX per byte" block.

**Phase C** (1c1c24bed) — Drop the `FastStringWriter attributesBuffer` intermediate buffer in `HtmlResponseWriter`. `writeAttribute` / `writeURIAttribute` now write directly to the underlying writer. Pass-through attributes are emitted inline in `closeStartIfNecessary` (and the empty-element branch of `endElement`) right before the `>` / ` />`, preserving the legacy wire order `<el regular_attrs passthrough_attrs>`. The chunked StringBuilder→buffer→writer dump is gone.

**Phase D** (be029dcfe) — Three localized cleanups to `HtmlResponseWriter`:
- Element-name classification via a private `ElementKind` enum (`SCRIPT` / `STYLE` / `CDATA` / `OTHER`) — one length-prechecked classifier replaces 4–5 `equalsIgnoreCase` calls per element.
- `HashMap` replaces `ConcurrentHashMap` for the per-element pass-through copy (single-threaded for the render-response phase).
- `ArrayDeque` replaces `LinkedList` for the element-name stack — no per-node allocation.

**API cleanup** (9865347db) — Drop the now-unused `buff` / `outbuf` scratch-buffer parameters from `HtmlUtils.writeText`, `writeAttribute`, and `writeTextForXML`. Update all call sites accordingly. Fixes the latent regression in `FaceletViewHandlingStrategy.renderView` (it was relying on the buffer being populated as a side effect) by switching to a `StringWriter` capture.

## Wire-observable changes

Documented and locked down by regression tests. All four are independently desirable; none break HTML/XML correctness.

1. **U+20AC (€) emitted as `&#8364;` instead of `&euro;`** when escape-Unicode mode is engaged. The numeric reference is universally valid; the named entity is however invalid in XHTML/XML without a DTD declaration (rare case). The dominant HTML5 + UTF-8 case never reaches this path (€ writes as native UTF-8 bytes).
2. **Pre-existing off-by-one fix in numeric reference emission**: legacy `i > 10000` (strict `>`) emitted `&#:000;` for U+2710 (✐); now correctly `&#10000;`.
3. **`script:` injection abort now drops the entire pending prefix** for inputs that fit in the legacy 1024-char buffer. Strictly more defensive than the legacy behavior, which auto-flushed after 1024 chars and then aborted.
4. **Short URLs (<16 chars) with literal space now emit `%20` instead of `+`**. The legacy long-URL path (>=16) already produced `%20`; this unifies the two paths on the RFC 3986-correct form.

## Latent bug fix

`isAmpEscaped(char[])` gained an internal bounds check — legacy code relied on a caller-side guard that only verified the byte AFTER `&` was in bounds, so a truncated `&am` / `&amp` near end-of-input would AIOOBE. Covered by `writeURL_truncatedAmpAtEndDoesNotCrash`.

## Test changes

- Consolidated three overlapping HtmlUtils test files (`TestHtmlUtils.java`, `HtmlUtilsTest.java`, and a new Phase-A regression file) into a single `HtmlUtilsTest.java` (currently 67 tests, was 8 across the three files).
- Added four pass-through invariant tests to `HtmlResponseWriterTest.java`: regular-then-passthrough wire order, suppression of regular attributes that collide with pass-through names, empty-element pass-through emission, mixed-attribute empty-element ordering.
- New manual perf harness `ResponseWriterPerfHarness.java`, JUnit-driven, `@EnabledIfSystemProperty(named = "perf", matches = "true")`. Run with `mvn -pl impl test -Dtest=ResponseWriterPerfHarness -Dperf=true`.

Full impl test suite (1051 Java tests + 344 TS tests) green on every commit.


BalusC:
TCK passed at my machine.

BalusC:

PR #5756, opened 2026-05-28
Add test/perf benchmark module to compare Faces lifecycle phase timings across mojarra branches

BalusC:
#5753 

## Summary

New `test/perf` Maven module: a WAR + JUnit/Arquillian driver that times every Faces lifecycle phase per scenario (forms, tables, ui:repeat, composite components, plus readonly / input / nested / heavy / ajax-postback / f:viewParam-GET variants), then dumps per-(scenario, phase) `{count, total, min, max}` from a `/perf-stats` endpoint. Intended for comparing micro-perf changes across mojarra branches.

- Gated behind `-Dperf=true` (same convention as `ComponentTreePerfHarness`) so `mvn install` does not run it.
- Iteration counts tunable via `-Dperf.warmup` (default 50) and `-Dperf.runs` (default 500).
- **18 scenarios** spanning GET-only, full postback, ajax-partial postback, and GET-runs-full-lifecycle (via `f:viewParam` + `f:viewAction`).
- `PhaseTimingListener` records `System.nanoTime()` per phase per request into a thread-safe singleton `PerfStats`; `PerfStatsServlet` exposes text / JSON / reset.
- `PerfBenchIT` driver: HttpClient + CookieManager, bootstraps each postback scenario once, then loops POSTs while refreshing `jakarta.faces.ViewState`. Ajax scenarios send `Faces-Request: partial/ajax` + `jakarta.faces.partial.*` markers and extract `ViewState` from the XML partial-response.
- Writes `target/perf-stats-<timestamp>.txt`.

Converter/validator classes carry `@Dependent` in addition to `@FacesConverter(managed=true)` / `@FacesValidator(managed=true)` so they're CDI-discoverable under the CDI 4.0 default `bean-discovery-mode=annotated` (otherwise lookup silently falls back to `new()` and the bench underestimates CDI cost).

See `test/perf/README.md` for the workflow to compare two mojarra versions (build impl, swap `jakarta.faces.jar` in `target/glassfish7/glassfish/modules/`, rerun).

## Findings — combined effect of 5752 + 5754 + 5755

To validate the bench is sensitive enough to detect the in-flight perf PRs, I ran it on two mojarra builds back-to-back from the same source tree:

- **baseline** = commit 14ee982e27d8a5210ceb4dc9289ab257e82ea5d4 (merge of #5751, immediately before the perf work started)
- **all 3 PRs** = baseline + #5752 + #5754 + #5755 merged locally

[...]

PR #5752, opened 2026-05-27
Cache CDI bean resolution to eliminate per-render BeanManager lookups

BalusC:
#5753

## Summary

Mojarra invoked `BeanManager.getBeans` + `resolve` on every `createConverter` / `createValidator` / `createBehavior` call — even for built-in Faces IDs like `jakarta.faces.Integer` that aren't CDI beans. On a typical page render this multiplied into hundreds of redundant CDI registry walks per request. MyFaces avoids this via class-level managedness classification; Mojarra had no equivalent.

This PR adds a per-BeanManager `(type, beanName, qualifiers) → Bean<?>` cache in `CdiUtils` with a `NO_BEAN` sentinel for negative results. `BeanManager.getReference` + `createCreationalContext` stay per-call to preserve scope semantics. Same treatment is applied to three more hot spots: `FacesContextImpl.release` (per-request `FacesContextProducer` destroy), `getDataModelClassesMap` (per `<ui:repeat>` row), and `getCurrentInjectionPoint` (per `@ManagedProperty` injection).

The cache assumes the BeanManager is fully bootstrapped when first observed; **prerequisite Fix #5708** (InstanceFactory no longer field-caches a possibly-transient bootstrap BeanManager) is cherry-picked from 4.1/master in the same branch.

## Performance numbers

Measured on Weld SE 5.1.2 / JDK 11 (median of 5 runs × 1M iterations):

| Scenario | raw BM | cached | Speedup |
|---|---:|---:|---:|
| `FacesContextImpl.release` destroy (every request) | 578 ns | 12 ns | **48×** |
| `resolveBean(InjectionPoint.class)` (every `@ManagedProperty`) | 328 ns | 64 ns | **5.1×** |
| `resolveBeanByName("comSunFacesDataModelClassesMap")` | 81 ns | 31 ns | **2.6×** |
| `createConverter(String)` unknown ID | 2343 ns | 994 ns | **2.4×** |
| `getBeanReference` managed converter | 1411 ns | 614 ns | **2.3×** |
| `createValidator(String)` unknown ID | 4919 ns | 2259 ns | **2.2×** |
| `createConverter(Class)` Long → Number walk | 4664 ns | 2211 ns | **2.1×** |
| `createBehavior(String)` unknown ID | 983 ns | 477 ns | **2.1×** |
| `createBehavior` managed | 1342 ns | 811 ns | **1.7×** |
| `createValidator` managed | 1819 ns | 1136 ns | **1.6×** |

Reproduce with:
```
[...]

PR #5757, opened 2026-05-28
Three small request-hot-path improvements: FacetsMap empty short-circuit, descriptor-map computeIfAbsent, and UIData/UIRepeat restore-state hoist

BalusC:
#5753 

## Summary

Three small, independent improvements to the Mojarra request hot path, surfaced by JFR profiling of the test/perf bench from #5756 and validated by back-to-back phase-time measurements:

1. **Skip empty-facets allocation in `UIComponentBase$FacetsMap.keySetIterator`.** Every iteration of a component's facets/values/entries currently allocates a snapshot `ArrayList` of the keys plus an `Iterator`, even when the facets map is empty (the overwhelming majority of components). Short-circuit the empty case to `Collections.emptyIterator()`. All three iterators (Keys, Values, Entries) route through this method, so one short-circuit removes the cost from all three iteration paths.

2. **`computeIfAbsent` instead of `containsKey + put + get` in `UIComponentBase.populateDescriptorsMapIfNecessary`.** Every `UIComponentBase()` constructor calls this to locate the per-class property-descriptor map. The current implementation does `containsKey + put + get` on the ApplicationMap (two lookups + non-atomic create) for the fast path and `containsKey + put` on the per-application descriptors map for the slow path. Replace with `computeIfAbsent` / `putIfAbsent` — one lookup, atomic. Drive-by: lift the duplicated magic string to a `COMPONENT_DESCRIPTORS_MAP_KEY` constant and fix the long-standing `compnent` typo (key only used inside this method; grep confirms no external reader).

3. **Hoist the per-row state-map lookup out of `UIData.restoreDescendantState` and `UIRepeat.restoreChildState`.** For an N-row UIData/UIRepeat with M stateful descendants per row, the restore walk currently calls `getStateHelper().get(PropertyKeys.saved)` (UIData) or `getChildState()` (UIRepeat) on every single recursive node — i.e. `N * M` times per render. The lookup target is invariant across the walk (the saved-state map does not change while we are restoring), so the repeated probe is pure overhead. Fetch the map once in the outer (no-arg) restore method and thread it through the recursive helper as a parameter.

Each commit is independently reviewable and reverts cleanly.

## Findings — back-to-back measurement on the #5756 bench

**Methodology.** Same WAR / same JDK / same JVM args / same iteration count. Two runs back-to-back (~30 s apart) so thermal and FS-cache state are as close as possible. The only delta between the two is the impl jar.

- baseline = mojarra 4.0 (with PR #5754 already merged) + PRs #5752 and #5755 merged locally, **no other patches**.
- with-fixes = baseline + the three commits in this PR.
- Conditions: same WAR from `test/perf` on `add_perf_bench_it`, JDK 21.0.1, stock GlassFish 7.0.25, `-Dperf.runs=1000`, 18 scenarios.

### Wall clock and totals

| | baseline (all 3 PRs) | + 3 fixes | Δ | Δ% |
|---|---:|---:|---:|---:|
| **Failsafe wall clock** | **616.3 s** | **603.6 s** | **−12.7 s** | **−2.1 %** |
| TOTAL avg µs `RESTORE_VIEW` | 747 | 741 | −6 | −0.8 % |
| TOTAL avg µs `APPLY_REQUEST_VALUES` | 672 | 641 | −31 | **−4.6 %** |
| TOTAL avg µs `PROCESS_VALIDATIONS` | 7,016 | 6,941 | −75 | −1.1 % |
| TOTAL avg µs `UPDATE_MODEL_VALUES` | 599 | 576 | −23 | **−3.8 %** |
| TOTAL avg µs `INVOKE_APPLICATION` | 34 | 33 | −1 | −2.9 % |
| TOTAL avg µs **`RENDER_RESPONSE`** | **23,892** | **23,407** | **−485** | **−2.0 %** |

### Why this looks like real signal

1. Every single TOTAL phase moved in the right direction — no regression on any phase. Direction so consistent across six independent buckets that this is not run-to-run noise.
2. The biggest wins land **exactly** where the hoist fires: `APPLY_REQUEST_VALUES` (−4.6 %) and `UPDATE_MODEL_VALUES` (−3.8 %) are the two phases where `UIData.setRowIndex` → `restoreDescendantState` runs on every row visit. `RENDER_RESPONSE` (−2.0 %) picks up the same hoist (it fires there too) plus the FacetsMap empty short-circuit and the per-construction `computeIfAbsent` win.
3. JFR profile (`warmup=20 runs=200`, separately) shows `ComponentStateHelper.get` dropping from 86 → 74 absolute samples on the http-listener threads (**−14 %**), confirming the hoist's specific signature is real and on its target method.
4. Wall clock −2.1 % is small in absolute terms but stacks on top of #5752/#5754/#5755's −7.6 % (per #5756). Cumulative wins.

### Caveats

- Single back-to-back run pair; multiple alternating runs would tighten error bars further. The direction is so uniformly negative on phase TOTALs that the call here is safe.
- The FacetsMap empty short-circuit doesn't fire heavily on this workload (the bench uses `<h:column>` with `<f:facet name="header">` in every table scenario, so most facets are non-empty). The benefit is more visible on facets-light real-world views.

## Methodology

Reproduces against the perf-bench module on #5756:

```bash
# Branch with this PR + the in-flight perf PRs
git checkout add_perf_bench_it
git checkout -b combine_perf_top_fixes
git merge --no-edit cdi_performance_improvements          # PR #5752
git merge --no-edit response_writer_performance_improvements  # PR #5755
git merge --no-edit fix_top_performance_hot_spots         # this PR

. java21 && mvn -pl impl install -DskipTests
mvn -pl test/perf -am clean package -DskipTests -Dglassfish.version=7.0.25
mvn -pl test/perf failsafe:integration-test failsafe:verify \
    -Dperf=true -Dperf.runs=1000
```

For the baseline run, repeat without the final `git merge` of this PR.

🤖 Generated with Claude Opus 4.7

### DIFF
--- a/impl/src/main/java/com/sun/faces/facelets/component/UIRepeat.java
+++ b/impl/src/main/java/com/sun/faces/facelets/component/UIRepeat.java
@@ -468,19 +468,22 @@ public class UIRepeat extends UINamingContainer {
 
     private void restoreChildState(FacesContext ctx) {
         // Unlike saveChildState, the restore-side walk must always run even when no
-        // descendants carry per-row state: restoreChildState(faces, c) below calls
-        // setId(getId()) on every component to invalidate its cached clientId so each row
-        // gets a row-specific clientId (e.g. repeat:N:foo). Skipping the walk would leave
+        // descendants carry per-row state: restoreChildState(faces, c, childState) below
+        // calls setId(getId()) on every component to invalidate its cached clientId so each
+        // row gets a row-specific clientId (e.g. repeat:N:foo). Skipping the walk would leave
         // stale clientIds and render duplicate id="..." attributes across rows.
         if (getChildCount() > 0) {
 
+            // The child-state map is invariant across this walk; fetch it once and thread it
+            // through the recursion instead of re-querying getChildState() at every node.
+            Map<String, SavedState> childStateMap = getChildState();
             for (UIComponent uiComponent : getChildren()) {
-                this.restoreChildState(ctx, uiComponent);
+                this.restoreChildState(ctx, uiComponent, childStateMap);
             }
         }
     }
 
-    private void restoreChildState(FacesContext faces, UIComponent c) {
+    private void restoreChildState(FacesContext faces, UIComponent c, Map<String, SavedState> childStateMap) {
         // reset id
         String id = c.getId();
         c.setId(id);
@@ -489,7 +492,7 @@ public class UIRepeat extends UINamingContainer {
         if (c instanceof EditableValueHolder) {
             EditableValueHolder evh = (EditableValueHolder) c;
             String clientId = c.getClientId(faces);
-            SavedState ss = getChildState().get(clientId);
+            SavedState ss = childStateMap.get(clientId);
             if (ss != null) {
                 ss.apply(evh);
             } else {
@@ -500,7 +503,7 @@ public class UIRepeat extends UINamingContainer {
         // continue hack
         Iterator itr = c.getFacetsAndChildren();
         while (itr.hasNext()) {
-            restoreChildState(faces, (UIComponent) itr.next());
+            restoreChildState(faces, (UIComponent) itr.next(), childStateMap);
         }
     }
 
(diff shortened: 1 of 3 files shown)


BalusC:

PR #5759, opened 2026-06-01
Reduce request-pipeline overhead (state, traversal, render output, events)

BalusC:
Internal performance optimizations for the Faces request pipeline, surfaced by JFR-profiling the impl across six servers (Mojarra vs MyFaces) under `test/perf`. All changes are implementation-internal — **no API, spec, or saved-state-format changes** — and 4.1-eligible.

### Optimizations (`impl`)
- **Per-component property/state access** — `UIComponentBase$AttributesMap` resolves the backing state map lazily and caches the property getter; getters are `setAccessible(true)` once per component class to skip the per-invoke reflective access check; `ComponentStateHelper.eval` drops a varargs allocation.
- **NamingContainer-ancestor cache** — cache the ancestor lookup behind `getClientId`, invalidated on `setParent` (not `setId`), so it survives UIData/UIRepeat per-row id refresh.
- **UIData / UIRepeat per-row state** — collect the descendant set once per iteration and walk flat lists instead of re-traversing the subtree every row for per-row save/restore + clientId reset. ≈6% server-side on the table/repeat/input scenarios (Tomcat).
- **Tree traversal** — indexed child iteration in `process*`/`encodeAll` (no per-call iterator allocation).
- **Event publishing** — skip the application-level listener lookup when no global listener is subscribed for an event class; resolve view listeners before the reentrancy guard.
- **Render output coalescing** — wrap the response writer once in `ExternalContextImpl.getResponseOutputWriter()` (cached `BufferedWriter`) so render fragments coalesce before the container writer. WildFly `RENDER_RESPONSE` −6.2% / total −5.0%; neutral on Grizzly/Coyote. Done at `getResponseOutputWriter()` (not in `renderView`) so render-view listeners share the same ordered buffer, flushed at request teardown — preserving Pre/PostRenderViewEvent output order.

### Behavior change to note
- **The duplicate-id check is skipped in `Production` stage by default**, matching MyFaces' `CHECK_ID_PRODUCTION_MODE=auto`. `com.sun.faces.disableIdUniquenessCheck` is now tri-state `true|false|auto` (default `auto`); set it to `false` to restore the always-on check.

### Test infrastructure (`test/perf`)
Multi-server microbenchmark support used to drive/validate the above: OpenLiberty bumped to 26.0.0.5-beta (EE 11 `faces-4.1`) and a `perf.heapArgument`→`perf.jvmArguments` rename.

### Validation
- 1099 impl unit tests pass.
- Jakarta Faces TCK green.

Follow-up to the cross-impl benchmarking in #5753 (see also PR #5756).

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/applicationimpl/Events.java
+++ b/impl/src/main/java/com/sun/faces/application/applicationimpl/Events.java
@@ -27,6 +27,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Logger;
 
 import com.sun.faces.application.applicationimpl.events.ComponentSystemEventHelper;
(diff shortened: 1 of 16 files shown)


BalusC:

PR #5760, opened 2026-06-01
Reduce request-pipeline overhead (state, traversal, render output, events) — 4.0 backport

BalusC:
Backport of #5759 to 4.0 — the request-pipeline performance optimizations, **impl only** (the `test/perf` multi-server infrastructure is intentionally not backported). Java 11-clean (4.0 targets 11).

Cherry-picked impl commits:
- Resolve attributes map lazily in `AttributesMap.get` + `setAccessible` on cached property getters
- Skip the id-uniqueness check in `Production` by default
- Use indexed child traversal in `process*`/`encodeAll`
- Reduce `publishEvent` overhead in the no-listener case
- Cache the NamingContainer-ancestor lookup
- Precollect UIData/UIRepeat descendants for per-row state save/restore
- Coalesce render output before the container writer (at `getResponseOutputWriter()`)

See #5759 for rationale, six-server benchmarks, and TCK validation.

### Behavior change to note
The duplicate-id check is skipped in `Production` stage by default (`com.sun.faces.disableIdUniquenessCheck` tri-state `true|false|auto`, default `auto`); set it to `false` to restore the always-on check.

### Validation
- Builds on JDK 11.
- impl unit tests pass.

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/applicationimpl/Events.java
+++ b/impl/src/main/java/com/sun/faces/application/applicationimpl/Events.java
@@ -27,6 +27,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Logger;
 
 import com.sun.faces.application.applicationimpl.events.ComponentSystemEventHelper;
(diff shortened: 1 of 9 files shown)


BalusC:
For the backport chain: the `auto` default for `com.sun.faces.disableIdUniquenessCheck` introduced here is changed to `false` (always check; skip stays opt-in via `auto`/`true`) in #5761.

Issue #4128, opened 2016-04-11
To track Bug 22845377 - JSF DUPLICATING AJAX EVENTS


ren-zhijun-oracle:
This issue is to track below WLS bug.

Bug 22845377 - JSF DUPLICATING AJAX EVENTS AFTER APPLYING WEBLOGIC SERVER 12.1.2.0.8 PSU PATCH
#### Affected Versions
[2.1.20-07]

Issue #3311, opened 2014-06-03
Remove findChildByTagId() recursion


ren-zhijun-oracle:
Back in 2.1.7, the following recursive call was introduced into ComponentSupport.findChildByTag():

```
/*
             * Make sure we look for the child recursively it might have moved
             * into a different parent in the parent hierarchy. Note currently
             * we are only looking down the tree. Maybe it would be better
             * to use the VisitTree API instead.
             */
            UIComponent foundChild = findChildByTagId(context, c, id);
            if (foundChild != null) {
return foundChild;
            }
```

This change was made as part of a larger effort to address various dynamic component issues reported in:

#1830 SystemEvent Acid Test

As mentioned in the comment in the above code, the intention is to locate a child component that has been re-parented. However, only children that have been reparented into a some other subtree within the same parent will be found.

Unfortunately, this recursive call to findChildByTagId() results in very poor tag execution performance.

There are two tag execution cases to consider:

1\. Execution to create a new component tree.
2\. Re-execution against an existing component tree (postbacks)

In case #1, each time we execute a tag to create a component, we must first traverse every subtree that is already rooted at the parent.

Prior to this change, rather than recursively traversing every subtree, findChildByTagId() simply iterated over the target parent's immediate children.

While subtree traversal overhead may not be significant for small subtrees, it is significant for large subtrees.

My guess is that this change is the underlying source of this old performance issue:

#2498 JSF page performance degrades significantly as page size increases.

Which complains of increased tree walking that was introduced between 2.1.5 and 2.1.11:

> We have gone from 1.7 seconds for the attached example to 3.5 seconds by upgrading from Mojarra 2.1.5 to Mojarra 2.1.11\. This is more than just an AJAX issue. Rendering of pages is slow and gets slower as the page size increases. By debugging Mojarra code I have found that for every "entry" in a page processed the tree is walked from the top node down. It finds the correct positions of the nodes to be modified by walking the tree 100s of times per page load. Is there a way Hash maps or other data structures could be used to avoid walking the full tree so many times per page load?

Fortunately, Ed has recently implemented a more general findChildByTagId() optimization. See:

#3197 Regression -JSF page performance degrades significantly as page size increases. Resurfaces in 2.2.5 release.

This optimizes findChildByTagId() away in non-postback cases, which addresses case #1.

Which gets us to case #2.

Case #2 is not as heavily impacted by the findChildByTagId() recursion as case #1.

[...]

PR #5761, opened 2026-06-02
Fix slow RESTORE_VIEW: replace the descendant-mark-id cache with a refresh-gated direct scan

### LINKED ISSUE #4811: RESTORE_PHASE is slow on Mojarra 2.3

BalusC:
## Problem

On a postback, RESTORE_VIEW re-applies the Facelet ("refresh"), and for each tag `ComponentSupport.findChildByTagId` locates the existing component by its `MARK_CREATED` tag id so it can be reused rather than recreated. On large views this lookup dominates RESTORE_VIEW — the regression reported in #4811 (RESTORE_VIEW up to ~26× slower on 2.3 than on 2.2).

### How it got slow

#4128 (ajax event duplication with component binding) and #4146 (a related WLS-PSU page-rendering bug) were fixed by removing the guards that made `findChildByTagId` return `null` while a view was being (re)built. That is correct — the refresh must search during the build to reuse a `binding`-supplied component — but it turned `findChildByTagId` into an O(n²) scan on large views: a wide sibling list is rescanned in full for every sibling.

#4808 then masked the O(n²) with a per-component **descendant-mark-id cache** — an O(1) lookup, but maintained by propagating every mark id up to **every ancestor** on each child add/remove (O(n·depth)), plus a `HashMap` allocated per component. #3311 ("Remove findChildByTagId() recursion", 2014) had already scoped the proper fix and #3310 (2015) unblocked it, but it was never carried out.

## Fix

Remove the cache entirely and replace both it and the whole-subtree recursion with a **bounded direct scan**, gated the way MyFaces gates its equivalent (`isRefreshingSection`):

- While applying the children of a **freshly created** component — a subtree with nothing existing to reuse — the scan is skipped (`BUILDING_FRESH_SUBTREE`). This is what restores O(n) on wide views.
- For **found or `binding`-supplied** parents — which arrive with their children already attached — the scan stays active, so component-binding reuse (#4128 / #4146) remains correct.

This is a **structural** fresh-vs-existing distinction (`!componentFound && childCount == 0`) rather than the phase-based guard that broke those cases. `cc:insertChildren` children relocated below the composite implementation facet are resolved by `findReparentedComponent` via a deep search that bypasses the gate. The now-dead cache-maintenance call sites in `UIComponentBase` and `ComponentStateHelper` are removed.

Net effect: O(n) refresh on wide views, faster than the cache it replaces, with no per-component `HashMap` and no add/remove maintenance.

### Also: EL component-stack resolution

`pushComponentToEL` / `popComponentFromEL` / `getCurrentComponent` re-resolve the component EL stack ~6000×/request via `HashMap.computeIfAbsent`. The stack is created once per request and present on every subsequent call, so the `computeIfAbsent` lambda is pure overhead; a plain `get` + lazy `put` makes it a single `HashMap.get`. Worth ~6% on both RESTORE_VIEW and RENDER_RESPONSE in the composite-heavy benchmark.

### Also: id-uniqueness check default

#5759 (commit 0c5ac17) defaulted `com.sun.faces.disableIdUniquenessCheck` to `auto`, which fully skips the duplicate-id tree walk in Production. That silently turns the check off in Production for existing applications, and it would skew the cross-server comparison below by letting Mojarra skip work the other implementations still perform. This PR defaults it to `false` (the check always runs); the skip stays opt-in via `auto`/`true`. The cost of keeping it on is minor — the walk already excludes the Facelets-generated static-text leaf wrappers.

## Validation

Full Jakarta Faces TCK green. The original #4128 regression test (component binding into a `dataTable`, asserting the action fires exactly once per click) was lost in the 3.0→4.0 migration; it has been resurrected as a modern Selenium IT (jakartaee/faces#2179) and is the gating test — it fails on the naive phase-guard variant and passes with this structural gate.

## Benchmark

Adds a `composite-heavy` scenario to `test/perf` (≈3000 components, the #4811 reproducer shape). Cross-server numbers to follow.

Closes #4811. Part of the #5753 performance review.

### DIFF
--- a/impl/src/main/java/com/sun/faces/config/WebConfiguration.java
+++ b/impl/src/main/java/com/sun/faces/config/WebConfiguration.java
@@ -769,7 +769,7 @@ public class WebConfiguration {
         ResourceUpdateCheckPeriod("com.sun.faces.resourceUpdateCheckPeriod", "5"), // in minutes
         CompressableMimeTypes("com.sun.faces.compressableMimeTypes", ""),
         DisableUnicodeEscaping("com.sun.faces.disableUnicodeEscaping", "auto"),
-        DisableIdUniquenessCheck("com.sun.faces.disableIdUniquenessCheck", "auto"), // true|false|auto; auto skips the check in Production
+        DisableIdUniquenessCheck("com.sun.faces.disableIdUniquenessCheck", "false"), // true|false|auto; default false (always check), opt-in auto skips it in Production
         FaceletsDefaultRefreshPeriod(ViewHandler.FACELETS_REFRESH_PERIOD_PARAM_NAME, "0"), // this is default for non-prod; default for prod is set in WebConfiguration
         FaceletsViewMappings(ViewHandler.FACELETS_VIEW_MAPPINGS_PARAM_NAME, ""),
         FaceletsLibraries(ViewHandler.FACELETS_LIBRARIES_PARAM_NAME, ""),
--- a/impl/src/main/java/com/sun/faces/facelets/tag/faces/ComponentSupport.java
+++ b/impl/src/main/java/com/sun/faces/facelets/tag/faces/ComponentSupport.java
@@ -57,7 +57,15 @@ public final class ComponentSupport {
 
     private final static String MARK_DELETED = "com.sun.faces.facelets.MARK_DELETED";
     public final static String MARK_CREATED = "com.sun.faces.facelets.MARK_ID";
-    private final static String MARK_ID_CACHE = "com.sun.faces.facelets.MARK_ID_CACHE";
+
+    /**
+     * FacesContext-scoped flag set by {@code ComponentTagHandlerDelegateImpl} while applying the children of a
+     * freshly created component. When set, {@link #findChildByTagId} skips its scan: a freshly built subtree has
+     * no pre-existing children to reuse, so searching it is pure waste (the source of the O(n^2) refresh on wide
+     * views). The scan stays active for existing/reused parents -- binding-supplied or refreshed subtrees -- which
+     * is what keeps component-binding reuse (#4128/#4146) correct.
+     */
+    public final static String BUILDING_FRESH_SUBTREE = "com.sun.faces.facelets.BUILDING_FRESH_SUBTREE";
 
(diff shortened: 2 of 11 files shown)


BalusC:

PR #5764, opened 2026-06-03
Reduce RESTORE_VIEW overhead: dynamic-add gate, field-backed Facelets markers, ancestor memoization

BalusC:
Follow-up to #5759 / #5761 on the `RESTORE_VIEW` performance work tracked in #5753. Five targeted, broadly-applicable reductions in per-component restore overhead (restore covers the postback `buildView` + partial-state restore), validated by JFR (composite-heavy, the #4811-scale scenario) and the [perf bench](https://github.com/eclipse-ee4j/mojarra/blob/4.1/test/perf/README.md).

### Commits

1. **Skip the per-component dynamic-add lookup during partial-state restore.** `StateContext.componentAddedDynamically` did `getAttributes().containsKey(DYNAMIC_COMPONENT)` for every component in the restore `visitTree`. Seeded from the saved dynamic-action list, a view with no dynamic actions skips it entirely. Default stays "always check" on every other path.

2. **Field-back the hot Facelets markers for authoritative attribute reads.** `MARK_CREATED`, `MARK_DELETED`, facet name, `MARK_CHILDREN_MODIFIED`, `REMOVED_CHILDREN`, `DYNAMIC_COMPONENT` were read via `getAttributes().get/containsKey` on every refresh — a state-map `HashMap` lookup hashing a long key string, and the common *absent* case ("is this deleted / a facet / dynamically added?") still paid for it. They're now cached on private `UIComponentBase` fields that `AttributesMap.get/containsKey` read **authoritatively** (no map fallthrough), so absent-marker checks become field reads. Writes still flow through the attributes map, so **saved state is unchanged**: partial-state restore is self-correcting (`buildView` re-puts), full-state restore repopulates the fields from the restored map. This is the same shape MyFaces uses (`_ComponentAttributesMap` → component fields).

3. **Memoize `getNamingContainerAncestor` through the parent's cached ancestor.** The cache (added in #5759) was recomputed with a full O(depth) parent-chain walk whenever cold — and `buildView` clears it per component via `setParent` — so it was O(n·depth) across the tree. Resolving via the parent's already-memoized ancestor makes it O(n), same value, same invalidation point.

4. **Override `getNamingContainer` to reuse that same memoized cache** (companion to 3). `UIComponent.getNamingContainer()` did its own uncached parent-chain walk, called per component during build (`assignUniqueId`, the text handlers). Override in `UIComponentBase` to reuse the cached ancestor: `this` if it is a `NamingContainer`, else the memoized ancestor.

5. **Defer `getViewRoot` in `assignUniqueId` to the non-`UniqueIdVendor` branch.** The view root was resolved unconditionally per component (a parent-chain walk), but is only used when the component's `NamingContainer` is *not* a `UniqueIdVendor` (rare — composites/forms/the view root all are). Resolve it lazily only on that branch.

### Measurements (JFR, composite-heavy restore phase)

| metric | before | after | note |
|---|--:|--:|---|
| `AttributesMap.get/containsKey` | 25.2% | **6.1%** | now below MyFaces' ~8% |
| `HashMap.getNode` (restore leaf) | 35.8% | **10.6%** | |
| `getNamingContainerAncestor` | 8.0% | **1.5%** | |
| RESTORE_VIEW wall (GlassFish) | ~16.7 ms | **~15.1 ms** | ~10%, same-session, non-overlapping ranges |

Cross-impl, Tomcat, full bench (Mojarra 4.1.10-SNAPSHOT *with this branch* vs MyFaces 4.1.4-SNAPSHOT): overall wall within ~5%; `APPLY_REQUEST_VALUES` −17%; `RENDER_RESPONSE` +4%; `RESTORE_VIEW` still +55%, concentrated in `composite-heavy` (+96%).

### Validation

- 1078 impl unit tests pass — including `SearchExpressionHandlerTest`'s exact multi-level nested clientId assertions (`form:outerContainer:innerContainer:source`), which directly exercise the `getNamingContainerAncestor` change.
- All 19 bench scenarios pass (real GET→postback cycles; a wrong marker/clientId would break field matching on postback).

### Where this leaves it

These (plus #5759 / #5761) take the concentrated per-component overhead off `RESTORE_VIEW` and flatten the profile: `AttributesMap` is no longer a factor, the ancestor walk is gone. After them, composite-heavy restore is **~62% `buildView`** — re-executing the Facelets template to rebuild ~3000 components on every postback (tag-handler application, unique-id/clientId construction, composite-stack management). The remaining per-component hotspots (clientId `StringBuilder` allocation, `privateOnComponentPopulated`, tag-attribute application) are each only a few percent.

The cross-impl gap on composite-heavy (~2×) is **`buildView` efficiency**, not view pooling: MyFaces view pooling is **off by default** (its `ViewPoolProcessor` is only created in Production with `org.apache.myfaces.CACHE_EL_EXPRESSIONS=alwaysRecompile` and `FACELETS_REFRESH_PERIOD=-1`; the default is `noCache`, so it is never instantiated), and the bench runs MyFaces on defaults. So MyFaces rebuilds the same tree ~2× faster than Mojarra with pooling off on both sides.

A direct Mojarra-vs-MyFaces `buildView` comparison (composite-heavy) places the residual in **per-component `HashMap` operations**: Mojarra's unique-id generation (`DefaultFaceletContext.generateUniqueId` `ids.get`/`put` + per-component `IdMapper.getMapper`/`createUniqueId`) and the per-component tag application, where MyFaces uses **section counters and instance fields** (`startComponentUniqueIdSection`). Commits 4–5 take the clean parts (the redundant ancestor / view-root walks); the rest is structural — closing it to par would mean a section/counter-based unique-id generation, which carries a sharp correctness constraint (generated ids must stay deterministic and identical between save and restore) and so is a separate, TCK-gated change, not incremental hotspot-picking.

This PR banks the broadly-applicable per-component restore wins. The remaining composite-heavy gap is structural in the Facelets build layer (section-based id generation, or — orthogonally — view pooling to skip the rebuild entirely), and is tracked as follow-up to #5753.

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/view/FaceletPartialStateManagementStrategy.java
+++ b/impl/src/main/java/com/sun/faces/application/view/FaceletPartialStateManagementStrategy.java
@@ -341,6 +341,11 @@ public class FaceletPartialStateManagementStrategy extends StateManagementStrate
         if (state != null) {
             try {
                 stateContext.setTrackViewModifications(false);
+                // No saved dynamic actions => no component carries DYNAMIC_COMPONENT, so the per-component
+                // marker lookup in componentAddedDynamically can be skipped across the whole restore traversal.
+                @SuppressWarnings("unchecked")
+                List<Object> savedActions = (List<Object>) state.get(DYNAMIC_ACTIONS);
+                stateContext.setHasDynamicComponents(!isEmpty(savedActions));
 
                 VisitContext visitContext = VisitContext.createVisitContext(context, null, SKIP_ITERATION_AND_EXECUTE_LIFECYCLE_HINTS);
                 viewRoot.visitTree(visitContext, (context1, target) -> {
@@ -366,6 +371,7 @@ public class FaceletPartialStateManagementStrategy extends StateManagementStrate
                 });
                 restoreDynamicActions(context, stateContext, state);
             } finally {
+                stateContext.setHasDynamicComponents(true);
                 stateContext.setTrackViewModifications(true);
             }
         } else {
--- a/impl/src/main/java/com/sun/faces/context/StateContext.java
+++ b/impl/src/main/java/com/sun/faces/context/StateContext.java
@@ -62,6 +62,10 @@ public class StateContext {
     private boolean partial;
     private boolean partialLocked;
     private boolean trackMods = true;
+    // A view whose saved state carries no dynamic add/remove actions has no component bearing the
+    // DYNAMIC_COMPONENT marker, so componentAddedDynamically can skip the per-component attribute
+    // lookup. The partial-state restore sets this false for such views; true (always check) otherwise.
+    private boolean hasDynamicComponents = true;
     private AddRemoveListener modListener;
     private ApplicationStateInfo stateInfo;
     private WeakReference<UIViewRoot> viewRootRef = new WeakReference<>(null);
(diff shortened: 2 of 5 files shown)


BalusC:
@tandraschko this `composite-heavy` benchmark is clearly where MyFaces view pooling approach is a big win. There's basically nothing left to further optimize in Mojarra. Very nice! :)

tandraschko:
great work! together with claude we can easy work on that stuff :)

i also wonder why MyFaces is so much faster in that case, Leo did a great job in the past.
Just to make it clear, not sure if i read it correctly: ViewPooling is deactivated per default - or did you try it with activated ViewPooling?

BalusC:
View pooling is by default off? The wall hit the `buildView` stage so I understood it's due to pooling. Well, it apparently boils further down to a much leaner `buildView` implementation in MyFaces. That'll be the next optimization target :)

PR description has been updated.

tandraschko:
yep
View Pooling and ValueExpression Cache is off by default

PR #5780, opened 2026-06-08
LruCache optimization to reduce lock contention

pizzi80:
I've refactored LRUCache to switch back to a native 
synchronized block and move the factory execution outside of it.

With JEP 491, synchronized is now 100% friendly with Virtual Threads (no pinning), 
and the previous ReentrantLock approach with computeIfAbsent 
created a sequential bottleneck inside the lock that heavily penalized performance under high concurrency.

### DIFF
--- a/impl/src/main/java/com/sun/faces/util/LRUCache.java
+++ b/impl/src/main/java/com/sun/faces/util/LRUCache.java
@@ -15,65 +15,69 @@
  */
 package com.sun.faces.util;
 
-import java.util.Objects;
-import java.util.concurrent.locks.ReentrantLock;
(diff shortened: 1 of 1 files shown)


BalusC:
I had Claude review this and this change is not acceptable because `Matcher` in `ELUtils` (the sole caller of this `LRUCache`) is not thread safe. It would work if `Matcher` is replaced with `Pattern` but then we're circling back to the very issue #4082 the original impl tried to fix.

But Claude did find other things based on source code revision history and has recommended a better solution. I need to review it myself and then I'll post it.

BalusC:
Superseded by https://github.com/eclipse-ee4j/mojarra/pull/5782

PR #5782, opened 2026-06-09
Replace composite-expression cache with guard + per-attribute memoization

BalusC:
Supersedes #5780. Targets `4.1`.

## Background

`ELUtils.isCompositeComponentExpr(String)` classifies whether an EL expression refers to a composite component (`#{cc.attrs.x}`). It is called per tag-attribute during view build from `TagAttributeImpl.getValueExpression`/`getMethodExpression` and from `ELText`.

Since 2020 (#4810, for issue #4082) the result has been memoized. The current `4.1` form caches it in a `static LRUCache` — a fair `ReentrantLock` around an access-ordered `LinkedHashMap`, with the cache factory running a single shared mutable `Matcher`.

That design has three problems:

1. **It serializes under load.** Because an access-ordered `LinkedHashMap` mutates on every `get`, the cache must take an exclusive lock for *reads*. With the fair `ReentrantLock`, every concurrent lookup queues. On a 20-core box throughput collapses from ~40M ops/sec single-threaded to **~0.1M ops/sec** once contended — i.e. under load the cache added to save CPU *is* the bottleneck.
2. **Shared mutable `Matcher`.** Safe only because the factory ran inside the lock. #5780 moves the factory out of the lock to cut contention, which re-introduces a data race on that `Matcher` (a quick stress harness produced hundreds of thousands of wrong answers plus regex exceptions).
3. **Static cache in a container deployment.** As a `static` field it is shared across all web-apps when Mojarra is hosted by the container rather than the WAR — flagged by Thomas Andraschko on the original PR in 2020.

MyFaces never cached this. It does two cheaper things, and both were recommended on #4810 / #4082 in 2020 but not taken:

- a `String.contains("cc")` guard before the regex, and
- storing the boolean **once on `TagAttributeImpl`** (alongside `isLiteral()`), since these checks are only called from there.

## Change

Drop the cache; do what MyFaces does.

- **`ELUtils.isCompositeComponentExpr()`** — probe with a cheap `String.contains("cc")` before the regex. The pattern requires a literal `cc.`, so `contains("cc")` is a sound *necessary-condition* prefilter (proven and empirically verified to never change the result). It skips the backtracking regex for the common non-composite case, and is now null-safe.
- **`TagAttributeImpl`** — classify the attribute's immutable `value` once in the constructor into a `final boolean compositeComponentExpr` field, mirroring the existing `final boolean literal`. The hot view-build paths now read a field instead of re-running the regex. The public 3-arg `getValueExpression(ctx, expr, type)` reuses the field only when `expr == value` (the internal path) and falls back to `ELUtils` for a foreign `expr`.
- **Remove `LRUCache`** — `ELUtils` was its only user; no test referenced it. `LRUMap` (used elsewhere) is untouched.

No behavioral change: the guard is a necessary-condition prefilter, and removing the cache only removes memoization of a pure function. The only observable difference is that `isCompositeComponentExpr(null)` now returns `false` instead of throwing NPE.

## Benchmark

Standalone microbench, 20-core / JDK 21, realistic expression mix (incl. the long non-`cc` expressions that motivated #4082), 1000-entry hit-warm working set. Million ops/sec, higher is better:

| variant | 1 thr | 4 | 16 | 64 |
|---|--:|--:|--:|--:|
| current `4.1` — LRU + fair lock | 40 | 0.1 | 0.1 | 0.1 |
| #5780 — `synchronized`, factory outside lock | 45 | 7.1 | 2.9 | 3.1 |
| **this PR — `ELUtils` guard + `find()`** (ELText / foreign expr) | 35 | 127 | **370** | 376 |
| **this PR — `TagAttributeImpl` field read** (view-build hot path) | 100 | 378 | **1077** | 1133 |
| reference: lock-free `ConcurrentHashMap` | 91 | 341 | 982 | 1110 |
| reference: MyFaces guard + anchored `matches()` | 30 | 106 | 304 | 304 |
| reference: pre-2020 plain `find()` (no cache) | 0.7 | 2.7 | 7.0 | 6.9 |

At 16 threads the per-attribute field read is ~10,000× the current locked cache and ~370× #5780, with no lock, no allocation per call, and no cross-app static state. The plain-`find()` row confirms the 2020 motivation was real (slow on pathological long expressions) — but the `contains("cc")` guard, not a cache, is what removes that cost.

## Notes

- Builds clean on `impl` (JDK 17 release).
- Closes the concurrency concern #5780 set out to address, without the lock or the shared-`Matcher` race.


pizzi80:
perfect !

kguelzau:
Nice to see that TagAttributeImpl finally has that field now instead of using the LRUCache like I [tried to introduce some time ago](https://github.com/kguelzau/eclipse-ee4j-mojarra/commit/d253c44acb8f297587fb211d3b8d58e3d55b2d5c) 👌😊

Issue #2028, opened 2011-04-05
RI does not comply with the API doc(the attribute noSelectionValue)


ren-zhijun-oracle:
The problem was reported against RI mojarra-2.1.0.

There is an attribute name "noSelectionValue" in tag f:selectItems.
But RI does use "noSelectionOption". Which one is correct?
#### Affected Versions
[2.1.0]

PR #5801, opened 2026-06-17
Fix CCE in isRendered() on String-valued rendered expression

BalusC:
## Problem

`UIComponentBase.isRendered()` reads the `rendered` attribute via a strict `(Boolean)` cast on the `StateHelper.eval()` result:

```java
return (Boolean) getStateHelper().eval(PropertyKeys.rendered, TRUE);
```

When the `rendered` value expression is bound with a **non-`Boolean` expected type** — e.g. through a templating layer that creates the `ValueExpression` with `Object.class` rather than the component's `boolean` property type — `eval()` returns the raw evaluated value, which may be a `String`. The cast then throws:

```
java.lang.ClassCastException: class java.lang.String cannot be cast to class java.lang.Boolean
    at jakarta.faces.component.UIComponentBase.isRendered(UIComponentBase.java:312)
```

## Regression source (4.x)

On 4.x this was introduced by PR #5754 (commit aafc13cc76, "Phase A of component-tree perf work"). Before that, `isRendered()` was `Boolean.valueOf(eval(...).toString())` — it called `toString()` on whatever `eval()` returned and parsed that back to a boolean, so a `String` coerced cleanly. #5754 replaced the String round-trip with the direct `(Boolean)` cast to drop the allocation, which silently removed that tolerance.

## Where it surfaced

GlassFish's Admin UI, where `rendered="#{sessionScope.restartRequired}"` resolves to a `String` stored in the session and is bound via jsftemplating with an `Object` expected type — bypassing the EL `String`→`Boolean` coercion that a standard Facelets attribute (created with the `boolean` property type) would apply.

## Actually observed on 5.0.0-M2 — different root cause

It was reported against Mojarra 5.0.0-M2, which **does not contain the #5754 perf work** — its `isRendered()` is still the lenient `Boolean.parseBoolean(eval(rendered, TRUE).toString())`. It nonetheless throws the same CCE, for a different reason: `StateHelper.eval` was generified to `<T> T eval(Serializable key, T defaultValue)` (#2028). Passing the `TRUE` default makes type inference resolve `T = Boolean`, so the **compiler inserts a synthetic `checkcast Boolean`** on the return value *before* `.toString()` even runs. A `String` trips that checkcast — same exception, same line, but produced by the compiler rather than written by hand.

So 5.0 has two independent routes to the identical CCE: (1) the generified-`eval` inference (#2028, present since M2), and (2) the strict `(Boolean)` cast later merged up from 4.x via #5754. Either alone is sufficient.

Implication for the 5.0 fix: dropping the `TRUE` default (single-arg `eval`, which returns `Object`) is **required** to defeat route (1) — assigning the result to an `Object` is not enough, because inference still picks `T = Boolean` from the `TRUE` argument and re-inserts the checkcast. This 4.x fix already drops the default, so it ports cleanly; 5.0 additionally gets the dev-stage warning.

## Note on correctness

Strictly, this is **user error and not spec compliant**: `rendered` is a boolean property, and the application should store a `Boolean` (or bind the VE with a `Boolean` expected type) so EL coerces it. Nothing in the spec requires `isRendered()` to accept a `String`. But the lenient behavior stood for years and apps came to rely on it, so restoring it is the backward-compatible fix and costs nothing on the common path.

## Scope: other boolean properties reviewed

`rendered` is the **only** regression. Every other boolean getter (`disabled`, `readonly`, `required`, `immediate`, `escape`, `ismap`, `multiple`, `redisplay`, and the rest of the generated `Html*` getters) has always used a strict `(Boolean)` cast — long before #5754 — so their behavior is unchanged:

- `isRendered()` was the **sole** getter that historically coerced leniently (`Boolean.valueOf(eval(...).toString())`); #5754 is what flipped it to a strict cast.
- A scan of all `jakarta.faces.component` boolean getters confirms none of the others ever did the `.toString()` round-trip — they read `(Boolean) eval(...)` as far back as they were generated.
- So a `String`-valued, `Object`-typed VE also throws on `disabled`/`readonly`/`immediate`/etc., but it always has, on every prior Faces release. They are not regressions, so they are intentionally left as-is — broadening the leniency there would be a *new* behavior change, not a fix. (On 5.0 these carry the same compiler-inserted checkcast as `rendered`, again with unchanged behavior.)

## Fix

Coerce leniently via `Util.toBoolean()`:

```java
return toBoolean(getStateHelper().eval(PropertyKeys.rendered), true);
```

The common `Boolean` path stays allocation-free (returned directly); only a non-`Boolean` value pays the parse.

In **5.0** this same fallback will additionally log a development-stage warning so the underlying user error gets surfaced and fixed; 4.x stays silent to keep maintenance branches log-output-neutral.

### DIFF
--- a/impl/src/main/java/jakarta/faces/component/UIComponentBase.java
+++ b/impl/src/main/java/jakarta/faces/component/UIComponentBase.java
@@ -25,9 +25,9 @@ import static com.sun.faces.facelets.tag.faces.core.FacetHandler.KEY;
 import static com.sun.faces.util.Util.isAllNull;
 import static com.sun.faces.util.Util.isAnyNull;
 import static com.sun.faces.util.Util.isEmpty;
+import static com.sun.faces.util.Util.toBoolean;
 import static jakarta.faces.application.Resource.COMPONENT_RESOURCE_KEY;
 import static java.beans.Introspector.getBeanInfo;
-import static java.lang.Boolean.TRUE;
 import static java.lang.Character.isDigit;
 import static java.lang.Character.isLetter;
 import static java.lang.Thread.currentThread;
@@ -401,7 +401,10 @@ public abstract class UIComponentBase extends UIComponent {
         if (cachedIsRendered != null) {
             return cachedIsRendered;
         }
-        return (Boolean) getStateHelper().eval(PropertyKeys.rendered, TRUE);
+        // Coerce leniently rather than casting: a "rendered" value expression bound with a non-Boolean expected
+        // type (e.g. via a templating layer) can evaluate to a String, which a direct (Boolean) cast rejects.
+        // Tolerating it is not spec-required but is kept for backward compatibility; a Boolean costs nothing here.
+        return toBoolean(getStateHelper().eval(PropertyKeys.rendered), true);
     }
 
     @Override
--- a/impl/src/main/java/jakarta/faces/component/UIComponentBase.java
+++ b/impl/src/main/java/jakarta/faces/component/UIComponentBase.java
@@ -404,6 +404,7 @@ public abstract class UIComponentBase extends UIComponent {
         // Coerce leniently rather than casting: a "rendered" value expression bound with a non-Boolean expected
         // type (e.g. via a templating layer) can evaluate to a String, which a direct (Boolean) cast rejects.
         // Tolerating it is not spec-required but is kept for backward compatibility; a Boolean costs nothing here.
+        // Mojarra 5.0 should log a warning on this and Mojarra 6.0 should remove this leniency.
         return toBoolean(getStateHelper().eval(PropertyKeys.rendered), true);
     }
 
--- a/impl/src/test/java/jakarta/faces/component/UIComponentTestCase.java
+++ b/impl/src/test/java/jakarta/faces/component/UIComponentTestCase.java
@@ -1642,6 +1642,7 @@ public class UIComponentTestCase extends JUnitFacesTestCaseBase {
     // Tests #5801: a "rendered" value expression bound with a non-Boolean expected type (e.g. through a
     // templating layer) can evaluate to a String; isRendered() must coerce it leniently rather than throwing
     // ClassCastException. The Boolean path must stay unchanged.
+    // NOTE: not spec-compliant, so this cannot end up in the TCK.
     @Test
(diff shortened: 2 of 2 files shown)


pizzi80:
I have some concerns about this PR:

1) is not easier to change the Glassfish admin UI variable from String to Boolean instead of all these custom convertions?

2) is Mojarra a Glassfish custom jsf implementation to accomodate their admin panel bugs?

3) they use ancient code here and there, probaly they could run claude over the codebase to remove old things

imho these things are the one that on the long term produce 
the same performance problems that you're trying to resolve 

for example Portlet is the next thing to standardize or remove from Faces

BalusC:
> 1. is not easier to change the Glassfish admin UI variable from String to Boolean instead of all these custom convertions?

GF admin UI might be not the only app. There might be other apps. This change preserves backwards compatibility and doesn't add back performance penalty. Note that the plan is to add a dev warning log in 5.0 and to remove this in 6.0

> 2. is Mojarra a Glassfish custom jsf implementation to accomodate their admin panel bugs?

GlassFish admin UI uses Woodstock which is a super ancient component library. It's the component library used behind the legacy "Visual JSF Editor" built into Netbeans some 20 years ago.

> 3. they use ancient code here and there, probaly they could run claude over the codebase to remove old things

Yes, they must indeed migrate their admin UI to something like PrimeFaces.

BalusC:

BalusC:
For reference, 5.0 will now log a warning here: https://github.com/jakartaee/faces/compare/651d209db8e486b692494a2a194de1467082c71a...7f5353aae3a11d66cf513603564f2e09820a1f85#diff-9c6f4e314dc79dfb08783265e9d19fb2c54c0fa8070c4c29b0d3f1c64514b2aa

arjantijms:
Absolutely agree the GlassFish Admin UI must fix this bug inside its code too, and so I indeed did. See 

https://github.com/eclipse-ee4j/glassfish/commit/12d65aee1980d093a309313c4998eb6c66a2751a#diff-587dd189b997a07872f1f0526a8fc818625ca7fb7543e2b80d4b573e74ba5faeR44

And also agree, the Admin UI should switch to PrimeFaces and normal Facelets. It's been on the TODO for at least 2018/2019.

PR #5811, opened 2026-06-24
Skip redundant render-time facelet re-apply for static views

BalusC:
#5753 

## Problem

`RenderResponsePhase.execute` calls `vdl.buildView()` unconditionally before rendering. On a postback the view is already built (from Restore View), so `FaceletViewHandlingStrategy.buildView` takes its `isViewPopulated` branch and **re-applies the entire facelet a second time** — re-running every tag handler, re-resolving every tag-attribute EL expression and regenerating ids. For a view with no build-time-dynamic content this reproduces an identical tree: pure waste.

JFR profiling (Tomcat, `form-inputs` scenario, 20000 postbacks, server state saving) attributes **~14.5% of render** to this redundant re-apply — the bulk of the render-phase gap versus MyFaces, which avoids it by defaulting `org.apache.myfaces.REFRESH_TRANSIENT_BUILD_ON_PSS=auto` (gating re-apply behind `isFilledView`). The rest of render — writers, reflective `AttributesMap`, EL, converters, state saving — is already at MyFaces parity.

## Change

Adds an opt-in boolean context param `org.glassfish.mojarra.disableRefreshTransientBuild` (default `false`, so behaviour is unchanged unless enabled). When enabled, the populated branch early-returns instead of re-applying when **all** of the following hold:

- the view is non-transient (excludes stateless views),
- partial state saving is active (excludes client-side state saving),
- there are no dynamic add/remove actions, and
- the view contains no build-time-dynamic content.

The last condition is detected **request-scoped**: the JSTL conditional/iteration handlers (`c:if`, `c:choose`, `c:forEach`, `c:catch`, `c:set`) and dynamic `ui:include`/`ui:decorate`/`ui:composition` (non-literal `src`/`template`) call the new `TagHandlerImpl.markDynamicTransientBuild` at the top of their `apply()`, setting a `FacesContext` attribute. The flag is set during the restore-time build and read during the render-time build of the same request. Runtime-iteration components (`ui:repeat`, `h:dataTable`) are correctly **not** flagged — their per-row children are built at encode time, not via the facelet re-apply.

## Results

| | render avg | JFR re-apply share |
|---|---|---|
| param **off** (current behaviour) | 720 µs | 14.5% |
| param **on** | **631 µs** | **0.4%** |
| MyFaces 4.1 | 630 µs | 2.6% |

Render reaches MyFaces parity. Verified behaviour-neutral with the param off (full impl unit suite green) and the full Faces TCK passes with the param enabled.

## Notes

- Second commit is an unrelated test-harness fix: the `tomcat-myfaces` perf profile was missing the `provided` compile-time `jakarta.*` APIs that the `tomcat` profile declares, so the perf WAR could not compile under it. Needed to run the cross-implementation comparison that motivated this change.
- A follow-up (not in this PR) can make the *unavoidable* `c:forEach` re-apply cheaper for unrolled views (MyFaces-style `IterationState`/id-base reuse), the complementary lever for the `*-unrolled` scenarios.

### DIFF
--- a/impl/src/main/java/org/glassfish/mojarra/RIConstants.java
+++ b/impl/src/main/java/org/glassfish/mojarra/RIConstants.java
@@ -40,6 +40,14 @@ public class RIConstants {
 
     public static final String SAVED_STATE = RI_PREFIX + "savedState";
 
+    /**
+     * Request-scoped flag set during a view build when a build-time-dynamic handler (a JSTL conditional/iteration or a
+     * dynamic ui:include/ui:decorate/ui:composition) participates, marking the view as one whose facelet must be
+     * re-applied on every (re)build. Read by {@code FaceletViewHandlingStrategy.buildView} to decide whether the
+     * redundant render-time re-apply may be skipped (see {@code disableRefreshTransientBuild}).
+     */
+    public static final String DYNAMIC_TRANSIENT_BUILD = RI_PREFIX + "dynamicTransientBuild";
+
     /*
      * <p>TLV Resource Bundle Location </p>
      */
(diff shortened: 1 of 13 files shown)


BalusC:

PR #5826, opened 2026-07-05
Add view-unrolled-nested perf scenario

BalusC:
Adds a `view-unrolled-nested` scenario to the `test/perf` bench: a nested `c:forEach` (outer over `dataBean.groups`, inner over `group.rows`) built as a static tree, mirroring `view-unrolled` but with a `c:forEach` inside a `c:forEach`.

## Why

`view-unrolled` is a single flat `c:forEach`. Nesting one loop inside another makes the inner tag build under an already-unrolled parent, which exercises a different build/re-apply render path that the flat loop never reaches — worth isolating as its own data point.

## What it surfaces

On Tomcat (1000 runs/scenario, warmup 50), the nested scenario widens the Mojarra-vs-MyFaces RENDER gap that the flat twin only hints at:

| scenario | RESTORE (Moj vs MyFaces) | RENDER (Moj vs MyFaces) |
|---|--:|--:|
| `view-unrolled` (flat) | −7% | +12% |
| `view-unrolled-nested` | −3% (parity) | **+35%** |

RESTORE is at parity; the entire deficit is in RENDER. This gives a reproducible target for closing the nested-render gap.

## Details

- Reuses the existing `groups` model (5×10, same as `repeat-nested`/`table-nested`/`composite-nested`) — no `DataBean` change.
- Registered as a full (non-ajax) postback in `PerfBenchIT`.
- Bumped `numberOfLogicalViews` / `NUMBER_OF_VIEWS_IN_SESSION` 16 → 17 (one more retained postback view per loop).
- `index.xhtml` nav link + `README.md` scenario entry.

Bench-only; no production code touched.


BalusC:
Closing off. We need yet better coverage. New PR coming.

BalusC:
Superseded by #5827, which folds this scenario into a full component-family × variant matrix (this page is now `foreach-nested`).

PR #5829, opened 2026-07-06
Reuse built-in by-type converters per target class

BalusC:
`Application.createConverter(Class)` re-ran the CDI lookup, the type-resolution walk, reflective instantiation and annotation application on **every** call — i.e. on every by-type conversion (`Integer`, `BigDecimal`, `Boolean`, `Enum`, … resolved via `converter-for-class`).

This caches the resolved standard `jakarta.faces.convert` converter per target class. The package guard only admits the stateless built-ins; custom `converter-for-class` registrations stay on the per-call path (they may be stateful or scoped). `EnumConverter` is cached safely too — the cache is keyed by **target class**, so each enum type gets its own `EnumConverter(SomeEnum)` instance.

### Benchmark (Tomcat, 50 warmup / 1000 runs, vs MyFaces 5.0)

Measured together with the complementary `jakarta.faces-api` formatter cache (jakartaee/faces#2199) — the two are complementary; both cut render-time converter allocation. See #5753.

| phase | Mojarra 5.0 | MyFaces 5.0 | Δ |
|---|--:|--:|--:|
| RENDER_RESPONSE | 36.9s | 39.2s | **−6%** |
| suite total | 71.1s | 71.5s | **−1%** |

Render-heavy views converting currency + date on every row now beat MyFaces by −14…−18% (`table-readonly`, `repeat-readonly`, `table-build`, `repeat-build`).

### DIFF
--- a/impl/src/main/java/org/glassfish/mojarra/application/applicationimpl/InstanceFactory.java
+++ b/impl/src/main/java/org/glassfish/mojarra/application/applicationimpl/InstanceFactory.java
@@ -152,6 +152,9 @@ public class InstanceFactory {
     private final ViewMemberInstanceFactoryMetadataMap<String, Object> validatorMap;
     private final Map<Class<?>, Constructor<?>> constructorCache;
 
+    // Standard built-in (jakarta.faces.convert) by-type converters, reused per target class (see createConverter).
+    private final Map<Class<?>, Converter<?>> builtinConverterByType = new ConcurrentHashMap<>();
+
     private final Set<String> defaultValidatorIds;
     private volatile Map<String, String> defaultValidatorInfo;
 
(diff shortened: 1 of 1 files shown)


pizzi80:
there is a chance to further improve caching:

 
if the converter is CDI converted and is `@ApplicationScoped` ... could be cached ....

BalusC:
An explicit `@ApplicationScoped` is basically already "cached" (application scoped) by CDI itself.

The current PR is more for built-in (implicit) by-type converters: IntegerConverter, LongConverter, BigDecimalConverter, EnumConverter, etc which are not CDI-managed in first place.

It are these automatic converters which are used in e.g. `<h:inputText value="#{bean.integer}">` without explicit converter.

pizzi80:
perfect! 

I can finally remove `@ApplicationScoped` on  built-in (implicit) by-type  converters in my fork :v: 

thanks!!!!!

PR #5849, opened 2026-07-11
Plain Java type and concurrency micro-improvements

kamilkrzywanski:
## Summary

Small, behavior-preserving plain-Java cleanups related to performance work around [#5753](https://github.com/eclipse-ee4j/mojarra/issues/5753). No lifecycle, EL, or public API changes—only collection/string/concurrency primitives on existing hot paths.

### Changes

| Location | Change | Rationale |
|----------|--------|-----------|
| `UIRepeat` | `StringBuffer` → `StringBuilder` | Client-id building is single-threaded per component; no need for synchronized buffer. |
| `TextUnit` | `StringBuffer` → `StringBuilder`; `Stack` → `ArrayDeque` | Facelets compile path is not shared across threads; `Stack` is Vector-backed. |
| `CompilationManager` | `Stack` → `ArrayList` used as a stack | Needs push/pop at the end **and** random access for the root unit at index 0; `ArrayList` fits both without Vector sync. |
| `ComponentValidators` | `LinkedList` → `ArrayList` | Stack ops at the end plus reverse random access when applying validators; `ArrayList` is better for both. |
| `IterationIdManager` | Drop class-level `synchronized`; `LinkedList` → `ArrayDeque` | `FaceletContext` is request-scoped and used on a single request thread, so global class locking only added contention. |
| `ApplicationResourceBundle` | DCL + `HashMap` → `ConcurrentHashMap.computeIfAbsent` | Locale → bundle cache can be hit concurrently across requests. |
| `DefaultFunctionMapper` | Partial sync on `HashMap` → DCL init + `ConcurrentHashMap` | Previous code synchronized only `put`, not `get`; map is now concurrent for both. |
| `OutputMessageRenderer` | `MessageFormat` + `StringBuffer` → `format(Object[])` | Avoids the synchronized `StringBuffer` overload while keeping the same formatted result. |

## Test plan

- [x] `mvn -pl impl -am compile -DskipTests`
- [x] `mvn -pl impl -Dtest=UIRepeatTest test`
- [x] CI full suite on this PR


BalusC (dismissed):
[Faces 5.0 TCK](https://github.com/jakartaee/faces/tree/5.0/tck) is failing on this branch. The changes in `IterationIdManager` broke Facelets compilation because `ArrayDeque` doesn't support `null` elements. NPE all over place in `tck/faces20/coretags`, `tck/faces20/templating`, `tck/faces20/viewhandler`, etc. 

**Requested change**:
- Please revert `ArrayDeque` in `IterationIdManager` back to `LinkedList`.

How to run Faces 5.0 TCK yourself the next time in order to verify if your changes don't break anything:
- assuming https://github.com/jakartaee/faces is cloned into `~/git/faces` and at `5.0` branch
- assuming https://github.com/eclipse-ee4j/mojarra is cloned into `~/git/mojarra` and at `master` (5.0) branch
- `cd ~/git/faces/api` then `mvn clean install`
- `cd ~/git/mojarra/impl` then `mvn clean install`
- `cd ~/git/faces/tck` then `mvn clean verify` (can take up to a hour, but you can use `-Tn` to speed up depending on amount of your RAM; `n = RAM/4` is generally a good direction, e.g. `mvn clean verify -T4` when having 16GB or `mvn clean verify -T8` when having 32GB)

Next issue, it didn't cause TCK failure, but the changes to `DefaultFunctionMapper` are odd:
- DCL was introduced while DCL was removed from `ApplicationResourceBundle` in same PR. Seems inconsistent? Better to remove DCL altogether.
- The `functions` field must be marked `volatile` for correct DCL behavior. But it's better to just remove DCL.
- This class is not in the hot path. It doesn't show up in JFR profiling.
- It turns out that this class is even not used at all; it was dead on arrival while integrating Facelets into Mojarra around August 2008 (https://github.com/javaee/mojarra/commit/a095ff305f118f7afca70a3733e6917e308e3e29), because it was superseded by Mojarra-specific `NamespaceHandler` and `CompositeFunctionMapper`. This unused class should just be removed.

**Requested change**: you have 2 options:
- Revert `DefaultFunctionMapper` back to original -- I will remove it myself later.
- _OR_ remove `DefaultFunctionMapper` -- it's nowhere used.

All other changes look good. Thank you! :)

On an unrelated note, when using AI to change the code and create commits/PRs, please explicitly mention the AI model+version used. E.g. "Claude Code (Opus 4.8)". This matters for traceability and accountability. Preferably also include it as co-author in commit message.

BalusC:
Full TCK confirmed green. Thank you!

BalusC (approved):

BalusC:
~Regression: #5989~

Nope. Nevermind.

PR #5853, opened 2026-07-12
Backport dynamic-component and state-saving perf improvements to 4.0 (#5848, #5849, #5851, faces#2209)

BalusC:
Backports four already-merged performance improvements onto 4.0. All were found while profiling #5753.

Because 4.0 uses the `com.sun.faces` package with `jakarta.faces` inlined (master uses `org.glassfish.mojarra` plus the split-out faces submodule), none of these could be cherry-picked; each was ported by hand and adapted to the 4.0 state-saving layout.

| Commit | Backport of | Summary |
|--------|-------------|---------|
| Skip the redundant dynamic-child reorder in buildView | #5848 | Return early when every dynamic child is already at its stored index, avoiding the O(n²) remove-all/re-add-all on every buildView. |
| Replace legacy collection and string types on hot paths | #5849 | `StringBuilder`, `ArrayList`/`ArrayDeque`, `ConcurrentHashMap` in place of synchronized `StringBuffer`/`Stack`/`HashMap` or `LinkedList`; drop the dead `DefaultFunctionMapper`. |
| Improve dynamic component add/remove performance | #5851 | Remove the dead `TREE_HAS_DYNAMIC_COMPONENTS` flag and collapse `DYNAMIC_CHILD_COUNT` to a set-once marker (~1000 fewer reflective `AttributesMap` ops per 500-child toggle). |
| Persist system-event listeners in the partial-state delta only when changed | jakartaee/faces#2209 | Gate the listener-map save on a `systemEventListenersModified` flag so `@ListenerFor`/`f:event` listeners (re-subscribed by buildView) stay out of the delta, keeping otherwise-static components on the view-root-only restore fast path. |

The only non-mechanical adaptation is the faces#2209 port: 4.0's field-backed value-expression state already uses this exact pattern under the name `bindingsModified`, so the system-event port mirrors it (`bindingsModified` / `rendererTypeSet`), and the partial-state discriminator is `values.length != 7` (4.0 partial delta = 6, full state = 7). Full state saving is unaffected in all four.

Impl compiles clean. Full TCK run still pending.


BalusC:
~Regression: #5989~

Nope. Nevermind.

PR #5859, opened 2026-07-16
Read resource content in bulk while scanning for EL expressions

### LINKED ISSUE #5858: ELEvaluatingInputStream has no bulk read(byte[],int,int): faces.js and CSS containing EL are streamed one synchronized byte at a time
Discovered while profiling the OmniFaces `CombinedResourceHandler`, but the mechanism is generic to Mojarra's resource pipeline.

## Problem

`ResourceHelper$ELEvaluatingInputStream` implements only `read()`. It inherits `InputStream`'s default `read(byte[],int,int)`, which loops calling `read()` once per byte. The underlying `ByteArrayInputStream.read()` is `synchronized`, so every byte costs a monitor acquire/release plus two virtual calls. An N-byte resource costs N synchronized single-byte reads per fetch.

`ResourceHelper` wraps it as `new BufferedInputStream(new ELEvaluatingInputStream(...))` (lines 171 and 191), but buffering cannot help: `BufferedInputStream` requests a bulk read, which lands on the inherited default, which loops per byte anyway.

Verified stack from a JFR recording:

```
java.io.ByteArrayInputStream.read()                                line: 146   <- single-byte, synchronized
com.sun.faces.application.resource.ResourceHelper$ELEvaluatingInputStream.read()   line: 578
java.io.InputStream.read(byte[], int, int)                         line: 296   <- default impl, loops per byte
java.io.BufferedInputStream.read1(byte[], int, int)                line: 345
java.io.BufferedInputStream.implRead(byte[], int, int)             line: 420
```

## Scope

`resourceSupportsEL()` (`ResourceHelper` line 466) returns true when the content type is `text/css` (the sole entry in `EL_CONTENT_TYPES`) or the resource is exactly `jakarta.faces:faces.js`.

Importantly, the wrapper self-corrects: `ELEvaluatingInputStream.close()` calls `info.disableEL()` when the stream reached EOF and no expression was ever evaluated, so later fetches take the plain `getNonCompressedInputStream` branch. That narrows the impact to:

- **CSS without EL** — pays this only on the first full read, then self-corrects. Not an ongoing cost.
- **CSS containing EL** (e.g. `#{resource['image:background.png']}`) — `expressionEvaluated` is set, so `disableEL()` never fires; every fetch is byte-at-a-time.
- **faces.js** — ships with 6 EL expressions (`#{applicationScope["com.sun.faces.mojarraVersion"].specversion}`, `#{facesContext.namingContainerSeparatorChar}`, ...), so it is never disabled. **Every fetch of faces.js, in every Faces application, is streamed one synchronized byte at a time** — 37126 bytes, so ~37000 single-byte reads per fetch.

Plain `.js` resources are unaffected (not `text/css`, not `faces.js`).

## Measurement

Tomcat 11.0.22 + Mojarra 4.1.10, JFR `settings=profile` with `-XX:+DebugNonSafepoints`, 20,000 requests at concurrency 8 against a single resource containing faces.js:

- **77%** of execution samples have `ByteArrayInputStream.read()` as the leaf
- \+ 5% `ELEvaluatingInputStream.read()`, + 3.4% `InputStream.read(byte[],int,int)`
- => **~86% of the CPU spent serving that resource is byte-at-a-time plumbing**

Actual EL evaluation, by contrast, is negligible:

- `readExpressionIntoBufferAndEvaluateIntoBuffer` 0.46%
- `evaluateExpressionIntoBuffer` 0.38%
- => **~0.8%**

So ~99% of the cost is *scanning* for `#{`, not evaluating the 6 expressions that are there. (4.1.9 measures the same: 79% / ~87%.)

## Proposed fix

Override `read(byte[], int, int)` in `ELEvaluatingInputStream`: pull a chunk from `inner` with a bulk read and scan the buffer for `#{`, dropping to the existing per-byte handling only while inside an expression. Scanning a buffer for `#{` is as correct as scanning a byte stream, so semantics are unchanged. This turns N synchronized `inner.read()` calls per fetch into N/8192.

Secondary, same class: `buf` is a `List<Integer>`, boxing each byte — only on the EL path, so not part of the measurement above.

BalusC:
Fixes #5858.

`ELEvaluatingInputStream` implemented only `read()`, so it inherited `InputStream`'s default `read(byte[],int,int)`, which loops calling `read()` once per byte. The inner stream's `read()` is typically `synchronized` (`ByteArrayInputStream`), so every byte of a served resource cost a monitor acquire/release plus two virtual calls. The surrounding `new BufferedInputStream(new ELEvaluatingInputStream(...))` could not help, because the bulk read lands on the inherited default, which loops per byte anyway:

```
java.io.ByteArrayInputStream.read()                          line: 146   <- single-byte, synchronized
...ResourceHelper$ELEvaluatingInputStream.read()             line: 578
java.io.InputStream.read(byte[], int, int)                   line: 296   <- default impl, loops per byte
java.io.BufferedInputStream.read1(byte[], int, int)          line: 345
```

## Change

Buffer `inner` in 8K chunks, and add a `read(byte[],int,int)` override which copies the run of bytes up to the next `'#'` straight out of that buffer, leaving every `'#'` to the existing single byte state machine. Scanning and expression evaluation semantics are unchanged — a `'#'` is still the only thing that can start an expression, and it is still the original code that decides.

## Who pays this today

`resourceSupportsEL()` is true for `text/css` and for `jakarta.faces:faces.js`. EL-free CSS takes the `disableEL()` shortcut in `close()` after its first full read, so it self-corrects. What does not self-correct:

- **faces.js** — it ships 6 EL expressions (`#{applicationScope["com.sun.faces.mojarraVersion"].specversion}`, `#{facesContext.namingContainerSeparatorChar}`, ...), so `expressionEvaluated` is always set and it is re-scanned byte-by-byte on **every** fetch, in every Faces application.
- **CSS containing EL** (e.g. `#{resource['image:background.png']}`) — same.

## Measurement

Tomcat 11.0.22, JFR `settings=profile` with `-XX:+DebugNonSafepoints`, 20000 requests at concurrency 8 against a single ~50K resource containing faces.js:

| | before | after |
|---|---|---|
| throughput | 5780 req/s | **9992 req/s** (1.73x) |
| latency | 1.384 ms | **0.801 ms** (-42%) |
| `ByteArrayInputStream.read` | **75.6%** of execution samples | gone from the profile |
| failed requests | 0 | 0 |

For reference, before the change ~82-86% of the CPU spent serving that resource was byte-at-a-time plumbing, while actual EL evaluation (`readExpressionIntoBufferAndEvaluateIntoBuffer` + `evaluateExpressionIntoBuffer`) was ~0.8% — i.e. nearly all of it was *looking for* `#{`, not evaluating it.

## Correctness

- **faces.js** (EL evaluated): output **byte-identical** before vs after, 51259 bytes, same build compared both ways.
- **EL-free CSS** (the `disableEL()` path): output **byte-identical**, and the second fetch — after `disableEL()` has fired — is identical too.
- No unevaluated `#{` left in the served output.
- Full 5.0 TCK green with this change.

The measurement above was taken with this same change applied to the 4.1 line (identical apart from the package rename), since the harness driving it is wired for Faces 4.1.

### DIFF
--- a/impl/src/main/java/org/glassfish/mojarra/application/resource/ResourceHelper.java
+++ b/impl/src/main/java/org/glassfish/mojarra/application/resource/ResourceHelper.java
@@ -541,6 +541,12 @@ public abstract class ResourceHelper {
         private boolean expressionEvaluated;
         private boolean endOfStreamReached;
 
+        // Pull from inner in bulk; it is read one byte at a time while scanning for '#{', and inner.read() is typically
+        // synchronized (e.g. ByteArrayInputStream), so a per-byte call to it costs a monitor on every byte of the resource.
+        private final byte[] innerBuf = new byte[8192];
+        private int innerPos;
+        private int innerLimit;
+
         // ---------------------------------------------------- Constructors
 
         public ELEvaluatingInputStream(FacesContext ctx, ClientResourceInfo info, InputStream inner) {
@@ -571,16 +577,16 @@ public abstract class ResourceHelper {
                     i = buf.remove(0);
                 } else {
                     writingExpression = false;
-                    i = inner.read();
+                    i = readInner();
                 }
             } else {
                 // Read a character.
-                i = inner.read();
+                i = readInner();
                 c = (char) i;
                 // If it *might* be an expression...
                 if (c == '#') {
                     // read another character.
-                    i = inner.read();
+                    i = readInner();
                     c = (char) i;
                     // If it's '{', assume we have an expression.
                     if (c == '{') {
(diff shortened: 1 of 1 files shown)


BalusC:
TCK green

Issue #4269, opened 2017-08-18
ComponentStateHelper#restoreState should not remove attributesThatAreSet


ren-zhijun-oracle:
ComponentStateHelper#restoreState removes the `attributeThatAreSet` in following code:

                defaultMap.remove(serializable);
                deltaMap.remove(serializable);

as a result all pass-through attributes are lost in subsequent ajax request rendering (for example rendering of outputTextLink). Attribute removal code was added in v2.3.

PR #5881, opened 2026-07-21
Fix #5880: merge attributesThatAreSet delta on restore instead of replacing

### LINKED ISSUE #5880: `ComponentStateHelper` delta list for `attributesThatAreSet` replaces instead of merges on restore, permanently dropping pass-through attributes after the 2nd postback (`styleClass` newly affected since 4.0.19/4.1.10)
Reported against 4.1.11.

Line numbers below are from `4.1.11-RELEASE`. `ComponentStateHelper.java`, `UIComponent.java`, `RenderKitUtils.java` and `TextRenderer.java` are byte-identical between that tag and the current branch `4.1`, so they apply unchanged there too.

Two things are tangled here, and it is worth separating them up front:

- The underlying `ComponentStateHelper` defect is **old**. It is the same one reported in #4269 ("`ComponentStateHelper#restoreState` should not remove `attributesThatAreSet`", 2021), which was auto-closed as stale without a fix, a mechanism, or a reproducer. `add`, `initList` and `restoreState` are identical on 4.0 and 4.1 today. Pass-through attributes (`style`, `title`, `dir`, `lang`, `tabindex`, `on*`) have been losing state this way for years.
- `styleClass` specifically is a **4.0.19/4.1.10 regression**, from the #5753 performance work — commit efa7445b22 "Track styleClass and skip its reflective read when unset" (2026-06-12, merged to 4.1 via f396d7f92c, first released in `4.1.10-RELEASE`). Before it, `TextRenderer.java:94` read `component.getAttributes().get("styleClass")` directly and `HtmlInputText.setStyleClass` did not call `handleAttribute`, so `styleClass` was immune to this corruption. The commit notes "Rendering is unchanged" — which holds only while `attributesThatAreSet` is trustworthy, and under partial state saving it is not.

So 4.1.10 did not introduce the bug, it enlarged the blast radius to the most visible attribute there is. Compare 7bc7d103a5 "Fix link target/onclick regression from attr gating" (2026-06-11), a second regression from the same gating series.

## Symptom

An `<h:inputText styleClass="foo bar"/>` renders without its `class` attribute from the **second** ajax postback onward, and stays broken for the rest of the view's life. Any component whose renderer consults `attributesThatAreSet` is affected, and it is not limited to `styleClass` — `style`, `title`, `dir`, `lang`, `tabindex` and all `on*` handlers go the same way.

The trigger is any code that calls `setValueExpression(name, ve)` with a **non-literal** `ve` on an already-built component after the view has been restored — e.g. from a `PostRestoreStateEvent` listener. Originally found via OmniFaces `<o:outputLabel>` (omnifaces/omnifaces#972), which sets the `label` binding on its `for` target from `processEvent`, but nothing about that is OmniFaces-specific.

## Reproducer (no third-party libraries)

```xhtml
<h:form id="test">
    <h:inputText id="input" value="#{bean.value}" styleClass="foo bar">
        <f:event type="postRestoreState" listener="#{bean.onRestore}"/>
    </h:inputText>
    <h:commandButton id="save" value="SAVE">
        <f:ajax execute="@form" render="@form"/>
    </h:commandButton>
</h:form>
```

```java
@Named @ViewScoped
public class Bean implements Serializable {

    private String value;

    public void onRestore(ComponentSystemEvent event) {
        var context = FacesContext.getCurrentInstance();
        var expression = context.getApplication().getExpressionFactory()
            .createValueExpression(context.getELContext(), "#{bean.label}", String.class);
        event.getComponent().setValueExpression("label", expression);
    }

    public String getLabel() { return "Label"; }
    public String getValue() { return value; }
    public void setValue(String value) { this.value = value; }
}
```

Press SAVE twice. After the first postback the input still renders `class="foo bar"`; after the second the `class` attribute is gone and never comes back.

Note the expression must be non-literal. Passing `createValueExpression("Label", String.class)` takes a different branch in `setValueExpression` and does not reproduce.

## Mechanism

**1.** `UIComponent.setValueExpression` records the attribute name in the private `attributesThatAreSet` StateHelper **list** when the expression is not literal text — `jakarta/faces/component/UIComponent.java:368-378`:

```java
if (binding != null) {
    if (!binding.isLiteralText()) {
        List<String> sProperties = (List<String>) getStateHelper().get(PropertyKeysPrivate.attributesThatAreSet);
        if (sProperties == null) {
            getStateHelper().add(PropertyKeysPrivate.attributesThatAreSet, name);
        } else if (!sProperties.contains(name)) {
            getStateHelper().add(PropertyKeysPrivate.attributesThatAreSet, name);
        }
```

**2.** `ComponentStateHelper.add` seeds the delta entry with a **fresh empty list** rather than a copy of the current default list — `jakarta/faces/component/ComponentStateHelper.java:242-264`:

```java
public void add(Serializable key, Object value) {
    initList(key);
    if (component.initialStateMarked()) {
        ((List<Object>) deltaMap.get(key)).add(value);
    }
    List<Object> items = (List<Object>) get(key);
    items.add(value);
}

private void initList(Serializable key) {
    if (component.initialStateMarked()) {
        deltaMap().computeIfAbsent(key, e -> new ArrayList<>(4));   // empty, not a copy of the default list
    }
    ...
}
```

So after the listener runs, the default list is `[styleClass, label]` but the saved delta is just `[label]`.

**3.** `restoreState` treats a restored `List` as a wholesale **replacement** — `ComponentStateHelper.java:352-365`:

```java
if (value instanceof Map) {
    for (Map.Entry<String, Object> entry : ((Map<String, Object>) value).entrySet()) {
        put(serializable, entry.getKey(), entry.getValue());     // merges
    }
} else if (value instanceof List) {
    if (defaultMap != null) { defaultMap.remove(serializable); } // discards
    if (deltaMap != null)   { deltaMap.remove(serializable); }
    List<?> values = (List<?>) value;
    values.stream().forEach(o -> add(serializable, o));
}
```

The list Facelets has just rebuilt during `buildView` — which correctly contains `styleClass`, because the tag handlers re-execute on every postback and `HtmlInputText.setStyleClass` re-registers the name via `HtmlComponentUtils.handleAttribute` — is thrown away and replaced by the partial delta `[label]`.

Note the asymmetry: the `Map` branch merges, the `List` branch replaces. The `Map` branch is the one that matches delta semantics.

**4.** The renderer then skips the attribute — `com/sun/faces/renderkit/RenderKitUtils.java:431-436`:

```java
public static Object getAttributeIfSet(UIComponent component, List<String> setAttributes, String name) {
    if (component.getClass().getName().startsWith(OPTIMIZED_PACKAGE)) {
        return setAttributes != null && setAttributes.contains(name) ? component.getAttributes().get(name) : null;
    }
    return component.getAttributes().get(name);
}
```

`html_basic/TextRenderer.java:95` reads `styleClass` through the single-read convenience overload of that method and only writes `class` (lines 142 and 160) when the result is non-null. The `styleClass` **value** is never lost — it sits untouched under `PropertyKeys.styleClass` the whole time. Only the "was it set" bookkeeping is destroyed.

## Why the second postback, and why it is permanent

`PostRestoreStateEvent` is delivered from `RestoreViewPhase` *after* `buildView` has called `markInitialState` and after the state manager has applied the delta, so the poisoned delta written at the end of postback N is only applied at the start of postback N+1.

| request | after `buildView` | after `restoreState` | after `PostRestoreState` | renders `class` | delta saved |
|---|---|---|---|---|---|
| initial GET | `[styleClass]` | — | `[styleClass]` (no event on GET) | yes | — |
| postback 1 | `[styleClass]` | `[styleClass]` | `[styleClass, label]` | yes | `[label]` |
| postback 2 | `[styleClass]` | **`[label]`** | `[label]` | **no** | `[label]` |
| postback 3+ | `[styleClass]` | `[label]` | `[label]` | no | `[label]` |

It never recovers: from postback 2 on, `sProperties.contains("label")` is true, so `setValueExpression` stops calling `add`, and the delta stays `[label]` forever.

## Suggested fix

Minimal and consistent with the `Map` branch — make the `List` branch in `restoreState` merge into the existing list instead of discarding it, adding only names not already present. That preserves whatever `buildView` legitimately re-registered this request while still applying the delta.

Alternatively, fix it at the source by having `initList` seed the delta list from the current default list, so the saved delta is complete rather than incremental. That also works, at the cost of a slightly larger view state.

Reverting the `styleClass` half of efa7445b22 would make this particular report go away, but I would not treat that as the fix — it leaves `style`, `title`, `dir`, `lang` and the `on*` handlers corrupting exactly as #4269 described in 2021.

BalusC:
Fixes #5880.

`ComponentStateHelper.restoreState` treated a restored `List` as a wholesale replacement: it discarded both the default and delta entries for the key, then re-added only the saved delta. But the delta list is add-only — it carries only entries appended after `markInitialState` — so replacing dropped everything the tag handlers had just re-registered during `buildView`.

For `attributesThatAreSet` this meant that once a non-literal `setValueExpression(name, ve)` ran on a restored component (e.g. from a `PostRestoreStateEvent` listener), the poisoned delta wiped the pass-through attributes from the second postback onward. Any of `styleClass`, `style`, `title`, `dir`, `lang`, `tabindex` and the `on*` handlers stopped rendering and never recovered. `styleClass` became newly affected in 4.0.19/4.1.10 (#5753); the rest go back to #4269 (2021).

## Fix

Merge the restored delta onto the list `buildView` already rebuilt this request, adding only entries not already present, matching the `Map` branch directly above it — which already merges. The `List` path serves exactly two keys (`attributesThatAreSet` and `UIViewRoot.phaseListeners`), both add-only with no overlap between the rebuilt default and the saved delta, so merge-with-dedup is lossless.

## Test

`UIComponentBaseTestCase#testAttributesThatAreSetMergeOnRestore` reproduces the two-postback drop: a tag attribute registered pre-`markInitialState`, a non-literal binding added as the delta, saved, then restored onto a rebuilt list. Asserts the merged `[styleClass, label]`. Verified it fails against the old replace logic (`expected <[styleClass, label]> but was <[label]>`) and passes with the fix; full `UIComponentBaseTestCase` (43 tests) green.

### DIFF
--- a/impl/src/main/java/jakarta/faces/component/ComponentStateHelper.java
+++ b/impl/src/main/java/jakarta/faces/component/ComponentStateHelper.java
@@ -354,15 +354,15 @@ class ComponentStateHelper implements StateHelper, TransientStateHelper {
                     put(serializable, entry.getKey(), entry.getValue());
                 }
             } else if (value instanceof List) {
-                if (defaultMap != null) {
-                    defaultMap.remove(serializable);
+                List<Object> current = (List<Object>) get(serializable);
+                for (Object item : (List<?>) value) {
+                    if (current == null || !current.contains(item)) {
+                        add(serializable, item);
+                        if (current == null) {
+                            current = (List<Object>) get(serializable);
+                        }
+                    }
                 }
-                if (deltaMap != null) {
-                    deltaMap.remove(serializable);
-                }
-
-                List<?> values = (List<?>) value;
-                values.stream().forEach(o -> add(serializable, o));
             } else {
                 put(serializable, value);
                 handleAttribute(serializable.toString(), value);
--- a/impl/src/test/java/jakarta/faces/component/UIComponentBaseTestCase.java
+++ b/impl/src/test/java/jakarta/faces/component/UIComponentBaseTestCase.java
@@ -447,6 +447,39 @@ public class UIComponentBaseTestCase extends UIComponentTestCase {
         assertEquals(Arrays.asList("attr2"), c.getAttributes().get("jakarta.faces.component.UIComponentBase.attributesThatAreSet"));
     }
 
+    /**
+     * The {@code attributesThatAreSet} delta list only carries entries added after
+     * {@code markInitialState()}. On restore it must therefore be <em>merged</em> onto the list that
+     * {@code buildView} already rebuilt for this request, not replace it. Replacing dropped every
+     * attribute the tag handlers re-registered before the delta was applied &mdash; e.g. a
+     * {@code styleClass} rendered on the first postback but gone from the second onward (issue #5880),
+     * triggered by any {@code setValueExpression(name, ve)} with a non-literal {@code ve} on a
+     * restored component, such as from a {@code PostRestoreStateEvent} listener.
+     */
+    @Test
+    public void testAttributesThatAreSetMergeOnRestore() throws Exception {
+        request.setAttribute("foo", "bar");
+
+        // buildView registers a tag attribute; a post-restore binding then adds a non-literal one.
+        ComponentTestImpl c = new ComponentTestImpl();
+        c.getAttributes().put("styleClass", "foo bar");
+        c.markInitialState();
+        c.setValueExpression("label", application.getExpressionFactory().createValueExpression(facesContext.getELContext(), "#{foo}", String.class));
+        assertEquals(Arrays.asList("styleClass", "label"), c.getAttributes().get("jakarta.faces.component.UIComponentBase.attributesThatAreSet"));
+
+        Object state = c.saveState(facesContext); // delta carries only [label]
+
+        // Next postback: buildView rebuilds styleClass before the delta is restored on top of it.
+        c = new ComponentTestImpl();
+        c.getAttributes().put("styleClass", "foo bar");
+        c.markInitialState();
+        c.pushComponentToEL(facesContext, c);
+        c.restoreState(facesContext, state);
+        c.popComponentFromEL(facesContext);
+
+        assertEquals(Arrays.asList("styleClass", "label"), c.getAttributes().get("jakarta.faces.component.UIComponentBase.attributesThatAreSet"));
+    }
+
     @Test
     public void testValueExpressions() throws Exception {
 


pizzi80:
Thanks for the fast fix!

Issue #5856, opened 2026-07-14
Performance review of view build time


BalusC:
#5753 closed the biggest gap with MyFaces. But the remaining gap is in the **view build time**. The latest bench numbers still show relatively big diffs in Restore View phase of *-build and *-get scenarios, although small in absolute time. This gap needs to be reviewed and closed.

pizzi80:
I merged `4.1` HEAD into my fork (let's call it 4.2.11) and used the new perf
suite to compare it against the official 4.1.11.

Caveat: my fork tracks `4.1` HEAD, which is ahead of the 4.1.11 tag,
so it includes also the following PR:
- https://github.com/eclipse-ee4j/mojarra/pull/5879
- https://github.com/eclipse-ee4j/mojarra/pull/5881
- https://github.com/eclipse-ee4j/mojarra/pull/5882

A note on methodology first, because I ran into some pitfalls with the suite:

- **Warmup matters a lot.** At `warmup=50` I was measuring JIT compilation state
  rather than steady-state behaviour — several scenarios flipped sign entirely
  when I raised it. I settled on `warmup=200 runs=2000`.
- **Run order introduces drift**, so I ran A-B-A (fork → official → fork) and
  used the two fork runs as a control: identical bytecode, so any difference
  between them is pure measurement noise.


With that control in place, the deltas sit 3–7× above the noise floor:

| Phase | Official 4.1.11 | Fork (mean of 2 runs) | Δ | Noise floor |
|---|---:|---:|---:|---:|
| `RESTORE_VIEW` | 201 µs | 197.5 µs | **+1.8 %** | 0.6 % |
| `PROCESS_VALIDATIONS` | 199 µs | 196.0 µs | **+1.5 %** | 0.2 % |
| `RENDER_RESPONSE` | 518 µs | 513.5 µs | **+0.9 %** | 0.2 % |
| **Full request** | 1088 µs | 1076.0 µs | **+1.1 %** | 0.4 % |

Positive Δ = fork is faster. Per-scenario figures are *not* reportable — in the
identical-bytecode control they swing by ±10–45 %, so only totals are meaningful.

**One limitation worth noting**: the perf suite is single-threaded, so it doesn't
capture the dimension where part of this work actually applies. Reduced
allocation shows up as GC pause frequency under concurrent load — and those
pauses are paid by every in-flight request, not just the one that allocated.
Lock contention isn't exercised at all.

[...]

PR #5889, opened 2026-07-30
Generate Facelets unique ids from per-tag counter slots

BalusC:
Part of #5856. This takes the flat-markup half of the remaining view-build gap; see the caveat at the end for what it does not take.

### Problem

`buildView` asked a per-request `HashMap` for every component's id counter, keyed by tag id: two hash lookups, a `HashMap.Node`, `Integer` boxing, and — as the map grows to one entry per component — repeated resizes.

That counter only earns its keep when a tag id *repeats*, which happens inside `c:forEach`, `ui:repeat` and composites. Ordinary markup uses each tag id once, so a flat view paid for the map and got nothing back. Allocation profiling of a bare build attributed **444 of 492 `HashMap$Node` allocations** to this one call site.

This is also why the deficit never showed up in the perf suite: every scenario in it builds through `c:forEach`, `ui:repeat`, `h:dataTable` or composites, i.e. exactly the shapes where the map stays small. Measured in isolation, flat markup was **+54% per component** against MyFaces (851 vs 551 ns), scaling linearly with view size — a 5000-component view paying ~1.5 ms of it per request.

### Fix

Reserve a dense counter slot per tag from its `Facelet`, and hold a build's counters in an `int[]` indexed by that slot, so counting an id costs an array increment.

Two details worth flagging for review:

- The slot binds to the Facelet **being applied**, not the one the tag was compiled in — a `ui:define` body applies under the template it is inserted into, and a tag that alternates between templates simply rebinds.
- Slots are keyed by tag id, so two handler delegates racing to resolve the same tag get the same slot. `ComponentHandler` caches its delegate in a non-volatile field, so two delegates can briefly exist for one handler; without this, they would take two counters and emit a duplicate id.

The map path stays for everything else (text and instruction handlers, and any wrapping or foreign `FaceletContext`), and the seam on `FaceletContextImplBase` defaults to it.

### Result

Isolated `buildView`, min of alternating rounds:

| view | before | after | Δ |
|---|--:|--:|--:|
| flat markup, 1600 components | 1338 µs | 1206 µs | **−9.2%** |
| flat markup, 400 components | 321 µs | 297 µs | **−7%** |

Views built through `c:forEach`, `ui:repeat`, `h:dataTable` or composites are unaffected, their repeating tag ids having kept the map small already; the 32-scenario suite therefore moves within noise.

### Verification

- Full TCK green.
- 1103 impl unit tests green.
- **Generated ids are unchanged**: rendered output byte-identical to the baseline across all 32 `test/perf` scenario pages, **9131 generated ids**, covering `c:forEach`, `ui:repeat`, `h:dataTable`, composites, nested composites, `f:viewParam` and the dynamic add/remove views.

### Harness

The first commit makes `BuildViewBenchIT` accept a comma-separated `-Dperf.scenarios` list (it silently fell back to `composite-build` when handed one, so a size sweep cost one server start per view) and adds the views to sweep with under `webapp/buildview/`. They are measuring instruments for `BuildViewBenchIT`, not lifecycle scenarios — `PerfBenchIT` does not list them, so suite totals stay comparable across rounds.

### What this does not fix

`table-build`'s `RESTORE_VIEW` is still ~2× MyFaces, and that residue is structural rather than a lever: measured, the whole remaining per-component bookkeeping in `ComponentTagHandlerDelegateImpl.apply` is worth ~3% there. Rejected along the way, each by measurement, so they don't get retried: an open-addressed replacement for the id map (no gain on top of this, and +3…+5% on `foreach`/`composite`), array-backing `FacetsMap` (already 4-bucket and null-safe), and dropping the facets-map iteration snapshot (−4.8% but it breaks resource relocation with a `ConcurrentModificationException`).


BalusC:

PR #5890, opened 2026-07-30
Accept all java.time type values when pattern and type co-exist

BalusC:
## Undraft only when https://github.com/jakartaee/faces/pull/2221 is merged!

---

Companion to jakartaee/faces#2211 / jakartaee/faces#2221, which add `instant`, `year`, `yearMonth` and `monthDay` to the `type` attribute of `DateTimeConverter`.

`ConvertDateTimeHandler.setAttributes` only forwards `type` to the converter alongside a `pattern` when `isJavaTimeType(type)` holds, and that method tested the first character:

```java
char c = type.charAt(0);
result = c == 'l' || c == 'o' || c == 'z';
```

That covers `local*`, `offset*` and `zoned*`, but silently drops the four new type values. The converter then keeps its default `type="date"` and takes the legacy `SimpleDateFormat` branch, which cannot handle a `java.time` value:

- `<f:convertDateTime type="monthDay" pattern="dd-MM" />` parses `30-07` into a `java.util.Date` of 1970-07-30, then fails with `IllegalArgumentException: Cannot convert 7/29/70, 8:00 PM of type class java.util.Date to class java.time.MonthDay` during Process Validations.
- `<f:convertDateTime type="instant" pattern="uuuu-MM-dd HH:mm:ss" />` fails with `IllegalArgumentException: Cannot format given Object as a Date` during Render Response.

Replaced the character check with an explicit set of the values listed in `DateTimeConverter#setType`, and inlined the now-trivial `isJavaTimeType` into its single call site.

Covered by `Spec2211IT` in jakartaee/faces#2221 (`tck/faces50/converters`), which fails on `master` and passes with this change. Verified locally: `faces50/converters` 31/31, `faces23/datetime-converter` 11/11, `faces20/api/converter-validator` green.

### DIFF
--- a/impl/src/main/java/org/glassfish/mojarra/facelets/tag/faces/core/ConvertDateTimeHandler.java
+++ b/impl/src/main/java/org/glassfish/mojarra/facelets/tag/faces/core/ConvertDateTimeHandler.java
@@ -16,6 +16,7 @@
 
 package org.glassfish.mojarra.facelets.tag.faces.core;
 
+import java.util.Set;
 import java.util.TimeZone;
 
 import jakarta.el.ELException;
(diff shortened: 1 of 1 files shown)


BalusC:
Accepted by MyFaces in https://github.com/jakartaee/faces/issues/2211#issuecomment-5258541117

PR #5891, opened 2026-07-31
Util.coalesce with 2 argument allocation-free (hot-path)

pizzi80:
2 args version that avoid varargs allocation and for loop
This is the most frequently used in Mojarra


pizzi80:
https://github.com/eclipse-ee4j/mojarra/issues/5856

BalusC:
Approved+merged but note that this is not really view build time related.

PR #5892, opened 2026-07-31
Cache the first id each Facelets tag generates

BalusC:
Every tag a view build applies gets a Facelets id, `prefix + "_" + tagId`, with both parts fixed for the life of the view — so every build of that view computes the same string, and Mojarra rebuilt it every time, once per component tag. Holding it on the (application-scoped) `DefaultFacelet`, indexed by the counter slot each tag already reserves since #5889 and keyed by prefix, hands it back instead.

Note this is the *Facelets* id, which is not the component id. It is what `findChildByTagId` matches a tag to its component by on a refresh (the `MARK_CREATED` marker), and `ComponentTagHandlerDelegateImpl.apply` generates one for every component tag regardless of whether the tag carries an `id` attribute. Where the tag has no `id`, the Facelets id is not the component id either: `IdMapper` aliases it to a short sequential one (`t1`, `t2`, …), and that is what seeds the component id, giving the familiar `j_idt5`. The Facelets id itself reaches the markup only under `org.glassfish.mojarra.useFaceletsID=true`, which is off by default. Where the author wrote `id="..."`, it plays no part in the component id at all. So this helps either way — `form-inputs` below carries 58 explicit ids and still gains.

Apache MyFaces has done the equivalent for years: `SectionUniqueIdCounter` serves ids out of a pre-generated `String[]` sized by `org.apache.myfaces.COMPONENT_UNIQUE_IDS_CACHE_SIZE` (default 300), falling back to generation past the end.

### What is and is not cached

Only a tag's **first** id within a build. A tag that generates more than one is being iterated over, where the number of ids follows the data rather than the view, so caching those would grow without limit. A cap of 64 prefixes bounds the other axis, since a Facelet included from inside an iteration produces one prefix per iteration; past the cap nothing is cached and ids are built as before.

Entries need no locking. An id being a pure function of the prefix and the tag makes every race benign: two builds racing to fill the same slot compute the same string, and a build that grows an array while another holds the pre-growth one loses cached entries but never returns a wrong id, because the loser simply rebuilds the string it could not find.

### Measured

Isolated Facelets build (`BuildViewBenchServlet`, `-Dperf.split=true` so the build is timed apart from `FacesContext` creation and `createView`), Tomcat 11, JDK 21, quiet machine.

| view (buildView step) | before | after | MyFaces |
| --- | --- | --- | --- |
| 1600 components, flat | 1061.0 µs | **987.1 µs (−7.0%)** | 871.1 µs |
| 400 components, flat | 233.6 µs | 226.6 µs | 188.4 µs |
| `form-inputs` | — | **163.7 µs** | 170.0 µs |

The MyFaces arm — untouched code — drifted +4.8% on the 400-component view between those two runs, so drift-corrected both flat views land near −7%. `form-inputs` now builds faster than MyFaces. The gap on the 1600-component view narrows from +21% to +13%; what remains is the second String allocation, in `UIViewRoot.createUniqueId`, which both implementations pay.

For context on the ceiling: a probe that made id generation entirely free measured −14% / −17%, i.e. parity with MyFaces at 1600 components. This change takes the reachable part of that without altering a single generated id.

### Verified

- Full Faces TCK, green.
- Full Mojarra impl unit suite, including four new `FirstIdCacheTest` tests covering entry persistence, per-prefix isolation, growth preserving cached entries, and the prefix limit.
- A self-check build that recomputed the id on every cache hit and compared it against the cached one was driven over all 32 `test/perf` scenarios, full lifecycles including postbacks: **zero mismatches**. Generated ids are unchanged.

Ref eclipse-ee4j/mojarra#5856

### DIFF
--- a/impl/src/main/java/org/glassfish/mojarra/facelets/impl/DefaultFacelet.java
+++ b/impl/src/main/java/org/glassfish/mojarra/facelets/impl/DefaultFacelet.java
@@ -83,6 +83,9 @@ final class DefaultFacelet extends Facelet implements XMLFrontMatterSaver {
     /** Dense per-tag counter slots, keyed by tag id. See {@link #getIdSlot(String)}. */
     private final Map<String, Integer> idSlots = new ConcurrentHashMap<>();
 
+    /** The first id each of this Facelet's tags generates, per id prefix. See {@link #firstIds(String)}. */
+    private final FirstIdCache firstIds = new FirstIdCache(this::getIdSlotCount);
+
     private Doctype savedDoctype;
 
     private String savedXMLDecl;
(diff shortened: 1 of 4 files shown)


BalusC:

PR #5905, opened 2026-08-04
Fix #5904: tolerate an empty fragment or query string

### LINKED ISSUE #5904: StringIndexOutOfBoundsException in UrlBuilder
We use Primefaces to render p:links, but this problem happens with the current UrlBuilder implementation.
It occurs with jakarta.faces-4.1.11 but same code is in the master as well.

``` 
UrlBuilder builder = new UrlBuilder("https://www.someurl.com/#", "UTF-8");
```

Leads to the following exception:
``` 
java.lang.StringIndexOutOfBoundsException: Index 0 out of bounds for length 0
	at java.base/jdk.internal.util.Preconditions$1.apply(Preconditions.java:55)
	at java.base/jdk.internal.util.Preconditions$1.apply(Preconditions.java:52)
	at java.base/jdk.internal.util.Preconditions$4.apply(Preconditions.java:213)
	at java.base/jdk.internal.util.Preconditions$4.apply(Preconditions.java:210)
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:98)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
	at java.base/java.lang.String.checkIndex(String.java:4822)
	at java.base/java.lang.StringLatin1.charAt(StringLatin1.java:46)
	at java.base/java.lang.String.charAt(String.java:1544)
	at com.sun.faces.context.UrlBuilder.cleanFragment(UrlBuilder.java:303)
	at com.sun.faces.context.UrlBuilder.extractSegments(UrlBuilder.java:249)
	at com.sun.faces.context.UrlBuilder.<init>(UrlBuilder.java:73)
	at com.sun.faces.context.ExternalContextImpl.encodeRedirectURL(ExternalContextImpl.java:1017)
``` 

Previous implementations (eg. 2.3.x) had a startsWith check which did not fail if the string is empty. Looks like the current impl is missing an `!f.isEmpty()` check before charAt(0). 

**Expected behavior**

No exception should occur.

**Additional context**

If you want I can also provide a pull request, since the fix seems trivial. 
But would first like to have another opinion.

If I should provide a PR I would also need a hint if/where i can find existing UrlBuilderTests (could not find some yet -> so  just start a new test class in master/impl/src/test/java/org/glassfish/mojarra/context and what should the test contain to start with?)

BalusC:
Closes #5904

`UrlBuilder.cleanFragment()` and `cleanQueryString()` strip the leading separator with `charAt(0)`, which throws `StringIndexOutOfBoundsException` when the segment is empty. A seed URL ending in a bare `#` or `?` produces exactly that, because `extractSegments` has already consumed the separator via `substring(index + 1)`. So `encodeRedirectURL("https://www.someurl.com/#")` — and `encodeBookmarkableURL` / `encodePartialActionURL` alike — crashes.

I fixed both methods, not only the reported one: `cleanQueryString` has the identical defect and crashes on a trailing `?`.

The fix restores the `startsWith()` form that 4.0 still has, rather than adding an `!isEmpty()` guard in front of the `charAt` — it removes the failure mode by construction, and the two branches now read the same.

## Scope

Regressed in #5251 (commit d0376338a32acee83f697669868914695a7b72f7, "+ fixed imports + little optimizations... still a bit ugly"), which swapped `startsWith(SEPARATOR)` for `charAt(0) == SEPARATOR_CHAR` in both methods and thereby dropped the empty-string tolerance that `startsWith` gave for free. That never reached 4.0, so only 4.1 and master are affected; master gets this via the usual upmerge.

## Tests

New `UrlBuilderTest`, 31 cases covering seed URL parsing, segment normalization and parameter assembly. With the fix reverted, 9 of them fail with the exact stack trace from the issue, at `cleanFragment` **and** `cleanQueryString`.

Beyond the regression itself it pins the surrounding contract, notably that exactly one leading separator is stripped from an extracted segment (`###` → `##`, `??a=1` → `?a=1`) and that a `?` inside a fragment stays part of the fragment rather than becoming a query string.

## Note, not addressed here

The strip-leading-separator leniency exists for the setters — `setFragment("#top")` and `setFragment("top")` should both mean `top`. `extractSegments` calls the same cleaners even though it has already removed the separator, so a genuine leading `?` in a query string is silently eaten: `page??a=1` builds as `page?a=1`, changing the parameter name from `?a` to `a`. Per RFC 3986 a `?` is legal in a query, so that one is arguably wrong; the fragment side is stripping something already malformed. Behaviour is shared with 4.0 and unrelated to this crash, so I left it alone and only pinned it in the tests. Happy to file it separately if it's worth changing.

### DIFF
--- a/impl/src/main/java/com/sun/faces/context/UrlBuilder.java
+++ b/impl/src/main/java/com/sun/faces/context/UrlBuilder.java
@@ -297,26 +297,20 @@ class UrlBuilder {
 
     private void cleanFragment() {
         if (fragment != null) {
-            // trim
             String f = fragment.trim();
-            // remove '#'
-            if ( f.charAt(0) == FRAGMENT_SEPARATOR_CHAR ) {
+            if (f.startsWith(FRAGMENT_SEPARATOR)) {
                 f = f.substring(1);
             }
-            // null if empty
             fragment = f.isEmpty() ? null : f;
         }
     }
 
     private void cleanQueryString() {
         if (queryString != null) {
-            // trim
             String q = queryString.trim();
-            // remove '?'
-            if ( q.charAt(0) == QUERY_STRING_SEPARATOR_CHAR ) {
+            if (q.startsWith(QUERY_STRING_SEPARATOR)) {
                 q = q.substring(1);
             }
-            // null if empty
             queryString = q.isEmpty() ? null : q;
         }
     }
--- a/impl/src/test/java/com/sun/faces/context/UrlBuilderTest.java
+++ b/impl/src/test/java/com/sun/faces/context/UrlBuilderTest.java
@@ -1,0 +1,316 @@
+/*
+ * Copyright (c) Contributors to Eclipse Foundation.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package com.sun.faces.context;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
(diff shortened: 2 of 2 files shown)


BalusC:
> Per RFC 3986 a `?` is legal in a query, so that one is arguably wrong; the fragment side is stripping something already malformed. Behaviour is shared with 4.0 and unrelated to this crash, so I left it alone and only pinned it in the tests. Happy to file it separately if it's worth changing.

Interesting. Leave it for now.

Issue #5903, opened 2026-08-03
Make ProjectStage gating consistent, and retire or rename the proprietary context parameters


BalusC:
Mojarra recognizes a group of proprietary context parameters which only make sense while developing, a handful whose defaults are wrong for production, and a few whose names do not describe what they do. They all have to be set, unset and reasoned about individually today, and nothing stops any of them from being switched on where they do harm.

[Documenting every context parameter](https://github.com/eclipse-ee4j/mojarra/blob/master/CONTEXT-PARAMS.md) for #5897 made the size of that group visible, and 5.0 is the moment to shrink it: the prefix rename from `com.sun.faces.*` to `org.glassfish.mojarra.*` already forces every application to revisit these entries in `web.xml`, so changing the semantics of a few rides along at no extra migration cost.

## The rule

**Per-request behavior gates on `Development`. Startup-time validation gates on not `Production`.**

That split is not invented here, it is what the implementation already does in about fifty-five places. `UnitTest` and `SystemTest` are deployed, reachable and frequently performance measured environments, so at request time they must behave like production: no stack traces in the response, no caching turned off, no security guard dropped. At boot time the opposite holds, since catching a configuration mistake before it ships is exactly what those stages exist for.

One deliberate exception. The Development mode configuration watcher starts at boot but is not validation: it polls every two seconds and tears the application down and back up when a file changed, which would restart the application underneath a running test. It stays on `Development`.

## 1. Stage checks which do not follow the rule

Six sites currently gate on `Production`. Four of them are per-request and get stricter, so that diagnostics which today still reach `UnitTest` and `SystemTest` stop reaching them.

| Site | Today | Becomes |
| --- | --- | --- |
| `OutcomeTargetLinkRenderer:110` | writes "no matching navigation case" **into the page** unless `Production` | only in `Development` |
| `AjaxExceptionHandlerImpl:165` | puts the raw exception message **into the ajax response** unless `Production` | only in `Development`, otherwise "See your server log for more information" |
| `ResourceHandlerImpl:474,509` | logs a missing resource at `WARNING` unless `Production` | `WARNING` in `Development`, otherwise the existing `t != null ? WARNING : FINE` |
| `Util:1219` | `disableIdUniquenessCheck=auto` skips the whole-tree duplicate id walk in `Production` | skips unless `Development`, since the walk is a per-request cost |

[...]

PR #5906, opened 2026-08-04
Fix #5902: delete the startup thread pool from ConfigManager

### LINKED ISSUE #5902: Threaded startup uses unmanaged threads because it looks up the wrong JNDI name
`ConfigManager` is the only place in the implementation which runs work on an executor. When `org.glassfish.mojarra.enableThreading` is enabled it parses the `faces-config.xml` and facelet taglib documents in parallel, and it obtains the executor like this:

```java
private static ExecutorService createExecutorService() {
    int tc = Runtime.getRuntime().availableProcessors();
    if (tc > NUMBER_OF_TASK_THREADS) {
        tc = NUMBER_OF_TASK_THREADS;
    }

    try {
        return (ExecutorService) new InitialContext().lookup("java:comp/env/concurrent/ThreadPool");
    } catch (NamingException e) {
        // Ignore
    }

    return Executors.newFixedThreadPool(tc);
}
```

`java:comp/env/concurrent/ThreadPool` is a resource reference in the **application's own** naming context. It resolves only when the application itself declared a matching `<resource-env-ref>` in `web.xml`. Practically no application does, so in the real world this lookup always fails and the method always returns `Executors.newFixedThreadPool(...)`.

The consequence is that on a Jakarta EE server, where a managed pool is available and is the only supported way to run application work on another thread, Mojarra starts raw unmanaged threads anyway. The container cannot manage, monitor or shut them down.

That context has to be recreated by hand today. Every task run on a foreign thread sets the parent's context class loader, resolves an `InitFacesContext`, propagates the web context, and unwinds all three in a `finally`. A managed executor propagates that itself, which is the point of using one.

## Proposal

Try the standard name first, then the current one, then fall back:

1. `java:comp/DefaultManagedExecutorService` — the Jakarta Concurrency default. Present without any application configuration on Payara, GlassFish, WildFly, TomEE and Liberty, and gives container-managed threads with context propagation.
2. `java:comp/env/concurrent/ThreadPool` — kept so that anyone who did declare the reference keeps their behavior.
3. `Executors.newFixedThreadPool(...)` — bare Jakarta Servlet containers such as Tomcat and Jetty, where no managed pool exists by definition. This stays the documented fallback rather than the accidental default it is today.

An executor obtained from JNDI must not be shut down by Mojarra, it belongs to the container; only the locally created one may be. The current code shuts down whatever it got, which is correct for the fallback and wrong for a managed pool, so that needs to become conditional.

## The fallback pool should use the thread factory which already exists

`Executors.newFixedThreadPool(tc)` uses `Executors.defaultThreadFactory()`, which produces non daemon threads named `pool-N-thread-M`. Mojarra already ships `MojarraThreadFactory`, which sets `setDaemon(true)` and names threads `Mojarra-<name>-<pool>-thread-<n>`, but it is used in exactly one place, the `WebResourceMonitor` pool in `ConfigureListener`.

The fallback should use it too:

- Non daemon threads are latent rather than actively harmful, since `executor.shutdown()` in the `finally` normally retires them. They bite when a task wedges: `shutdown()` is graceful and does not interrupt running work, so a hung parse leaves live non daemon threads and the JVM will not exit.
- `pool-3-thread-2` in a thread dump identifies nothing. `Mojarra-ConfigParser-1-thread-2` identifies the owner, which is exactly what is wanted while diagnosing a startup which did not finish.

## Why it matters beyond tidiness

- It is the difference between "supported" and "works but is not allowed" on a Jakarta EE server.
- It removes the main argument against ever defaulting `org.glassfish.mojarra.enableThreading` to enabled, since the risky path becomes the fallback rather than the normal case. That matters for a separate proposal to enable `org.glassfish.mojarra.validateXml` outside `Production`, whose startup cost this parallelism would offset.

## Three things which came out while reading this

These are separate from the executor lookup, but they sit in the same method's blast radius and decide how far this can be taken.

**Parallelism is already unconditional elsewhere.** `ConfigManager` line 347 runs `configProcessors.stream().parallel().forEach(e -> e.initializeClassMetadataMap(servletContext, facesContext))`. That uses the common `ForkJoinPool`, is not gated by `org.glassfish.mojarra.enableThreading`, and propagates no context. So the parameter does not mean startup is single threaded, it means those two document parsing calls are.

**The thread context machinery the parameter guards is dead.** Both `configProcessors` loops are plain sequential `.stream().forEach(...)`, so `currentThreadId != parentThreadId` is never true and the block which sets the parent context class loader, resolves an `InitFacesContext` and propagates the web context never runs. It is residue from when those loops were parallel. Either delete it, or make the loops parallel again deliberately.

**Config processor failures are swallowed.** Both loops carry:

```java
} catch (Exception e2) {
    // TODO Auto-generated catch block
    e2.printStackTrace();
}
```

A processor which throws prints to stderr and startup continues with incomplete configuration. This is worth fixing on its own merits, and it should be fixed before anything changes how this phase executes, since otherwise the change is evaluated against error reporting which does not work.

## Can the parameter go away entirely?

Plausibly yes, in this order:

1. fix `createExecutorService`, the managed lookup, the daemon thread factory and the conditional shutdown
2. fix the swallowed exceptions
3. delete the dead propagation block, or make the loops parallel on purpose
4. drop `org.glassfish.mojarra.enableThreading` and parse in parallel always

By step four the parameter protects nothing. Document ordering is already safe, `getXMLDocuments(...)` results pass through `sortDocuments(...)` afterwards, so parse order does not affect `absolute-ordering` semantics.

## Not proposed: ForkJoinPool

`ForkJoinPool.commonPool()` is a worse fit than what is there now. Its worker threads do not carry the application's thread context class loader, which the tasks depend on and currently set by hand, and the container can neither manage nor monitor them. A dedicated `new ForkJoinPool(n)` would be just as unmanaged as the fixed pool while buying nothing: parsing N documents is flat fan out with no recursive splitting, so there is no work stealing advantage to collect.

## Note on the scope of the win

The compiled schemas are already cached per application in `DbfFactory.getSchema`, through a `synchronizedMap` on the `ServletContext`, so schema compilation is serialised regardless. What parallelises is the parse and, when `validateXml` is enabled, the per document validation pass, which is the part that scales with how many `faces-config.xml` files the classpath carries. Note this is a smaller win than it was on 4.x, where the same executor also ran the annotation scan; 5.0 delegates that to CDI bean discovery and the `AnnotationScanner` SPI is deprecated.

BalusC:
Closes #5902.

The issue originally proposed fixing `ConfigManager`'s executor: look up `java:comp/DefaultManagedExecutorService` first, keep `java:comp/env/concurrent/ThreadPool` as a fallback, give the bare pool a daemon thread factory, and shut down only what Mojarra created. Measuring first turned that around. The executor is deleted instead.

## What this does not change

The pool was never on by default. `org.glassfish.mojarra.enableThreading` defaults to `false`, and `ConfigManager` created an executor only when an application had explicitly set it to `true`; every other deployment already ran the sequential branch which `Documents.getXMLDocuments(...)` has always carried for `executor == null`. So for anything which did not explicitly opt in, and that is practically everything, this deletes the branch which was never taken and the code path stays exactly what it was.

For anyone who did opt in, the measurement below says they were paying for it rather than gaining from it.

## Why

The heaviest realistic document set, PrimeFaces 16 (`faces-config.xml` plus `primefaces.taglib.xml` and `primefaces.urn.taglib.xml`, ~950 KB each) and OmniFaces 5.5 (`faces-config.xml` plus four taglibs), eight documents and 2.52 MB, parsed to DOM with Mojarra's settings, sequentially versus on `Executors.newFixedThreadPool(5)`, twenty cores, JDK 21:

| JIT state | sequential | 5 thread pool |
| --- | --- | --- |
| cold, first parse in the JVM | ~103 ms | ~120 ms |
| one small document pre parsed | ~70 ms | ~82 ms |
| fully warm, after three rounds | ~20 ms | ~8 ms |

The pool wins by 2.5x, but only once Xerces is fully JIT compiled, and configuration parsing runs exactly once per application, so startup never reaches that state. At the state which does hold at deployment time the pool is 15 to 20 percent slower, consistently. Pre starting all five threads off the clock gives the same result, so it is not thread startup cost but classload and JIT contention: five threads racing to interpret the same uncompiled parser code while the compiler threads compete for the same CPUs.

A `ManagedExecutorService` therefore cannot help. It changes who owns the thread, not the JIT state, and adds a JNDI lookup and context propagation on top.

## What changed

`ConfigManager`

- `createExecutorService()`, `NUMBER_OF_TASK_THREADS`, `useThreads(...)` and `getThreadContext(...)` deleted, along with the `java:comp/env/concurrent/ThreadPool` lookup. `useThreads(...)` was the `org.glassfish.mojarra.enableThreading` gate, so nothing in this file consults that parameter any more. The lookup behind it resolved only for an application which declared the matching `resource-env-ref` in its own `web.xml`, so on the rare occasion it was reached at all it still fell through to a raw unmanaged pool.
- The two configuration processor loops, `subList(0, 3)` and `subList(3, size)`, merge into one loop over the full list. The split existed only because the second half used to be dispatched to the pool; sequentially it is the same order.
- Both `e2.printStackTrace()` blocks are gone. A configuration processor which throws now propagates into the existing handler which releases the factories and rethrows `ConfigurationException`, instead of printing to stderr and letting startup continue with incomplete configuration.
- The thread context propagation block is deleted. Its `currentThreadId != parentThreadId` guard has been unreachable ever since both loops became sequential streams.
- `initializeConfigProcessors` drops `stream().parallel()`. This one is the exception to the paragraph at the top: it was not gated by `org.glassfish.mojarra.enableThreading` and ran on the common `ForkJoinPool` on every startup regardless, for thirteen calls which each do one `getAttribute`, one `new ConcurrentHashMap` and one `setAttribute`.

`Documents.getXMLDocuments(...)`

- Loses the `ExecutorService` parameter and all the `FutureTask` scaffolding. The `List<FutureTask>` plus `HashSet processedUris` collapses into a single `LinkedHashSet<URI>`, which is the same order and the same first occurrence wins deduplication.
- A `ConfigurationException` passthrough keeps the message `ParseConfigResourceToDOMTask` already builds, rather than rewrapping it as the `ExecutionException` path did.

SPI

- `ThreadContext` is left without a caller and is deprecated for removal. `InjectionProviderFactory` warns about it the way it already warned about `AnnotationScanner`, now through one shared check rather than two copies.
- `AnnotationScanner`'s deprecation javadoc claimed it was retained for binary compatibility with GlassFish HK2. GlassFish deleted `GlassFishInjectionProvider`, the class which implemented it, in the same commit which ported its integration to `org.glassfish.mojarra.spi`, leaving only `GlassFishHighAvailabilityEnabler`. The javadoc now states the real reason it survives, which is the warning above.
- `ConfigurationResourceProvider.validateXml(URI, boolean)` is removed. It has never had a caller, neither here nor in the pre import history: `getXMLDocuments` passes the application level flag straight to the parse task and never consults the per provider hook.

## Tests

New `DocumentsTest`, four cases: documents come back in provider order, a URI offered by two providers is parsed once at its first position, a provider which throws fails the configuration, and an unreadable resource fails the configuration. The last two are the executable form of the `printStackTrace` fix and would fail against the old behavior.

`mvn clean install` in `impl`: 1148 tests, no failures.

## Not in scope

`org.glassfish.mojarra.enableThreading` itself stays, because it has a second and unrelated use as the opt out for the Development mode configuration watcher in `ConfigureListener.shouldInitConfigMonitoring()`. Tying that watcher to the project stage and dropping the parameter is #5903. `CONTEXT-PARAMS.md` is updated meanwhile to describe only what the parameter still does.


BalusC:

PR #5908, opened 2026-08-04
Update DefaultFaceletFactory to forbid absolute URLs 

jasondlee:
- Update DefaultFaceletFactory to forbid absolute URLs 
- Add test to verify fix

CVE CVE-2026-46581


BalusC (approved):
Nit: rceIT should start with capital

Claude hints that it should also cover jar:file:/ paths.

jasondlee:
> Nit: rceIT should start with capital
> 
> Claude hints that it should also cover jar:file:/ paths.

Noted. I'll file a follow up issue, if that's OK.

PR #5910, opened 2026-08-04
Document the faces submodule in DEVELOPERS.md

BalusC:
Since 5.0 the API lives in [jakartaee/faces](https://github.com/jakartaee/faces/tree/5.0), wired into this repository as the `faces` git submodule. Its sources are needed at build time — among others for the `faces.d.ts` type definitions which the TypeScript sources of `faces.js` import — so a clone made without `--recurse-submodules` leaves `faces/` empty and the build dies in rollup on four files at once:

```
[INFO] [!] (plugin typescript) RollupError: [plugin typescript] faces/ajax.ts (1:28): @rollup/plugin-typescript TS2307: Cannot find module '../../../../../faces/api/src/main/resources/META-INF/resources/jakarta.faces/faces' or its corresponding type declarations.
```

DEVELOPERS.md only said to checkout `master`, which is why this keeps biting people. The 5.0 build instructions now lead with `git config --global submodule.recurse true`, which keeps every later `git pull`, `git checkout` and `git switch` in sync, and then clone with `--recurse-submodules`. Both are needed: per git-config, `submodule.recurse` applies to checkout, fetch, grep, pull, push, read-tree, reset, restore and switch, but explicitly not to `clone`. For an existing checkout there is `git submodule update --init`. The Eclipse instructions point at both, since File -> Import -> Git does not fetch submodules either.

CI is blind to this: `build.yml` checks `jakartaee/faces` out into `path: faces` with a second `actions/checkout` rather than using submodules, so it never sees an empty `faces/`.

Docs only, no code changes.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -9,17 +9,33 @@ In case you want to checkout this repository and manually build from source your
 ### Mojarra 5.0
 
 1. Make sure that you have JDK 17 and Maven installed.
-2. Checkout branch [`master`][28].
-3. Run the following commands from the `impl` directory of the project:
+2. Make sure that git is configured to keep submodules in sync:
+
+    ```bash
+    git config --global submodule.recurse true
+    ```
+
+    This covers every future `git pull`, `git checkout` and `git switch`, but not `git clone`, so the initial checkout in the next step needs its own flag.
+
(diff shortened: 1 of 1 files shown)


BalusC:
cc: @jasondlee

Issue #5912, opened 2026-08-04
Replace the provisional EE 12 and Faces 5.0 schemas once published


BalusC:
Mojarra ships its own copies of the XML schemas it validates configuration documents against, under `impl/src/main/resources/org/glassfish/mojarra`. Three of them do not come from https://jakarta.ee/xml/ns/jakartaee/ because they could not: the corresponding release was not out yet when they were needed.

## What is provisional right now

`jakartaee_12.xsd`, `web-facesconfig_5_0.xsd` and `web-facelettaglibrary_5_0.xsd` were added in #5903. All three return `404` from `https://jakarta.ee/xml/ns/jakartaee/`, so the two Faces schemas were taken from the API and `jakartaee_12.xsd` was derived from `jakartaee_11.xsd`, carrying the only delta EE 12 has over EE 11, which is dropping the web services client include and its group reference.

That is the same route `jakartaee_11.xsd` originally took, being copied and modified from `jakartaee_10.xsd` before EE 11 was released. It was then never revisited, and by the time #5903 looked at it, it was missing the eighteen Jakarta Concurrency definitions the published schema had gained: `context-service`, `managed-executor`, `managed-scheduled-executor`, `managed-thread-factory` and their types. 576 lines short of what it claimed to be, for however many releases. It has been replaced by the published file, and `jakartaee_12.xsd` re-derived from that.

Nothing referred to the missing types from Faces, so no behavior changed, which is exactly why nobody noticed.

## What to do

Replace all three with the published files once EE 12 and Faces 5.0 appear under https://jakarta.ee/xml/ns/jakartaee/, and diff rather than assume, since a provisional copy which is close enough to work is precisely the kind that survives unexamined.

## Worth considering

The same check is worth running over the whole directory rather than only these three, and worth automating rather than repeating by hand. `FacesSchemaTest` currently proves every schema a supported classpath may declare resolves and compiles, which is what caught `jakartaee_12.xsd` missing during #5903, but it cannot know whether the file is the right one. Verifying the shipped copies against the published ones needs the network and therefore does not belong in the unit test run, but it would fit a release check.

For the record, the state of the rest as of #5903: `web-facesconfig_4_1.xsd` and `web-facelettaglibrary_4_1.xsd` are byte identical to the published ones, and the `4_0` pair differs only in trailing whitespace and comment indentation, with no declaration, type or element differing.

🤖 Generated with Claude Opus 5

Issue #5911, opened 2026-08-04
Consolidate every Mojarra context parameter into one typed registry


BalusC:
Mojarra recognizes its own context parameters through several unrelated mechanisms, and which one a parameter happens to use decides whether it supports the legacy `com.sun.faces` prefix, whether it can be deprecated, whether it carries a type, and whether the documentation test can check it. That is not a property any parameter should have.

## Where they live today

`WebConfiguration.WebContextInitParameter` and `WebConfiguration.BooleanWebContextInitParameter` hold most of them, 25 and 21 entries. They support the legacy prefix, values supplied by `faces-config.xml`, JNDI environment entries, a cached list accessor, and, since #5903, deprecation with a replacement. They carry no type: every entry is a name and a `String` default, and the boolean enum exists only because the type could not be expressed otherwise.

[`FacesContextParam`](https://github.com/eclipse-ee4j/mojarra/blob/21c3e00711727a8c312be48a8df7b426fecf6775/impl/src/main/java/org/glassfish/mojarra/context/FacesContextParam.java) holds the 32 specification parameters and is the better design of the two. It carries a real type, default value suppliers, which is how `FACELETS_REFRESH_PERIOD` derives its default from `ProjectStage`, and `StringArray` separators for the list valued ones. It resolves on every call and caches nothing.

Legacy prefix support for the three parameters which used to be Mojarra's is a third mechanism again, sitting in neither enum: `WebConfiguration.PROMOTED_PARAMETERS` plus the static `WebConfiguration.getValue(FacesContextParam, FacesContext)`, which callers have to remember to use instead of `FacesContextParam.getValue(FacesContext)`.

Until #5903 there was also `MojarraContextParam`, two entries read through `ContextParamUtils`, which asked the `ServletContext` for the init parameter and nothing else. It has been folded into `WebConfiguration` and removed, because it was the weakest of the three and its two entries were silently missing legacy prefix support. That fold is a stopgap, not the destination.

## Why it matters

Three bugs of the same shape, two of them found while doing #5903.

`org.glassfish.mojarra.enableDistributable` was declared in both `MojarraContextParam` and `WebConfiguration` and read both ways, and the two disagreed. `ConfigureListener` enables the option when `web.xml` declares `<distributable/>`, which is invisible to a plain `getInitParameter`, so `ExternalContextImpl` used the non replicating session map while four other readers believed the application was distributable.

[...]

Issue #5648, opened 2026-01-09
5.0: implement jakarta.faces.EXCEPTION_TYPES_TO_IGNORE_IN_LOGGING


BalusC:
jakartaee/faces#2097

PR #5915, opened 2026-08-04
Make ProjectStage gating consistent, and retire or rename the proprietary context parameters

### LINKED ISSUE #5903: Make ProjectStage gating consistent, and retire or rename the proprietary context parameters

BalusC:
Closes #5903.

Mojarra's proprietary context parameters had accumulated three separate problems: behavior which should follow the project stage was instead spread over individual parameters, several parameters no longer earned their name, and a few no longer earned their existence. 5.0 is where that can be fixed, because the prefix rename from `com.sun.faces.*` to `org.glassfish.mojarra.*` already forces every application to revisit these entries in `web.xml`.

## The rule

**Per-request behavior gates on `Development`. Startup-time validation gates on not `Production`.**

That split is not invented here; it is what the implementation already did in about fifty-five places. `UnitTest` and `SystemTest` are deployed, reachable and frequently performance measured, so at request time they have to behave like production: no stack traces in the response, no caching turned off, no security guard dropped. At boot the opposite holds, since catching a configuration mistake before it ships is what those stages are for.

Four checks did not follow it and now do, all of them getting stricter rather than looser: `OutcomeTargetLinkRenderer` wrote "no matching navigation case" into the rendered page, `AjaxExceptionHandlerImpl` put the raw exception message into the ajax response, `ResourceHandlerImpl` logged a missing resource at `WARNING`, and `disableIdUniquenessCheck=auto` skipped its per-request tree walk in `Production` only.

One deliberate exception: the Development mode configuration watcher stays on `Development`, because it is not a check. It polls every two seconds and tears the application down and back up when a file changed, which would restart the application underneath a running test.

## Removed

| Parameter | Why |
| --- | --- |
| `validateXml` | derived from the stage instead |
| `verifyObjects` | derived from the stage instead |
| `enableThreading` | #5902 deleted the parsing half; the rest gated the Development watcher, which now runs on the stage alone |
| `enableJSStyleHiding` | wrapped inline script in `<!--` for browsers which do not understand `<script>`, no such browser exist anymore |
| `ENABLE_HTTP_METHOD_RESTRICTION_PHASE_LISTENER` | `OPTIONS` is answered by `FacesServletImpl` itself |

`OPTIONS` deserves a note. `FacesServletImpl` already filtered HTTP methods in `service()`, before the `FacesContext` is acquired, and now answers `OPTIONS` there with `200`, an `Allow` header derived from the configured method set, and no body, which is what RFC 9110 section 9.3.7 asks for. That collapses three behaviors into one: by default `OPTIONS` used to run the whole lifecycle and return the rendered view, which is wasted work and hands the page to a request which did not ask for it; the phase listener turned that into an empty `200` while running on every request to compare one string; and excluding `OPTIONS` from `allowedHttpMethods` returns `400`, which breaks CORS preflight and so was never equivalent.

## Renamed

`autoCompleteOffOnViewState` becomes `viewStateAutocomplete`, a string written verbatim as the `autocomplete` attribute of the `ViewState` field, defaulting to `one-time-code`. The old name described the old behavior: before #5577 it meant "write `autocomplete=off`", and that commit made the attribute unconditional and repurposed `true` to mean "give me `off` back". So the parameter is a deliberate escape hatch for a browser which mishandles `one-time-code`, and only its name went stale. A string means the next change of recommended value needs no new parameter.

`numberOfLogicalViews` becomes `numberOfStatefulPagesPerSession` and `numberOfViewsInSession` becomes `numberOfViewStatesPerStatefulPage`. Two problems there. The nesting was inverted: the session holds a map sized by the first, each of whose entries is a map sized by the second, so the number of views retained is the product and neither name said so. The pre-rename aliases prove it, `NUMBER_OF_VIEWS_IN_SESSION` having named the inner map and `NUMBER_OF_VIEWS_IN_LOGICAL_VIEW_IN_SESSION` the outer one. And "logical view" was borrowed: the specification already uses it for `ViewHandler.deriveLogicalViewId`, which returns a view id string, whereas this one is a random long.

The new names describe what an application sizes: the pages which keep state, and how far back each of them may go. Two things they cannot carry, and the documentation therefore does. It counts page *visits* rather than distinct pages, so the same page open in two browser tabs occupies two entries. And only a non ajax postback adds a view state, because that is what adds a browser history entry; an ajax postback reuses the current one, so a page which polls or validates over ajax does not evict its own history.

Renaming rather than swapping the two back is deliberate, since swapping would silently change the meaning of every existing configuration with no diagnostic.

## Deprecated

`allowTextChildren` cannot be made coherent either way. `TextRenderer.getRendersChildren()` returns `true` unconditionally, so the lifecycle always routes the children of an `<h:outputText>` or `<h:inputText>` through `encodeChildren`, which then returns immediately unless the parameter is set. The default is therefore to accept the children and discard them silently. Setting it does not fix that: for a `UIOutput` the children bypass the `escape` attribute, and for a `UIInput` the `<input>` element has already been written and is void, so they land after it.

`disableClientStateEncryption` is folded into `enableClientStateDebugging`, which inherits its value. Both were added by the same change for the same reason, and dropping the `ByteArrayGuard` is half of what debugging that state means.

## Honored only in Development

`enableClientStateDebugging` drops the `ByteArrayGuard`, which provides both the confidentiality and the tamper detection of the client side state. `generateUniqueServerStateIds=false` makes the server side state identifier predictable. Both revert to their default outside `Development` and say so, because silently dropping a setting which weakens a deployment would be worse than honoring it.

Reverting to the default rather than forcing a value is deliberate: the two have opposite polarity, the risky setting being `true` for one and `false` for the other, and in both cases the default is the safe one.

## Stage derived defaults

All four default to `auto`, and this is what `auto` resolves to:

| Parameter | Was | `Development` | `UnitTest`, `SystemTest` | `Production` |
| --- | --- | --- | --- | --- |
| `cacheResourceModificationTimestamp` | `false` | `false` | `true` | `true` |
| `resourceUpdateCheckPeriod` | `5` | `5` | `-1` | `-1` |
| `disableIdUniquenessCheck` | `false` | `false` | `true` | `true` |
| `displayConfiguration` | `false` | `true` | `true` | `false` |

The middle column is the point of the rule. The first three decide per-request behavior and so treat `UnitTest` and `SystemTest` as production, while `displayConfiguration` reports at startup and so treats them as development.

All four join `disableUnicodeEscaping` in taking `auto`, which is the default and leaves the decision to the project stage, rather than defaulting to a value production does not want and then correcting for it somewhere else. The first two were corrected in two different somewhere elses: `WebConfiguration` rewrote them at startup when the stage was `Development`, and both `ClientResourceInfo` constructors additionally ANDed the first with not being in `Development`, so an explicit `true` was silently discarded while developing. It is honored now, which is what an explicit value in a tri-state means, and the stage rule lives in the parameter alone. `displayConfiguration` is kept rather than removed because a deployment whose parameters are substituted at build time can legitimately want it in `Production`.

An unusable value is reported and behaves as though the parameter was never set. `Boolean.parseBoolean` makes `false` of everything which is not `true`, so a typo, or a plausible spelling such as `yes` or `1`, silently meant `false`, and an unparseable period silently meant never checking, which is the answer for `Production` even when the stage is `Development`. All of them resolve through one helper now, which also trims: a container is not obliged to strip whitespace from a `<param-value>`, so ` true ` used to be `false`.

## validateXml never validated anything

This is the one which needs the closest look.

`loadDocument` guarded the validation on `builder.isValidating()`, which reports whether the parser does DTD validation and is unaffected by `setSchema`, so it was always false. Enabling `validateXml` resolved and compiled a schema and then returned the document untouched. Worse, resolving that schema throws for an unrecognized version, and that happens while evaluating an argument of the call which was about to skip validation anyway, so a library declaring a configuration version newer than the runtime took the deployment down without a single document ever being validated.

Validation now actually runs. Every violation is reported and validation continues, so one run lists all of them, and none of it fails the deployment: a configuration file which a previous release accepted has to keep deploying, and until now none of these were reported at all. An unrecognized version warns and leaves that document unvalidated.

It also names the mistake behind the most confusing failure this produces. A namespace is compared as an exact string and never resolved, so `xmlns:xsi="https://www.w3.org/2001/XMLSchema-instance"` declares an entirely different namespace than the XML Schema instance one, and the schema then rejects `xsi:schemaLocation` while naming the attribute rather than the declaration which actually went wrong. Every namespace declared on the document element is now compared against the ones a configuration document may legitimately declare, and one which matches after swapping `http` for `https` is reported first.

`FacesSchema` stopped at 4.1, so a document declaring version 5.0 raised "Unknown Schema version". The 5.0 schemas are added, taken from the API since neither is published under jakarta.ee yet. `jakartaee_11.xsd` was derived from `jakartaee_10.xsd` back when EE 11 was not released and never caught up, missing the eighteen Jakarta Concurrency definitions the published schema gained; it is replaced by the published file and `jakartaee_12.xsd` re-derived from that. Nothing in Faces referred to the missing types, which is why nobody noticed. Tracked for replacement once published in #5912.

## Reporting

The project stage is reported once at startup when it is not `Production`, since an application accidentally deployed with a development stage otherwise looked exactly like one which was not, and the stage now decides considerably more than it did.

Nothing validated the names an application declares, so a typo such as `enableTreading`, or a parameter which has been removed, did nothing and said nothing. Every unrecognized name is reported once at startup, outside `Production`. Only Mojarra's own namespace is checked, because it is the only one this release is authoritative about.

Three parameters were promoted to the specification and only two were still accepted under the name they had; `exceptionTypesToIgnoreInLogging` from #5648 was never added. All three now also accept the `org.glassfish.mojarra` spelling, a name none of them ever had, because they were promoted in the same release which renamed the prefix and that is what the documented migration produces.

## Found along the way

These were not in the issue.

`org.glassfish.mojarra.enableDistributable` was declared in two registries and read two ways which disagreed. `ConfigureListener` enables it when `web.xml` declares `<distributable/>`, which a plain `getInitParameter` cannot see, so `ExternalContextImpl` used the non replicating session map while four other readers believed the application was distributable.

`org.glassfish.mojarra.allowedHttpMethods` was a private constant in `FacesServletImpl` read straight off the `ServletContext`, so `com.sun.faces.allowedHttpMethods` carried over from 4.x was silently ignored.

`MojarraContextParam` and `ContextParamUtils` are removed, their two entries folded into `WebConfiguration`. That is a stopgap, not the destination; the consolidation is #5911.

`isDevModeEnabled` is renamed to `isDevelopment` throughout, matching the `DefaultFacelet` method which was already called that.

`ConfigureListener` resolved the stage by reading the context parameter directly and therefore never saw one configured through JNDI, so the configuration watcher stayed off for anybody configuring it that way.

## Testing

1212 unit tests pass. Ten new test classes, each verified to fail when the production change is reverted rather than passing vacuously.

The TCK fix rides along in the `faces` submodule bump (jakartaee/faces@0418e273d): `never-unwrap-exceptions` wraps `FacesServlet` in a servlet which catches the `ServletException` and prints what it caught, so it only means anything when `FacesServlet` throws, which it does not in `Development` with no error page declared. It is pinned to `Production` there, for the same reason `refresh-period-production` is.

## Not in scope

`org.glassfish.mojarra.disableUnicodeEscaping` and `org.glassfish.mojarra.preferXHTML` are explicitly **kept**, both of which look like they belong in the removals and do not. The first decides on the response encoding and skipping `&#nnnn;` escapes on a UTF-8 response makes the output smaller and faster. The second is a tie break inside the `Accept` negotiation rather than an override, so a client which does not offer `application/xhtml+xml` still gets `text/html`; it still fires today, and the negotiated content type is what `HtmlResponseWriter` derives `isXhtml` from, which selects `<![CDATA[` wrapping for inline script.


BalusC:

Issue #5909, opened 2026-08-04
Fix ui:include testing


jasondlee:
The module/directory name doesn't follow conventions, and the test itself should be more robust. This affects all version 4.0 and forward.

BalusC:
Closed by #5916 and #5917

PR #5917, opened 2026-08-04
Expand ui:include RCE integration test coverage

jasondlee:
Rename test module from rce to uiIncludeRce and rewrite UiIncludeRceIT
with broader attack-vector coverage:

- Positive cases: same-directory and subdirectory includes still work
- Remote schemes: https, http, file, jar, ftp, data protocols all blocked
- Path traversal: plain ../, URL-encoded (%2F, %2e), double-encoded
  (%252F), and traversal with .xhtml suffix
- Absolute paths: /WEB-INF/web.xml blocked; /WEB-INF/secret.xhtml
  (valid Facelets suffix) verifies container-level WEB-INF protection
- Edge cases: malformed URI, empty param, missing param, non-Facelets
  suffix (.properties)
- Content-leak assertions: blocked tests assert sensitive content is
  absent, not just that an error message appeared
- assertBlocked helper: matches all three guard error messages instead
  of hardcoding "Invalid path"

Signed-off-by: Jason Lee <jason@steeplesoft.com>
Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

Addresses: #5909


BalusC:

PR #5922, opened 2026-08-05
Sort out FACELETS_SUFFIX and FACELETS_VIEW_MAPPINGS semantics

### LINKED ISSUE #5918: FACELETS_SUFFIX honors only the first configured extension
`jakarta.faces.FACELETS_SUFFIX` takes a space separated list of extensions since 4.0, and the javadoc of `ViewHandler.FACELETS_SUFFIX_PARAM_NAME` specifies which one wins:

> This list is a space separated list of values of the form *`.<extension>`*. The first physical resource whose extension matches one of the configured extensions will be the suffix used to create the view ID.

Mojarra never consults the physical resource. `MultiViewHandler.convertViewId` returns from within the first iteration of its loop along both arms, so only the first configured extension is ever considered:

```java
for (String ext : configuredExtensions) {
    if (viewId.endsWith(ext)) {
        return viewId;
    }

    appendOrReplaceExtension(viewId, ext, length, extIdx, buffer);

    return buffer.toString();
}
```

The existence check which does exist runs after the fact and does not recover: `derivePhysicalViewId` calls `viewExists` only when its `checkPhysical` argument is set, and returns `null` on failure rather than trying the next extension.

## How to reproduce

```xml
<context-param>
    <param-name>jakarta.faces.FACELETS_SUFFIX</param-name>
    <param-value>.xhtml .html</param-value>
</context-param>
```

With the `FacesServlet` extension mapped to `*.xhtml`, and `page.html` present on disk while `page.xhtml` is not, requesting `/page.xhtml` yields a 404. The view id already ends with the first configured extension, so it is returned unchanged and the configured `.html` is never tried.

Reversing the value to `.html .xhtml` moves the problem rather than solving it: `/page.xhtml` then resolves to `page.html`, and a view which exists only as `.xhtml` becomes unreachable. Whichever extension is listed first is the only one which works, which makes the list pointless.

## Affected versions

4.0, 4.1 and master. `convertViewId` is identical on all three.

This is distinct from b141b1b114, which fixed master resolving the parameter as a single `String` instead of as a list. That restored the splitting of the value; this issue is about which of the split extensions gets used, and it predates that regression.

## Suggested fix

`convertViewId` has to try each configured extension and return the first whose resource actually exists, falling back to the first configured extension when none does, so that a genuine not-found still produces a usable view id for the error path.

That is a behavioral change rather than the type correction b141b1b114 was, and it needs a decision on the fallback, which is why it is filed separately.

The TCK does not cover this. `faces23/facelets-suffix` is the only module configuring the parameter, and it declares a single extension.

### LINKED ISSUE #5920: Declaring FACELETS_VIEW_MAPPINGS without a *.xhtml entry breaks deployment
Discovered while implementing the integration test for #5918.

Declaring `jakarta.faces.FACELETS_VIEW_MAPPINGS` with a value which does not contain a `*.xhtml` entry makes the application fail to deploy on GlassFish. The documented prefix form is enough:

```xml
<context-param>
    <param-name>jakarta.faces.FACELETS_VIEW_MAPPINGS</param-name>
    <param-value>/faces/*</param-value>
</context-param>
```

No other configuration is involved. `jakarta.faces.FACELETS_SUFFIX` need not be declared, and declaring it does not help. Any view will do, the failure happens at deployment before a request is served.

## Cause

`FaceletViewHandlingStrategy.handlesByPrefixOrSuffix` treats `FACELETS_VIEW_MAPPINGS` as a replacement for the Facelets suffix rather than as an addition to it:

```java
// If there's no extensions array or prefixes array, then assume defaults.
// .xhtml extension is handled by the FaceletViewHandler
if (extensionsArray == null && prefixesArray == null) {
    return isMatchedWithFaceletsSuffix(viewId) ? true : viewId.endsWith(DEFAULT_FACELETS_SUFFIX);
}
```

`initializeMappings` assigns both arrays whenever the parameter is non-empty, possibly zero length, so declaring it at all takes that whole branch out of play, including the `DEFAULT_FACELETS_SUFFIX` fallback on the same line.

That fallback is load bearing during startup. `ApplicationAssociate.PostConstructApplicationListener` resolves a synthetic view id to force the Facelets VDL to instantiate early:

```java
// cause the Facelet VDL to be instantiated eagerly, so it can
// become aware of the resource library contracts
viewHandler.getViewDeclarationLanguage(context, FACES_PREFIX + "xhtml");
```

`jakarta.faces.xhtml` matches no user supplied mapping, so `handlesByPrefixOrSuffix` returns `false` and `handlesViewId` falls through to `isViewIdExactMappedToFacesServlet`, which reaches `ServletContext.getServletRegistrations()`. That call is illegal from a `ServletContextListener` which was added programmatically, and deployment dies:

```
java.lang.UnsupportedOperationException: Not allowed to call this jakarta.servlet.ServletContext method from a ServletContextListener that was neither declared in the application's deployment descriptor nor annotated with WebListener
	at org.apache.catalina.core.ApplicationContext.getServletRegistrations(ApplicationContext.java:431)
	at com.sun.faces.util.Util.getExistingFacesServletRegistration(Util.java:196)
	at com.sun.faces.util.Util.getFacesServletMappings(Util.java:186)
	at com.sun.faces.util.Util.isResourceExactMappedToFacesServlet(Util.java:1092)
	at com.sun.faces.util.Util.isViewIdExactMappedToFacesServlet(Util.java:1076)
	at com.sun.faces.application.view.FaceletViewHandlingStrategy.handlesViewId(FaceletViewHandlingStrategy.java:842)
	at com.sun.faces.application.view.ViewHandlingStrategyManager.lambda$getStrategy$0(ViewHandlingStrategyManager.java:55)
	...
	at com.sun.faces.application.view.MultiViewHandler.getViewDeclarationLanguage(MultiViewHandler.java:406)
	at com.sun.faces.application.ApplicationAssociate$PostConstructApplicationListener.processEvent(ApplicationAssociate.java:313)
```

A value of `*.xhtml` deploys, because the synthetic view id then matches the user supplied mapping by accident. That is presumably why this has gone unnoticed: it is the one value which happens to be safe, and it is also the most commonly copied one.

## The two parameters are not alternatives

The javadoc of `FACELETS_VIEW_MAPPINGS_PARAM_NAME` describes a widening mechanism, not a replacement:

> Allow the web application to define a semicolon (;) separated list of strings that is used to forcibly declare that certain pages in the application must be interpreted as using Facelets, regardless of their extension.

"Forcibly declare that certain pages ... must be interpreted as using Facelets, **regardless of their extension**" adds pages, it does not withdraw the ones already recognized by their suffix. Section 7.6.2.1 of the specification, `deriveViewId`, likewise consults both in sequence rather than choosing between them: it forms `candidateViewId` from the `FACELETS_SUFFIX` entries first and only then walks `faceletsViewMappings`.

To be precise about what is and is not a specification violation: `handlesViewId` is not a specified method, so the deployment failure is a plain implementation bug rather than a conformance issue. The specification and javadoc are cited only for the design question of whether the two parameters are meant to be mutually exclusive, and they say they are not.

## Since when

Not a regression. `handlesByPrefixOrSuffix`, the `isViewIdExactMappedToFacesServlet` fall through in `handlesViewId`, and the startup probe in `ApplicationAssociate` are all present with this exact logic in 6f612e6797 (2018-04-16), the initial Mojarra contribution to Eclipse, so this predates the git history in this repository. 34bef08ac4 (2020-10-29) only moved the method, byte for byte.

## Affected versions

4.0, 4.1 and master. `handlesByPrefixOrSuffix` is identical on all three, as is the startup probe.

Confirmed on GlassFish 7.0.16. A container which does not enforce the `ServletContextListener` restriction on `getServletRegistrations()` will not fail to deploy, but still evaluates `handlesViewId` incorrectly, so the Facelets VDL declines view ids which are recognized by their suffix rather than by a declared mapping.

## Suggested fix

Consult the Facelets suffix regardless of whether `FACELETS_VIEW_MAPPINGS` is declared, rather than as an alternative to it. `getConfiguredSuffix` needs the same treatment, it carries the same guard.

Keeping the `DEFAULT_FACELETS_SUFFIX` fallback is deliberate: `handlesViewId` answers which view declaration language can process a resource, which is not the same question as which suffix a view id may carry, and Facelets can always process an `.xhtml` resource. Removing it breaks the startup probe for every webapp whose `FACELETS_SUFFIX` omits `.xhtml`.

### LINKED ISSUE #5921: FACELETS_SUFFIX and FULL_STATE_SAVING_VIEW_IDS mishandle redundant whitespace in the value
Discovered while implementing the integration test for #5918.

`WebConfiguration.getOptionValue(WebContextInitParameter, String)` splits a list valued context parameter and hands the result out as is:

```java
result = split(appMap, value, sep);
```

Neither the entries nor the value are normalized, so redundant whitespace within the `<param-value>` survives into the list. Two consequences.

## Empty entries

`Pattern.split` on a literal single space keeps interior empty strings, so a redundant space between two entries produces an entry which is the empty string:

```
" ".split(".xhtml  .html")   ->  [".xhtml", "", ".html"]
```

Every consumer of this list asks `endsWith`, and `endsWith("")` is true for everything, so one empty entry turns the list into a match-anything list. For `jakarta.faces.FACELETS_SUFFIX` that means `FaceletViewHandlingStrategy.handlesViewId` claims every view id, and `ViewHandler.getViews` reports every file in the webapp as a view, which `jakarta.faces.AUTOMATIC_EXTENSIONLESS_MAPPING` then maps to the `FacesServlet`.

Only interior whitespace is involved. Leading and trailing whitespace cannot reach here from a deployment descriptor, because the Servlet specification requires the container to strip it:

> Web containers must remove all leading and trailing whitespace, which is defined as `S(white space)` in XML 1.0, for the element content of the text nodes of a deployment descriptor.

That leaves `ServletContext.setInitParameter` as the only way to get a leading empty entry, which is a fringe case. `overrideContextInitParameter` already trims.

The hazard is understood elsewhere in the codebase, in the two consumers which split for themselves and filter explicitly rather than the shared method doing it once:

```java
/**
 * @return the excluded file extensions, without empty entries, as an empty entry would exclude every resource
 */
static String[] parseExcludedExtensions(Map<String, Object> appMap, String excludesParam) {
    return Stream.of(Util.split(appMap, excludesParam, " ")).filter(extension -> !extension.isEmpty()).toArray(String[]::new);
}
```

and in `AnnotationScanner`:

```java
for (String option : options) {
    if (option.length() == 0) {
        continue;
    }
```

`jakarta.faces.FACELETS_VIEW_MAPPINGS` is covered too, `initializeMappings` skips any entry of length 1 or less. `jakarta.faces.FACELETS_SUFFIX` is the one which has no guard.

## Untrimmed entries

An entry which keeps its surrounding whitespace no longer compares equal to the value it names. `jakarta.faces.FULL_STATE_SAVING_VIEW_IDS` is a comma separated list which `ApplicationStateInfo` puts in a `HashSet` and looks up by equality:

```java
String[] viewIds = config.getOptionValue(FullStateSavingViewIds, ",");
fullStateViewIds = new HashSet<>(viewIds.length, 1.0f);
fullStateViewIds.addAll(asList(viewIds));
...
return partialStateSaving && !fullStateViewIds.contains(viewId);
```

So a value written the way a comma separated list is normally written:

```xml
<param-name>jakarta.faces.FULL_STATE_SAVING_VIEW_IDS</param-name>
<param-value>/a.xhtml, /b.xhtml</param-value>
```

registers `/a.xhtml` and `" /b.xhtml"`, and `/b.xhtml` silently keeps using partial state saving. There is no warning. The container's stripping does not help here, it removes whitespace around the element content, not around the separators inside it.

## Since when

Not a regression, the unnormalized split is unchanged since 6f612e6797 (2018-04-16), the initial Mojarra contribution to Eclipse, and so predates the git history in this repository.

The two guards which do exist were added when their own consumer started needing one, rather than upstream: `AnnotationScanner` has had its check since that same initial contribution, and `parseExcludedExtensions` acquired its filter only in 213a3866c0 (2026-07-20, #5878), when `RESOURCE_EXCLUDES` moved from regex matching to `endsWith` and inherited the same hazard.

## Affected versions

Empty entries in `FACELETS_SUFFIX`: 4.0 and 4.1. Not master, where `FacesContextParam.StringArray.SPACE_SEPARATED` splits on `\\s+` (845c423f6c, 2025-03-01) and so collapses the interior run.

Untrimmed `FULL_STATE_SAVING_VIEW_IDS`: 4.0, 4.1 and master. Master's `COMMA_SEPARATED` is `\\s*,\\s*`, but `ApplicationStateInfo` does not use it, it still goes through `WebConfiguration.getOptionValue(FullStateSavingViewIds, ",")`.

## Suggested fix

Trim each entry and drop the empty ones in the shared split, so every list valued parameter gets it rather than the consumers which happen to do it themselves.

BalusC:
`jakarta.faces.FACELETS_SUFFIX` and `jakarta.faces.FACELETS_VIEW_MAPPINGS` are read in several places which each interpret them slightly differently. This sorts out what each of them answers, and fixes the four defects that came out of the confusion.

Closes #5918. Closes #5920. Closes #5921.

## The two questions

Two distinct sets were being conflated, and `WebConfiguration` now names them apart:

| method | returns | answers |
| --- | --- | --- |
| `getFaceletsSuffixes()` | exactly the configured `FACELETS_SUFFIX` values | which physical resource an incoming request may resolve to, so `deriveViewId` builds its candidate view ids from these |
| `getFaceletResourceSuffixes()` | those, plus the extension entries of `FACELETS_VIEW_MAPPINGS`, plus always the default suffix | whether a resource is a Facelet at all, which a template or an included fragment is without ever being requested |

`getConfiguredExtensions` was the old name of the first one, which said neither.

## The commits

**Distinguish Facelet resource suffixes from view suffixes.** A webapp declaring `FACELETS_SUFFIX` without `.xhtml` could no longer include an `.xhtml` fragment, and one declaring an extension through `FACELETS_VIEW_MAPPINGS` could not include a fragment carrying it, because the include admissibility check consulted the view suffixes.

**Fix #5920.** Declaring `FACELETS_VIEW_MAPPINGS` took the Facelets suffix branch of `handlesByPrefixOrSuffix` out of play entirely, so the two parameters behaved as alternatives rather than as the two widening mechanisms they are. On GlassFish the application then fails to deploy outright, because the runtime resolves a synthetic view id carrying the default suffix during startup to instantiate the Facelets VDL, and that resolution falls through to a `ServletContext.getServletRegistrations()` call which is illegal from a `ServletContextListener`. `/faces/*` alone is enough to trigger it; `*.xhtml` is the one value which happens to be safe, which is presumably why this has gone unnoticed since the initial contribution. That startup probe now asks the factory for all view declaration languages instead of going through a view id at all.

**Fix #5921.** Redundant whitespace in a list valued context parameter survived into the split entries. An empty entry matches every candidate a consumer tests it against with `endsWith`, and an untrimmed entry no longer compares equal to the value it names, which silently drops a `FULL_STATE_SAVING_VIEW_IDS` entry written as `/a.xhtml, /b.xhtml`.

**Fix #5918.** `convertViewId` returned from within the first iteration of its loop along both arms, so only the first configured suffix was ever considered. It now returns the first candidate whose physical resource exists and falls back to the first configured suffix, with a single configured suffix short-circuiting so the common case does not pay for the existence check. It also only treats a dot as an extension separator when it follows the last `/`: `/a.b/foo` previously resolved to `/a.xhtml`.

**Compare the archive rather than the authority of a nested URL.** A `jar:` url carries no authority, so comparing one against the facelet it was resolved from said nothing whenever that facelet is itself served from an archive, which is how a container serves a facelet packaged under `META-INF/resources`.

## Testing

Two integration test modules, `test/issue5918` and `test/issue5920`, between them covering the suffix and view mapping configurations which differ in behaviour; each of their tests was checked to fail without its fix. `test/uiIncludeRceJar` adds a facelet served from an archive so that the nested URL case is reachable at all, which it is on TomEE but not on GlassFish. Unit coverage went to the two existing test classes rather than to new ones: candidate generation in `MultiViewHandlerTest`, and the `handlesByPrefixOrSuffix` matrix in `FaceletViewHandlingStrategyTest`, including the literal startup probe id.

Full integration suite green on GlassFish, and `test/uiIncludeRce` additionally green on TomEE and WildFly.

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/ApplicationAssociate.java
+++ b/impl/src/main/java/com/sun/faces/application/ApplicationAssociate.java
@@ -17,7 +17,6 @@
 package com.sun.faces.application;
 
 import static com.sun.faces.RIConstants.FACES_CONFIG_VERSION;
-import static com.sun.faces.RIConstants.FACES_PREFIX;
 import static com.sun.faces.config.WebConfiguration.BooleanWebContextInitParameter.AutomaticExtensionlessMapping;
 import static com.sun.faces.config.WebConfiguration.BooleanWebContextInitParameter.FaceletsSkipComments;
 import static com.sun.faces.config.WebConfiguration.WebContextInitParameter.FaceletsDecorators;
@@ -31,6 +30,7 @@ import static com.sun.faces.util.Util.getFacesServletRegistration;
 import static com.sun.faces.util.Util.split;
 import static jakarta.faces.FactoryFinder.FACELET_CACHE_FACTORY;
 import static jakarta.faces.FactoryFinder.FLOW_HANDLER_FACTORY;
+import static jakarta.faces.FactoryFinder.VIEW_DECLARATION_LANGUAGE_FACTORY;
 import static jakarta.faces.application.ProjectStage.Development;
 import static jakarta.faces.application.ViewVisitOption.RETURN_AS_MINIMAL_IMPLICIT_OUTCOME;
 import static java.util.Collections.emptyList;
@@ -68,6 +68,7 @@ import jakarta.faces.event.SystemEvent;
 import jakarta.faces.event.SystemEventListener;
 import jakarta.faces.flow.FlowHandler;
 import jakarta.faces.flow.FlowHandlerFactory;
+import jakarta.faces.view.ViewDeclarationLanguageFactory;
 import jakarta.faces.view.facelets.FaceletCache;
 import jakarta.faces.view.facelets.FaceletCacheFactory;
 import jakarta.faces.view.facelets.TagDecorator;
@@ -303,14 +304,11 @@ public class ApplicationAssociate {
                 LOGGER.log(SEVERE, null, ex);
             }
 
-            // cause the Facelet VDL to be instantiated eagerly, so it can
-            // become aware of the resource library contracts
+            // Instantiate every view declaration language eagerly, so that the Facelets one
+            // becomes aware of the resource library contracts.
+            ((ViewDeclarationLanguageFactory) FactoryFinder.getFactory(VIEW_DECLARATION_LANGUAGE_FACTORY)).getAllViewDeclarationLanguages();
 
             ViewHandler viewHandler = context.getApplication().getViewHandler();
-
-            // FindBugs: ignore the return value, this is just to get the
-            // ctor called at this time.
-            viewHandler.getViewDeclarationLanguage(context, FACES_PREFIX + "xhtml");
 
             String facesConfigVersion = getFacesConfigXmlVersion(context);
             context.getExternalContext().getApplicationMap().put(FACES_CONFIG_VERSION, facesConfigVersion);
--- a/impl/src/main/java/com/sun/faces/application/view/FaceletViewHandlingStrategy.java
+++ b/impl/src/main/java/com/sun/faces/application/view/FaceletViewHandlingStrategy.java
@@ -46,7 +46,6 @@ import static jakarta.faces.application.Resource.COMPONENT_RESOURCE_KEY;
 import static jakarta.faces.application.StateManager.IS_BUILDING_INITIAL_STATE;
 import static jakarta.faces.application.StateManager.STATE_SAVING_METHOD_CLIENT;
 import static jakarta.faces.application.ViewHandler.CHARACTER_ENCODING_KEY;
-import static jakarta.faces.application.ViewHandler.DEFAULT_FACELETS_SUFFIX;
 import static jakarta.faces.application.ViewVisitOption.RETURN_AS_MINIMAL_IMPLICIT_OUTCOME;
 import static jakarta.faces.component.UIComponent.BEANINFO_KEY;
 import static jakarta.faces.component.UIComponent.COMPOSITE_FACET_NAME;
@@ -164,8 +163,8 @@ public class FaceletViewHandlingStrategy extends ViewHandlingStrategy {
 
     private DefaultFaceletFactory faceletFactory;
 
-    // Array of viewId extensions that should be handled by Facelets
-    private String[] extensionsArray;
+    // Array of suffixes which identify a resource as a Facelet
+    private String[] faceletResourceSuffixes;
 
     // Array of viewId prefixes that should be handled by Facelets
     private String[] prefixesArray;
@@ -826,11 +825,11 @@ public class FaceletViewHandlingStrategy extends ViewHandlingStrategy {
 
     /**
      * @param viewId the view ID to check
-     * @return <code>true</code> if assuming a default configuration and the view ID's extension in {@link ViewHandler#FACELETS_SUFFIX_PARAM_NAME}
-     * Otherwise try to match the view ID based on the configured extensions and prefixes in {@link ViewHandler#FACELETS_VIEW_MAPPINGS_PARAM_NAME}
+     * @return <code>true</code> if the view ID carries one of the suffixes which identify a Facelet, or matches one of
+     * the prefixes configured in {@link ViewHandler#FACELETS_VIEW_MAPPINGS_PARAM_NAME}, or is exact mapped to the
+     * {@code FacesServlet}.
      *
-     * @see com.sun.faces.config.WebConfiguration.WebContextInitParameter#FaceletsSuffix
-     * @see com.sun.faces.config.WebConfiguration.WebContextInitParameter#FaceletsViewMappings
+     * @see com.sun.faces.config.WebConfiguration#getFaceletResourceSuffixes()
      */
     @Override
     public boolean handlesViewId(String viewId) {
@@ -903,13 +902,14 @@ public class FaceletViewHandlingStrategy extends ViewHandlingStrategy {
      * Initialize mappings, during the first request.
      */
     protected void initializeMappings() {
+        faceletResourceSuffixes = webConfig.getFaceletResourceSuffixes();
+
         String viewMappings = webConfig.getOptionValue(FaceletsViewMappings);
         if (viewMappings != null && viewMappings.length() > 0) {
             Map<String, Object> appMap = FacesContext.getCurrentInstance().getExternalContext().getApplicationMap();
 
             String[] mappingsArray = split(appMap, viewMappings, ";");
 
-            List<String> extensionsList = new ArrayList<>(mappingsArray.length);
             List<String> prefixesList = new ArrayList<>(mappingsArray.length);
 
             for (String aMappingsArray : mappingsArray) {
@@ -919,15 +919,10 @@ public class FaceletViewHandlingStrategy extends ViewHandlingStrategy {
                     continue;
                 }
 
-                if (mapping.charAt(0) == '*') {
-                    extensionsList.add(mapping.substring(1));
-                } else if (mapping.charAt(mappingLength - 1) == '*') {
+                if (mapping.charAt(mappingLength - 1) == '*' && mapping.charAt(0) != '*') {
                     prefixesList.add(mapping.substring(0, mappingLength - 1));
                 }
             }
-
-            extensionsArray = new String[extensionsList.size()];
-            extensionsList.toArray(extensionsArray);
 
             prefixesArray = new String[prefixesList.size()];
             prefixesList.toArray(prefixesArray);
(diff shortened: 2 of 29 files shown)


BalusC:

Issue #5923, opened 2026-08-05
Case folding for ASCII comparisons depends on the default locale


BalusC:
Discovered while reassessing how `FacesServlet` guards against `WEB-INF` and `META-INF` probing, and then swept for the same pattern across the rest of the code base.

Several places fold case with `String.toUpperCase()` / `toLowerCase()` and then compare the result against ASCII literals. Both overloads use the JVM default locale, so under a Turkish or Azeri locale the result differs: `i` uppercases to dotted capital `İ` (U+0130) and `I` lowercases to dotless `ı` (U+0131). Every comparison against a literal spelled with `i` or `I` then fails, and the behaviour of the application depends on the locale of the server it is deployed on.

## Protected path check

`FacesServlet.notProcessWebInfIfPrefixMapped()` is the clearest case, because the guard is explicitly written to cover the lowercase spellings:

```java
pathInfo = pathInfo.toUpperCase();
if (pathInfo.contains("/WEB-INF/") || pathInfo.contains("/WEB-INF") || pathInfo.contains("/META-INF/") || pathInfo.contains("/META-INF")) {
```

```text
locale=en_US  /web-inf/web.xml       -> /WEB-INF/WEB.XML       blocked=true
locale=en_US  /meta-inf/MANIFEST.MF  -> /META-INF/MANIFEST.MF  blocked=true
locale=tr_TR  /web-inf/web.xml       -> /WEB-İNF/WEB.XML       blocked=false
locale=tr_TR  /meta-inf/MANIFEST.MF  -> /META-İNF/MANIFEST.MF  blocked=false
```

Reaching a protected file additionally requires a case-insensitive filesystem, since `ServletContext.getResource("/web-inf/web.xml")` must resolve. Disclosure has not been demonstrated end to end — `.xml` is not a configured Facelets suffix, so the view is most likely not handled at all — so this is filed as an ordinary correctness bug in a defense-in-depth check rather than as a security issue. The guard should nevertheless behave identically regardless of the deploying JVM's locale.

[...]

PR #5924, opened 2026-08-05
Make ASCII case folding independent of the default locale

BalusC:
Addresses #5923 on 4.0.

`String.toUpperCase()` and `toLowerCase()` without an argument fold with the JVM default locale. Under a Turkish or Azeri locale `i` uppercases to dotted capital `İ` (U+0130) and `I` lowercases to dotless `ı` (U+0131), so every subsequent comparison against an ASCII literal spelled with `i` or `I` fails and the behaviour of the application depends on the locale of the server it happens to be deployed on. Each commit pins one such fold to `Locale.ROOT`, or drops it where it is redundant.

One commit per site, so any of them can be reverted on its own:

| Commit | Site | Effect under `tr`/`az` |
|---|---|---|
| ff40e4bd3f | `FacesServlet` | `/faces/web-inf/web.xml` passes a guard explicitly written to cover the lowercase spelling |
| 6fd712c2c6 | `UIViewRoot` | `locale="IT"` resolves to language `ıt` |
| 9816991242 | `NamedEventConfigHandler`, `ViewScopeContextManager`, `FacesComponentTagLibrary` | `ItemBean` derives `ıtemBean`, so event, bean and tag names no longer match what the application writes in EL |
| ca8a6e1c9a | `HtmlResponseWriter` | `iso-8859-1` uppercases to `İSO-8859-1`, leaving ISO escaping enabled. Fails safe, output is only over-escaped |
| d79f1cd9bb | `TextRenderer` | `type="FILE"` folds to `fıle`, so the development-mode warning about a discommended type is skipped |
| a85307e33c | `ResourceHandlerImpl` | `TEXT/JAVASCRIPT` folds to `javascrıpt`, so no renderer type is derived |
| b1f0148fa1 | `ResourceManager` | Correct today, since the traversal literals carry no `i` or `I`. Pinned because a security check should not rest on a locale-dependent fold |

The `FacesServlet` commit also drops two dead checks. `contains("/WEB-INF/")` is subsumed by `contains("/WEB-INF")`, likewise for `/META-INF`. They are a fossil: the guard was introduced in 2005 as `startsWith("/WEB-INF/") || equals("/WEB-INF")`, where the two forms did different jobs, and was mechanically widened to `contains` for all four in 2012 without collapsing the now-redundant pair. The `service()` javadoc lists the same four strings and is spec text, so it is left as it is.

## Testing

`FacesServletTestCase` and `UIViewRootTest` each gained a regression test that exercises the fold under `en_US`, `tr_TR` and `az_AZ`. Both were verified to fail without their fix (`tr_TR must reject /web-inf/web.xml ==> expected: <404> but was: <200>` and `tr_TR ==> expected: <it> but was: <ıt>`) and to pass with it. The full `impl` unit suite is green: 86 classes, 0 failures, 0 errors.

The remaining five commits carry no test. Their seams are private or renderer-level, and each is a one-token change with no behavioural difference outside a Turkish or Azeri locale; adding a seam purely to test them did not seem worth the churn.

## Not in this PR

4.1 and master still need the port. The six non-`FacesServlet` sites are common to all branches, but on master the guard lives in `org.glassfish.mojarra.webapp.FacesServletImpl`, so that one is a hand-port rather than a cherry-pick.

`Util.getLocaleFromString` carries the same fold as `UIViewRoot` but reaches it only after `Locale.forLanguageTag` has already failed, which for a two letter language never happens, so its copy is unreachable and is left alone.

---

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/annotation/NamedEventConfigHandler.java
+++ b/impl/src/main/java/com/sun/faces/application/annotation/NamedEventConfigHandler.java
@@ -21,6 +21,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 
 import com.sun.faces.application.ApplicationAssociate;
@@ -88,7 +89,7 @@ public class NamedEventConfigHandler implements ConfigAnnotationHandler {
             name = name.substring(0, index);
         }
 
-        name = annotatedClass.getPackage().getName() + ("." + name.charAt(0)).toLowerCase() + name.substring(1);
+        name = annotatedClass.getPackage().getName() + ("." + name.charAt(0)).toLowerCase(Locale.ROOT) + name.substring(1);
         nem.addNamedEvent(name, (Class<? extends SystemEvent>) annotatedClass);
 
         String shortName = ((NamedEvent) annotation).shortName();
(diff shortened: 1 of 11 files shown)


BalusC:

PR #5926, opened 2026-08-06
Issue #5856: cut the per-component cost of pass-through attributes

BalusC:
Continues the view build time work of #5856, this time on the pass-through attribute path, which the `flat-*` scenarios of the perf bench ranked as the worst remaining cell: `flat-passthru` rendered at **1.5x MyFaces 4.1.4** while every other flat scenario sat at or below parity.

`flat-passthru` is a GET, so its `RENDER_RESPONSE` carries the Facelets build as well as the encode. A JFR split put Mojarra at 46% build / 51% encode against MyFaces' 42% / 55% — build +40%, encode +20% — and five per-component costs accounted for most of it.

## What was costing what

| | cost | fix |
|---|---|---|
| `UIComponentBase.PassThroughAttributesMap` extended `ConcurrentHashMap` | `put` was **9.9% of the whole Facelets build** (`casTabAt` + `addCount` per attribute) | extends `LinkedHashMap`. The contract in `UIComponent.getPassThroughAttributes(boolean)` is serializable, NPE on a null key or value, IAE on a non-String key — nothing about thread safety, and MyFaces backs the same map with the plain `StateHelper` map. `putAll` is overridden because the inherited bulk path bypasses the overridden `put` and would otherwise let nulls in |
| `HtmlResponseWriter.considerPassThroughAttributes` copied the map per element | `Traverser.advance` 4.5% + `putMapEntries` 2.6% of the encode | reads the component's map in place, copying only when it carries the `elementName` key. `clearPassthroughAttributes` no longer calls `clear()` — on an aliased map that would wipe the component's own attributes |
| `RenderKitUtils.renderPassThruAttributesOptimized` sorted `attributesThatAreSet` per component render | `binarySort` + `String.compareTo` = 7.3% of the encode, to re-sort a two-element list 400 times | `HtmlComponentUtils.handleAttribute` records the attributes in order instead |
| `getPassThroughAttributes(false)` went through `getStateHelper()` | allocated a `ComponentStateHelper` for a stateless component, on every element rendered — `startElement` consults it for all of them | takes `getStateHelper(false)` and returns null when there is none |
| `ComponentSupport.copyPassthroughAttributes` searched both pass-through namespaces per component tag applied | 3.8% + `TagAttributesImpl.getAll` 2.2% of the build, paid by every component tag whether or not it has a single `p:` attribute | `TagAttributesImpl` singles them out when the tag is compiled; the namespace search remains as a fallback for a non-standard `TagAttributes` |

## Numbers

Tomcat, the 400-component `test/perf` flat views, `min_us` (the fastest of 2000 requests, which is what survives background load on a shared machine), median of interleaved runs, Mojarra vs MyFaces 4.1.4-SNAPSHOT:

| scenario | phase | before | after | MyFaces | Δ | ratio |
|---|---|--:|--:|--:|--:|--:|
| flat-passthru | RENDER | 1323 | **1111** | 866 | **−16.0%** | 1.53x → **1.28x** |
| flat-readonly | RENDER | 1261 | 1188 | 1218 | −5.8% | 1.04x → 0.98x |
| flat-build | RENDER | 722 | 700 | 760 | −3.0% | 0.95x → 0.92x |
| flat-wide | RENDER | 1435 | 1394 | 1258 | −2.9% | 1.14x → 1.11x |
| flat-table-nofacets | RENDER | 978 | 953 | 1164 | −2.6% | 0.84x → 0.82x |
| flat-table-facets | RENDER | 900 | 896 | 995 | −0.4% | 0.90x → 0.90x |
| flat-text (control) | RENDER | 1215 | 1222 | 1345 | +0.6% | — |
| **flat suite total** | | **8708** | **8425** | **8728** | **−3.2%** | at MyFaces → **3.5% below** |

`flat-text` is the control: its markup is raw `<span>`s with no components, so none of these changes can reach it, and it does not move. `flat-readonly` is where the fifth lever shows up cleanest — 400 `h:outputText` with no pass-through attributes at all, i.e. pure per-tag overhead being removed.

## Verification

- **Rendered-output equivalence.** The perf WAR deployed twice on the same Tomcat with only `jakarta.faces.jar` swapped, all 39 scenario pages fetched, `jsessionid` and the `ViewState` value normalised, then diffed: **38 of 39 byte-identical**. `flat-passthru` differs only in the ordering of the pass-through attributes within each start tag — same byte count, and the new order is the one the view declares (`data-row data-col data-kind data-state data-index`) rather than the previous hash order.
- **Unit tests.** 1253 green. Eleven new ones pin the invariants this change relies on: the map's iteration order and its null/serializable contract; that rendering leaves the component's own map intact and that `elementName` still renames the element without being emitted; that recording an attribute keeps the list ordered and never records one twice; and that a tag's pass-through attributes are collected in declaration order across both namespaces.

One subtlety worth calling out, since it is the trap in this area: `attributesThatAreSet` has a **second writer**. `UIComponentBase.AttributesMap.put` appends any attribute that has no component property to the tail of that same list through `StateHelper.add`, so the list is not globally ordered and a binary-search insert can miss an entry that is already there and record it twice — which the render loop then emits as a duplicate HTML attribute. The per-render `Collections.sort` had been masking that. `handleAttribute` therefore scans rather than binary-searches; `HtmlAttributesThatAreSetTest.propertylessAttributesNeitherDuplicateNorDisplaceTheProperties` reproduces the duplicate against the binary-search version.

The measurements were taken on the 4.1 branch, where `test/perf` lives; the code here is identical apart from `instanceof` pattern matching, which 4.0 compiles at `-source 11` cannot use.

Closes nothing on its own — #5856 stays open for the remaining build gap, which is now diffuse: nothing above ~4% of the build, spread across component state maps, the attribute marker lookups and event publishing. Note that a profile of the Facelets build will show `ComponentTagHandlerDelegateImpl.apply`, `DefaultFaceletContext.generateUniqueId` and `ensurePrefix` trading a large self-time share between them from run to run — they are inlined into one another, and that share is an attribution artifact, not work. The id path itself is two field reads and an array increment in the common case, and accounts for 0.35% of the run's allocations.

### DIFF
--- a/impl/src/main/java/com/sun/faces/facelets/tag/TagAttributesImpl.java
+++ b/impl/src/main/java/com/sun/faces/facelets/tag/TagAttributesImpl.java
@@ -22,6 +22,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
+import com.sun.faces.facelets.tag.faces.PassThroughAttributeLibrary;
+
 import jakarta.faces.view.facelets.Tag;
 import jakarta.faces.view.facelets.TagAttribute;
 import jakarta.faces.view.facelets.TagAttributes;
@@ -42,6 +44,8 @@ public final class TagAttributesImpl extends TagAttributes {
 
     private final List nsattrs;
 
+    private final TagAttribute[] passthroughAttrs;
+
     private Tag tag;
 
     /**
@@ -49,6 +53,7 @@ public final class TagAttributesImpl extends TagAttributes {
      */
     public TagAttributesImpl(TagAttribute[] attrs) {
         this.attrs = attrs;
+        passthroughAttrs = collectPassthrough(attrs);
 
         // grab namespaces
         int i = 0;
@@ -75,6 +80,21 @@ public final class TagAttributesImpl extends TagAttributes {
         }
     }
 
+    private static TagAttribute[] collectPassthrough(TagAttribute[] attrs) {
+        List<TagAttribute> passthrough = null;
+
+        for (TagAttribute attr : attrs) {
+            if (PassThroughAttributeLibrary.NAMESPACES.contains(attr.getNamespace())) {
+                if (passthrough == null) {
+                    passthrough = new ArrayList<>(attrs.length);
+                }
+                passthrough.add(attr);
+            }
+        }
+
+        return passthrough == null ? EMPTY : passthrough.toArray(new TagAttribute[passthrough.size()]);
+    }
+
     /**
      * Return an array of all TagAttributesImpl in this set
      *
@@ -83,6 +103,17 @@ public final class TagAttributesImpl extends TagAttributes {
     @Override
     public TagAttribute[] getAll() {
         return attrs;
+    }
+
+    /**
+     * Return the attributes that are in a pass-through namespace, in the order the tag declares them. Every applied
+     * component tag is asked for these and almost none has any, so they are singled out once here, when the tag is
+     * compiled, rather than searched per namespace on every apply.
+     *
+     * @return a non-null array of TagAttribute
+     */
+    public TagAttribute[] getPassthroughAttributes() {
+        return passthroughAttrs;
     }
 
     /**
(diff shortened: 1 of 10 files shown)


BalusC:

PR #5928, opened 2026-08-06
Fix #5927: honor SKIP_ITERATION in UIRepeat, degrade rowStatePreserved cleanly

### LINKED ISSUE #5927: Full state saving corrupts ui:repeat and UIData: SKIP_ITERATION is ignored, and rowStatePreserved never resets row client IDs
Discovered while running 5.0 TCK with `-Dwebapp.partialStateSaving=false` since context param improvements in https://github.com/eclipse-ee4j/mojarra/issues/5903.

Two defects in row-scoped client ID handling. Both are invisible under partial state saving and both break the row-less/row-scoped client ID contract that full state saving depends on, so they are reported together.

## 1. `UIRepeat` ignores `SKIP_ITERATION` when the behavior source is inside it

Under full state saving, any `ui:repeat` containing the ajax source loses **all** component state for its children on postback, and the render-time facelet re-apply then duplicates the entire subtree.

`UIRepeat.requiresRowIteration(VisitContext)` overrides the caller's `SKIP_ITERATION` hint whenever the request's `jakarta.faces.source` starts with the repeat's client ID. `FaceletFullStateManagementStrategy.saveComponentState()` and `restoreComponentState()` both visit with `SKIP_ITERATION` and key the saved state by client ID, so they depend on that hint being honored. On the initial GET the override is inactive and state is filed under `form:repeat:checkbox`; on the ajax postback the override fires and the restore visit looks up `form:repeat:0:checkbox`, which is not in the state map. Every component inside the repeat is therefore restored with no state at all — including its `facelets.MARK_ID` — so `ComponentSupport.findChildByTagId` cannot match it during the render-time re-apply and the tag handlers append a second copy.

Instrumented trace of one ajax postback:

```
@@@SAVEKEY@@@    j_idt6:j_idt8:checkbox    HtmlSelectBooleanCheckbox
@@@RESTOREKEY@@@ j_idt6:j_idt8:0:checkbox  HtmlSelectBooleanCheckbox hasState=false
@@@APPLY@@@ tagId=…3db77da parent=HtmlForm/j_idt6  found=true   tag=ui:repeat
@@@APPLY@@@ tagId=…3db7734 parent=UIRepeat/j_idt8 found=false  tag=h:selectBooleanCheckbox
```

`jakarta.faces.component.UIData.requiresRowIteration` carries no such override — it is exactly `return !ctx.getHints().contains(VisitHint.SKIP_ITERATION);`. Aligning `UIRepeat` with the spec component fixes it:

```java
 private boolean requiresRowIteration(VisitContext ctx) {
-    boolean shouldIterate = !ctx.getHints().contains(VisitHint.SKIP_ITERATION);
-    if (!shouldIterate) {
-        FacesContext faces = ctx.getFacesContext();
-        String sourceId = BEHAVIOR_SOURCE_PARAM.getValue(faces);
-        boolean containsSource = sourceId != null ? sourceId.startsWith(super.getClientId(faces) + getSeparatorChar(faces)) : false;
-        return containsSource;
-    } else {
-        return shouldIterate;
-    }
+    return !ctx.getHints().contains(VisitHint.SKIP_ITERATION);
 }
```

## 2. `rowStatePreserved` renders every row under row 0's client ID

With `rowStatePreserved="true"` and full state saving, every row of a `ui:repeat` or `h:dataTable` renders its children under row 0's client ID. This shows on the plain initial GET, before any postback:

```html
<input id="form:outer:0:inner:0:input"><input id="form:outer:0:inner:0:input">
<input id="form:outer:0:inner:0:input">  <!-- all six cells of a 2x3 matrix -->
```

`UIRepeat.markInitialState()` and `UIData.markInitialState()` capture `initialDescendantFullComponentState` only when `StateManager.IS_BUILDING_INITIAL_STATE` is present. Full state saving never sets it, because `FaceletViewHandlingStrategy.markInitialState(ctx, view)` is gated on partial state saving, and the field is not part of `saveState` either. Both branches of `setRowIndexRowStatePreserved` / `setRowIndexWithRowStatePreserved` are then skipped — and the per-row client ID reset, `component.setId(component.getId())` in `restoreFullDescendantComponentStates` (commented *"reset the client id (see spec 3.1.6)"*), lives inside that same skipped branch.

Preserving full per-row state is inherently a partial-state-saving construct, so it cannot be honored here — but it must degrade to the ordinary per-row submitted-value state rather than stop managing client IDs. Select the branch on whether the state it needs is actually available:

```java
 public void setRowIndex(int rowIndex) {
-    if (isRowStatePreserved()) {
+    if (isRowStatePreserved() && _initialDescendantFullComponentState != null) {
         setRowIndexRowStatePreserved(rowIndex);
     } else {
         setRowIndexWithoutRowStatePreserved(rowIndex);
     }
 }
```

The same change applies to `UIRepeat.setIndex`. Note the `UIData` half lives in `jakarta.faces.component.UIData` in the API repository, so this defect spans both repositories.

Unlike defect 1 this one is not masked by the id-uniqueness walk: it reproduces in Development stage too.

## Affected branches

`requiresRowIteration` is byte-identical on `4.0`, `4.1` and `master`, and dates to the initial Oracle contribution in 2018 — it is not a regression. What changed is only whether it is observable:

| branch | `disableIdUniquenessCheck` default | full state saving |
|---|---|---|
| 4.0 | `"false"` — always check | masked |
| 4.1 | `"false"` — always check | masked |
| master (5.0) | `auto` — skipped unless Development | **exposed** since 28c663f31e |

`Util.checkIdUniqueness` walks the whole tree calling `getClientId()` on every node during `saveView`, which keeps the client ID key space row-less and accidentally hides defect 1. 4.0 and 4.1 still run that walk by default, but reach the bug as soon as `disableIdUniquenessCheck=true` is set, which is a documented performance opt-in.

Defect 2 is structural wherever the feature exists: `UIData.rowStatePreserved` long-standing, `UIRepeat.rowStatePreserved` since 4.1.

## Verification

Jakarta Faces TCK with `-Dwebapp.partialStateSaving=false`.

Fixed by defect 1's change: `Issue1821IT`, `Issue2640IT`, `Issue2910IT` (faces22/ajax-lifecycle), `Issue1581IT`, `Issue5078IT` (faces23/facelets).

Fixed by defect 2's change: `NestedRepeatRepeatRowStatePreservedIT`, `NestedTableTableRowStatePreservedIT` (faces23/faces-data-model).

Fully green under full state saving with both: faces22/ajax-lifecycle 22/22, faces23/faces-data-model 25/25, faces41/api 18/18, faces50/facelets 33/33. No partial-state-saving regressions across 250 tests in the 12 TCK modules that combine `ui:repeat` or `h:dataTable` with ajax.

BalusC:
Closes #5927.

`UIRepeat.requiresRowIteration` now honors `SKIP_ITERATION` unconditionally, matching `UIData`. It is a literal revert of 42a30d96bd (JAVASERVERFACES-3833, 2015), whose own regression test — `Issue3837IT` in the TCK — passes without it under both state saving modes.

`UIData.setRowIndex` only takes the `rowStatePreserved` path when the initial descendant state that drives it is actually available, instead of silently leaving the descendant client IDs unmanaged.

`mvn -pl impl clean install`: 1253 tests, 0 failures.

Behavioural verification was done on the 5.0 branch against the Jakarta Faces TCK, where both changes are identical; the 5.0 TCK cannot run against a 4.0 API. With `-Dwebapp.partialStateSaving=false` the first change fixes `Issue1821IT`, `Issue2640IT`, `Issue2910IT`, `Issue1581IT` and `Issue5078IT`, and the second fixes `NestedRepeatRepeatRowStatePreservedIT` and `NestedTableTableRowStatePreservedIT`. No partial-state-saving regressions across 250 tests in the 12 TCK modules combining `ui:repeat` or `h:dataTable` with ajax.

Note `Issue3837IT` is `@Disabled` in the TCK per the accepted challenge in jakartaee/faces#1757, so a plain TCK run does not exercise the scenario the reverted check was added for; it has to be enabled explicitly.

### DIFF
--- a/impl/src/main/java/com/sun/faces/facelets/component/UIRepeat.java
+++ b/impl/src/main/java/com/sun/faces/facelets/component/UIRepeat.java
@@ -17,7 +17,6 @@
 package com.sun.faces.facelets.component;
 
 import static com.sun.faces.cdi.CdiUtils.createDataModel;
-import static com.sun.faces.renderkit.RenderKitUtils.PredefinedPostbackParameter.BEHAVIOR_SOURCE_PARAM;
 import static com.sun.faces.util.Util.isNestedInIterator;
 
 import java.io.IOException;
@@ -762,16 +761,11 @@ public class UIRepeat extends UINamingContainer {
         return false;
     }
 
+    // SKIP_ITERATION is honored unconditionally, as UIData does. A visit that asks for it wants the
+    // row-less clientIds: full state saving keys the saved component state by them, so iterating anyway
+    // makes the restore look up repeat:N:foo for state filed under repeat:foo and restore nothing.
     private boolean requiresRowIteration(VisitContext ctx) {
-        boolean shouldIterate = !ctx.getHints().contains(VisitHint.SKIP_ITERATION);
-        if (!shouldIterate) {
-            FacesContext faces = ctx.getFacesContext();
-            String sourceId = BEHAVIOR_SOURCE_PARAM.getValue(faces);
-            boolean containsSource = sourceId != null ? sourceId.startsWith(super.getClientId(faces) + getSeparatorChar(faces)) : false;
-            return containsSource;
-        } else {
-            return shouldIterate;
-        }
+        return !ctx.getHints().contains(VisitHint.SKIP_ITERATION);
     }
 
     // Tests whether we need to visit our children as part of
--- a/impl/src/main/java/jakarta/faces/component/UIData.java
+++ b/impl/src/main/java/jakarta/faces/component/UIData.java
@@ -485,7 +485,11 @@ public class UIData extends UIComponentBase implements NamingContainer, UniqueId
      * @throws IllegalArgumentException if <code>rowIndex</code> is less than -1
      */
     public void setRowIndex(int rowIndex) {
-        if (isRowStatePreserved()) {
+        // Preserving the full per-row state needs the initial descendant state that markInitialState()
+        // captures; without it that path can neither save nor restore a row, and in particular never
+        // resets the descendant clientIds, so every row would render row 0's ids. Fall back to the
+        // per-row submitted-value state, which keeps the ids correct.
+        if (isRowStatePreserved() && _initialDescendantFullComponentState != null) {
             setRowIndexRowStatePreserved(rowIndex);
         } else {
             setRowIndexWithoutRowStatePreserved(rowIndex);


BalusC:

PR #5931, opened 2026-08-07
Give the perf suite a real pass-through control and full tag coverage

BalusC:
Three changes to the perf suite, all about measuring the right thing rather than about impl code.

## flat-nopassthru replaces flat-wide

`flat-wide` is documented as "the control that isolates passthru", but it is 400 `h:outputText` carrying five renderer properties while `flat-passthru` is 400 `h:panelGroup` carrying five `p:data-*` attributes. Different component, different renderer, different attribute kind — subtracting them folds the whole renderer difference into the delta and isolates nothing.

`flat-nopassthru` is the same `h:panelGroup layout="block" styleClass="cell"` tree at the same count, with the pass-through attributes removed and nothing else changed. They share a multiplier in the generator so the counts cannot drift apart.

What that buys, on Tomcat against MyFaces 4.1.4-SNAPSHOT (warmup 200, runs 2000, avg µs/request, RENDER_RESPONSE):

| scenario | Mojarra 4.1.14-SNAPSHOT | MyFaces 4.1.4-SNAPSHOT | Δ |
|---|--:|--:|--:|
| flat-nopassthru | 649 | 556 | +17% |
| flat-passthru | 1194 | 958 | +25% |

So the pass-through increment is 545µs vs 402µs, and on `min` (the load-robust statistic) 470µs vs 382µs, i.e. **+23%** — while the bare component underneath is already **+17%**. Roughly two thirds of `flat-passthru`'s gap turns out not to be the pass-through path at all. That split was not visible with `flat-wide` as the control.

## flat-allattrs is new

One of every tag `faces.html.taglib.xml` declares, each carrying every attribute declared for it — 34 tags, 941 attributes on 4.1. The attribute set is read from the taglib rather than written out, so the page follows the taglib per branch instead of rotting.

Only `binding` is skipped: it moves component creation to a bean, so the components would no longer come from `Application.createComponent` and the page would stop measuring the build path it exists for. `rendered` is an expression rather than a literal, since a literal is folded at tag-compile time and would never reach the per-component expression path.

Attributes that are alternatives to each other are split across variants rather than dropped, so every declared attribute is still covered:

| tag | v1 | v2 | v3 |
|---|---|---|---|
| graphicImage | `library`+`name` | `url` | `value` |
| dataTable | `rowClasses` | `rowClass` | |
| panelGrid | `rowClasses` | `rowClass` | |

`BaseTableRenderer.renderRowStart` throws `IOException: Cannot define both rowClasses on a table and rowClass`, so the table split is required, not cosmetic. Two attributes are genuinely dropped, both converters facing a value that is not text: `selectBooleanCheckbox` (Boolean) and `inputFile` (Part).

Verified by rendered-output comparison that both impls emit the same thing — identical counts across 14 element types (24 `input`, 8 `select`, 24 `option`, 54 `td`, 48 `span`, …), bytes within 1%. Worth doing explicitly: `PerfBenchIT` asserts HTTP 200 and nothing about output equivalence, so a scenario where one impl renders less silently measures a difference in markup rather than in speed.

## buildview keeps only its sweeps

`buildview/` holds size points now — the `scale-*` and `tb-cols-*` sweeps, whose fitted slope is per-component and per-column build cost, which no single-sized scenario can report. The shape variants that measured attribute and facet shape (`scale-400-bare`/`-wide`/`-passthru`, `tb-cols-200-nofacet`/`-twotext`, `tb-nofacet`, `tb-noconv`, `tb-noform`, `tb-plain`, `tb-plain-200`) go: the `flat-*` scenarios measure the same shapes on the render side with the component count held fixed, and the facet pair there varies nothing but the facet, where `tb-cols-200`/`-twotext` also swapped a literal-text node for an `h:outputText`.

`txt-el`/`txt-lit` stay. They isolate the literal-text and instruction-node build path, which is what `ComponentSupport.isBuildingNewComponentTree` is called on once per node and what the `isPostback` lever in #5929 targeted; `flat-text` covers only the EL side and only inside the lifecycle.

## Note for whoever reads a number after this

The scenario mix changed, so this re-baselines the suite — the mix itself moves the absolute figures, and no result from a previous round is comparable to one taken after this commit.


BalusC:

PR #5932, opened 2026-08-07
Stop applying a pass-through attribute as an ordinary one too

BalusC:
A pass-through attribute is currently applied twice: once to the component's pass-through attribute map, where it belongs, and once to the component's ordinary attributes map, where it does not.

`MetaRulesetImpl` keys a tag's attributes by local name and never consults the namespace, so `p:data-row` enters the ruleset as `data-row`. `ComponentTagHandlerDelegateImpl.createMetaRuleset` ignores only `binding` and `id`, and `finish()` walks the rules backwards, so `ComponentRule` sees it first: the attribute is literal and `getWriteMethod("data-row")` is null, so it returns a `LiteralAttributeMetadata` whose `applyMetadata` runs `component.getAttributes().put("data-row", value)` on every build — on top of what `ComponentSupport.copyPassthroughAttributes` already did.

Observed on Tomcat with the same WAR and only the implementation swapped, reading the built component back through EL:

| | Mojarra 4.0 | MyFaces 4.1.4 |
|---|---|---|
| `passThroughAttributes['data-row']` | `ROWVAL` | `ROWVAL` |
| `attributes['data-row']` | **`ROWVAL`** | empty |
| `attributesThatAreSet` | **`[data-row, styleClass, data-col]`** | empty |

Both render the same markup; the difference is entirely internal.

## Why it costs

The second write takes `AttributesMap`'s descriptor-miss path, per attribute per build: a `getPropertyDescriptor` lookup that cannot hit, a `contains` scan of `attributesThatAreSet`, a list append and a state-helper write.

Landing in `attributesThatAreSet` has two further effects. The attribute is saved into component state a second time. And `RenderKitUtils.renderPassThruAttributes` iterates exactly that list, so every render walks entries that can never match a renderer attribute — a `knownAttributes.get` miss plus an `isBehaviorEventAttribute` miss each, per attribute per component per render.

## Why it is also wrong

Because the ruleset is keyed by local name alone, a pass-through attribute whose local name is a real property's displaces that property. Before this change `<h:panelGroup layout="block" p:layout="inline"/>` rendered as `inline`. A test pins that.

## Measurements

Isolated buildView (`BuildViewBenchIT`), Tomcat, 407 components each carrying five pass-through attributes, min of three passes:

| | before | after | MyFaces 4.1.4 |
|---|--:|--:|--:|
| ns per build | 487,273 | **315,811** | 286,990 |
| pass-through attributes alone | 206,019 | **67,568** | 69,890 |
| per attribute | 103 ns | **34 ns** | 35 ns |

buildView drops 35%, and the attributes themselves stop costing three times what they cost MyFaces — the remaining 34 ns is the pass-through map write both implementations do.

The render-side effect of the shorter `attributesThatAreSet` list is not measured here.

## Verification

Three unit tests, all failing before the change: the attribute is absent from `getAttributes()` for both pass-through namespaces and `attributesThatAreSet` stays empty; an ordinary attribute is still applied; a pass-through attribute no longer displaces the property of the same name. Full impl suite green. The full TCK was run green against this change ported to 5.0.

### DIFF
--- a/impl/src/main/java/com/sun/faces/facelets/tag/MetaRulesetImpl.java
+++ b/impl/src/main/java/com/sun/faces/facelets/tag/MetaRulesetImpl.java
@@ -28,6 +28,7 @@ import java.util.WeakHashMap;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import com.sun.faces.facelets.tag.faces.PassThroughAttributeLibrary;
 import com.sun.faces.util.FacesLogger;
 import com.sun.faces.util.Util;
 
@@ -70,6 +71,9 @@ public class MetaRulesetImpl extends MetaRuleset {
         // setup attributes
         TagAttribute[] attrs = this.tag.getAttributes().getAll();
         for (int i = 0; i < attrs.length; i++) {
+            if (PassThroughAttributeLibrary.NAMESPACES.contains(attrs[i].getNamespace())) {
+                continue;
+            }
             if (attrs[i].getLocalName().equals("class")) {
                 attributes.put("styleClass", attrs[i]);
             } else {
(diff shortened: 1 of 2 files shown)


BalusC:

PR #5933, opened 2026-08-08
Issue #5856: cut the per-component cost of reading component attributes

BalusC:
Follow-up to #5926 and #5929 on #5856, this time on the render side. Profiling the `flat-nopassthru` scenario (400 bare `h:panelGroup layout="block" styleClass="cell"`, a GET so its RENDER carries the Facelets build) put Mojarra ~15% behind MyFaces, the largest remaining render deficit in the suite now that `flat-passthru` is at parity.

The gap is not reflection. Both implementations read component attributes through a reflective property map and spend the same doing it — `DirectMethodHandleAccessor` presence inside `GroupRenderer` is 14.81% (95 µs) on Mojarra against 16.52% (96 µs) on MyFaces. What differs is string and map traffic: `String.equals` 3.14% vs 0.59% and `HashMap.getNode` 1.84% vs 0.59%, together roughly the whole 25 µs by which `GroupRenderer` is slower.

A counting probe explains it exactly. Per request on that view, `AttributesMap.get` is called 3222 times — 8 per component — of which 1606 (4 per component) are the `attributesThatAreSet` key and 800 (2 per component) are reflective property reads of `styleClass`, once in `encodeBegin` and again in `encodeEnd`. Every one of those 1606 lookups ran `markerGet`'s six `String.equals` against the Facelets marker constants and then a full sixty-character `equals`, because `RenderKitUtils` held a separately computed instance of the same key, before reaching the state helper.

## What is in here

1. **`answer the set-attribute list by identity`** — the key becomes a literal everywhere, so every caller passes the same interned instance and `AttributesMap.get`/`containsKey` settle it on a reference check ahead of the marker keys, with the full `equals` left in place for a foreign key. The key and the package it gates on move to `UIComponentBase`, where `ComponentStateHelper` can read what `AttributesMap` writes. The dead `UIComponent.attributesThatAreSet` field goes too — the state helper has held that list since partial state saving arrived in 2009, so the field has had no writer since and the guard reading it was always taken. `VIEW_ROOT_ATTRIBUTES_DOCTYPE_KEY` goes as well, unused since `isOutputHtml5Doctype` started reading `UIViewRoot.getDoctype()`.
2. **`read panel group attributes through their typed getters`** — `GroupRenderer` read `layout` through `HtmlPanelGroup` but `styleClass` and `style` through the attributes map. Now all three go through the typed getters, with the generic path kept as the fallback for non-standard components.
3. **`hold the row index of UIData on a field`** — every component under a table reads the row index through `getClientId` while its own client id is being built. It is not component state: every phase leaves it at -1 before state is saved and no page can set it. This is what `UIRepeat` and MyFaces' own `UIData` already do. One semantic narrowing worth naming rather than leaving to be found: `getStateHelper().eval` fell back to `getValueExpression("rowIndex")` when the map held no value, and a field cannot. It is unreachable from a page — `rowIndex` is not a declared `h:dataTable` attribute, so only a programmatic `setValueExpression("rowIndex", ...)` against a bound component could get there, which neither `UIRepeat` nor MyFaces honours either. The full TCK is green with it.
4. **`give UIInstructionHandler a unique-id slot`** — component tags have counted their generated ids through a reserved slot since #5929; text and EL nodes still counted through the tag-id map, which grows and rehashes once per distinct tag id per build. The reservation `ComponentTagHandlerDelegateImpl` held privately is extracted so both handlers share it.
5. **`record the attributes that are set only for standard components`** — a bug found while reading the code above. `ComponentStateHelper.handleAttribute` gated on `getClass()`, which is the helper itself — package private in `jakarta.faces.component`, so the condition was always true, and restoring state started the tracking list for every component including third-party ones whose setters never maintain such a list. Harmless in effect, since `RenderKitUtils.getAttributeIfSet` and `canBeOptimized` both gate on the *component's* package and a foreign component therefore takes the unoptimized path and never reads the list — the cost is state that is saved and never used, on the partial state path as much as the full one. Two tests come with it.

6. **`release the FacesContext that FacesContextImplPostbackTest creates`** — also unrelated to the levers, also found while verifying them. Constructing a `FacesContext` makes it the current one for the thread and the test never cleared it, so a stale context stayed current for every test that ran after this class in the JVM. `FactoryFinder` keys its per-classloader cache on the current `FacesContext`, so the entry `releaseFactories()` removes in `setUp` is not necessarily the entry the following `getFactory()` resolves — see the closing note.

## Measurements

Tomcat, 400-component flat views, two independent interleaved rounds of 4 reps per arm, medians of `min_us` (see the note on why `min_us` is the load-robust metric for A/B):

| scenario, RENDER | round 1 | round 2 |
|---|--:|--:|
| **flat-nopassthru** | **-16.5%** | **-15.6%** |
| **flat-passthru** | **-11.7%** | **-12.1%** |
| flat-build | -2.5% | -1.4% |
| flat-readonly, flat-text, flat-table-facets, flat-table-nofacets, flat-allattrs | ±0-4%, inside the per-cell spread | ±0-4%, inside the per-cell spread |

`flat-nopassthru` was 597 µs against MyFaces at 525 and is now 498, so that cell goes from +13.7% to -5.1%. Suite total for the `flat-*` family: -2.4% and -1.2% across the two rounds.

That table measures commits 1 and 2. Commit 3 is present in **both** of its arms, so its effect does not appear there — it was measured separately, in paired rounds with the jar swapped between back-to-back runs (a third round is omitted: the machine was not idle for it, so its baseline drifted):

| scenario, RESTORE | round 1 | round 2 |
|---|--:|--:|
| **flat-table-nofacets** | **-22%** | **-19%** |
| **flat-table-facets** | **-20%** | **-20%** |
| table-build | -2% | -8% |
| table-readonly | +6% | -8% |
| flat-build *(control, no table)* | -2% | -6% |

Both table scenarios move by about **-105 µs per request**, which takes `flat-table-nofacets` from +33% against MyFaces to roughly +6% — the largest single move in this set. They are postbacks, where buildView lands in RESTORE_VIEW and a per-component state helper lookup on the client-id path costs the most.

The controls behave as controls should: `flat-build` has no `h:dataTable` and swings across zero, and `table-readonly`'s RESTORE_VIEW is ~47 µs, below the base at which a percentage carries meaning.

Commit 4 is a DRY change whose measured effect is inside the noise band (about -27 µs on that same phase); it is in here because both handlers should share one implementation, not because it moves the number.

## Verification

- Full Faces TCK green against the 5.0 port of this change set (`jakarta.faces` API changes in the `faces` repository, implementation changes here).
- `faces_render_diff.sh` — the perf WAR deployed twice on the same Tomcat with only the Faces jar swapped, all 40 scenario pages fetched and diffed after normalising jsessionid and ViewState: **40/40 byte-identical** on 4.1 (825872 bytes both arms) and on 5.0 (839998 bytes both arms). That check matters here because commit 2 changes how a renderer decides its element name and its `class` attribute. It cannot run on 4.0, where `test/perf` does not exist yet.
- `impl` unit tests green, including the two new `ComponentStateHelperTest` cases. The negative one was confirmed to fail against the unfixed gate with `expected: <null> but was: <[rendered]>`.

## Notes for the upmerge

4.0 is `source/target 11`, so `GroupRenderer`'s helpers use `instanceof HtmlPanelGroup` plus a cast rather than a pattern variable, and `UniqueIdSlot` uses a nested class rather than a record. 4.1 and 5.0 both want the modern form, so expect a small conflict on exactly those lines. On 5.0 the `jakarta.faces.component` half of commits 1, 3 and 5 lives in the `faces` submodule and has to move there by hand; the constants hoisted onto `UIComponentBase` here belong next to the marker keys in `PackageUtils` on that branch.

On the `FacesContextImplPostbackTest` flake (#5930): it surfaced once in roughly nineteen full-suite runs here, as `IllegalArgumentException: HTML_BASIC` from `addRenderKit`, i.e. a duplicate registration on a factory that survived `releaseFactories()`. Commit 6 removes one mechanism that can produce exactly that — the leaked current `FacesContext` participating in `FactoryFinder`'s cache key — but **it is not proven to be the cause**: the flake did not reproduce in twelve instrumented full-suite runs, so there was nothing to fix-and-confirm against. The commit stands on its own merits regardless, since a test that creates a `FacesContext` should not leave it current for every test after it. If it does recur, the leak was not it, and the next step is making the registration idempotent through `getRenderKitIds()` — note that `getRenderKit()` throws when the kit is *absent*, so it cannot be used for that check.

### DIFF
--- a/impl/src/main/java/com/sun/faces/facelets/UniqueIdSlot.java
+++ b/impl/src/main/java/com/sun/faces/facelets/UniqueIdSlot.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) Contributors to Eclipse Foundation.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package com.sun.faces.facelets;
+
+import jakarta.faces.view.facelets.Facelet;
+import jakarta.faces.view.facelets.FaceletContext;
+
+/**
+ * The unique-id counter slot a tag handler reserves once and reuses for every build.
+ * <p>
+ * A handler that asks for its id by tag id makes the context count in a map keyed by that id, which is one lookup and,
+ * while the map is still growing, a share of a rehash for every tag applied. Reserving a slot moves the counter into an
+ * array indexed per tag, so the same id costs an array increment. The slot belongs to the {@link Facelet} it was
+ * reserved from, so it is re-reserved whenever a handler is applied from a different one.
+ * <p>
+ * A handler holds one of these per instance and is shared across concurrent builds, hence the volatile: a stale read
+ * only costs a re-reservation, never a wrong id.
+ */
+public final class UniqueIdSlot {
+
+    private volatile Reservation reservation;
+
+    /**
+     * Returns the next unique id for {@code tagId}, through the reserved slot where the context supports it and
+     * through the public {@link FaceletContext} API otherwise, which a wrapping context or a foreign implementation
+     * may supply.
+     *
+     * @param ctx the context to generate the id from
+     * @param tagId the tag id to generate an id for
+     * @return the generated unique id
+     */
+    public String generateUniqueId(FaceletContext ctx, String tagId) {
+        if (!(ctx instanceof FaceletContextImplBase)) {
+            return ctx.generateUniqueId(tagId);
+        }
+
+        FaceletContextImplBase context = (FaceletContextImplBase) ctx;
+        Facelet owner = context.getUniqueIdSlotOwner();
+
+        if (owner == null) {
+            return ctx.generateUniqueId(tagId);
+        }
+
+        Reservation current = reservation;
+
+        if (current == null || current.owner != owner) {
+            current = new Reservation(owner, context.getUniqueIdSlot(tagId));
+            reservation = current;
+        }
+
+        return context.generateUniqueId(tagId, current.owner, current.slot);
+    }
+
+    private static final class Reservation {
+
+        private final Facelet owner;
+        private final int slot;
+
+        private Reservation(Facelet owner, int slot) {
+            this.owner = owner;
+            this.slot = slot;
+        }
+    }
+}
--- a/impl/src/main/java/com/sun/faces/facelets/compiler/UIInstructionHandler.java
+++ b/impl/src/main/java/com/sun/faces/facelets/compiler/UIInstructionHandler.java
@@ -19,6 +19,7 @@ package com.sun.faces.facelets.compiler;
 import java.io.IOException;
 import java.io.Writer;
 
+import com.sun.faces.facelets.UniqueIdSlot;
 import com.sun.faces.facelets.el.ELText;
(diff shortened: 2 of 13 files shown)


BalusC:

PR #5934, opened 2026-08-10
Issue #5856: cut the per-component id cost of a view build

BalusC:
#5856 

Follow-up to #5933 on the same issue. `RESTORE_VIEW` on a postback is 83% Facelets rebuild (state restore itself does not register in a JFR profile), and the per-component id path is 13% of that. This takes two levers off it. Both are generic view-build wins, so they also pay on the render-side build.

### `UIComponentBase.validateId` — ASCII fast path

Every generated id (`j_id_2c` and friends) was classified with `Character.isLetter`/`isDigit`, i.e. full Unicode lookup via `Character.getType`, which showed up at 1.2% self-time in the profile with the whole check at 3.6% of the phase. Ids are ASCII by construction, so `a-zA-Z0-9-_` is now answered directly and `Character` is consulted only above `0x7F`, where it still decides — at or below it `isLetter` matches exactly `a-zA-Z` and `isDigit` exactly `0-9`, so the verdicts are unchanged. The `i == 0` test also moves out of the loop.

### `IdMapper` off the per-component path

`IdMapper.getMapper(ctx)` is a `Util.notNull` plus a lookup in the `FacesContext` attributes map, and it ran once per component tag *and* once per text node — 4.6% of the phase. But `DefaultFacelet.apply` installs the mapper immediately before it constructs the `DefaultFaceletContext` and removes it only once the whole build is done, so it cannot change while any context in the chain is alive. `DefaultFaceletContext` now holds it as a field (inherited through the nested-context constructor) and answers a new `FaceletContextImplBase.getAliasedId(String)` from it. A wrapping context or a foreign `FaceletContext` implementation keeps the old map lookup, via `ComponentSupport.getAliasedId(FaceletContext, String)`.

### Measurements

Two independent interleaved A/B rounds, Tomcat, 4 reps per arm, 4000 requests per scenario, medians of `avg_us`:

| cell | round 1 | round 2 |
|---|--:|--:|
| flat-table-nofacets RESTORE_VIEW | -3.4% | -5.1% |
| flat-table-facets RESTORE_VIEW | -2.4% | -3.0% |
| flat-build RESTORE_VIEW | -4.6% | -3.2% |
| suite total | -1.1% | -1.5% |

Nothing with a mechanism regressed. Round 2 also carried a `validateId`-only arm: it accounts for most of the movement on the `flat-table-*` cells, with the `IdMapper` hoist worth roughly a further point. Per-lever attribution is only weakly resolved at this noise floor, so treat that split as indicative.

Note that `min_us` is the wrong metric for `RESTORE_VIEW` on the postback scenarios: `PerfBenchIT` issues one GET before the postback block, and that GET's restore (~23µs against ~450µs) is always the minimum, so the column measures the GET.

### Verification

- 1289 unit tests green, including two new ones. `UIComponentIdValidationTest` pins the id alphabet on both sides of the ASCII boundary (a non-ASCII letter is accepted, a non-ASCII digit is rejected as the first character but accepted after it, a non-ASCII character that is neither is rejected). `FaceletContextIdMapperTest` pins that an included Facelet aliases through the outermost build's mapper and that an id is its own alias when no mapper is installed.
- Render output byte-identical across all 40 `test/perf` views (837k bytes both arms), which is the real check here since both levers feed generated component ids.
- Full Faces TCK green against the 5.0 port of this change.

### DIFF
--- a/impl/src/main/java/com/sun/faces/facelets/FaceletContextImplBase.java
+++ b/impl/src/main/java/com/sun/faces/facelets/FaceletContextImplBase.java
@@ -18,6 +18,8 @@ package com.sun.faces.facelets;
 
 import java.io.IOException;
 
+import com.sun.faces.facelets.impl.IdMapper;
+
 import jakarta.el.ELException;
 import jakarta.faces.FacesException;
 import jakarta.faces.component.UIComponent;
@@ -66,6 +68,19 @@ public abstract class FaceletContextImplBase extends FaceletContext {
      */
     public String generateUniqueId(String base, Facelet owner, int slot) {
         return generateUniqueId(base);
+    }
+
+    /**
+     * Returns the alias the {@link IdMapper} in effect for this build gives to the given id, or the id itself when no
+     * mapper is in effect. The mapper is fixed for the duration of a build, so an implementation that resolves it once
+     * answers this without the lookup in {@link jakarta.faces.context.FacesContext#getAttributes()} that this default
+     * performs per call.
+     *
+     * @param id the id to alias
+     * @return the aliased id
+     */
+    public String getAliasedId(String id) {
+        return IdMapper.getAliasedId(getFacesContext(), id);
     }
 
     /**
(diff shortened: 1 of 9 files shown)


BalusC:

PR #5935, opened 2026-08-10
Lazily allocate HtmlResponseWriter CDATA buffers

BalusC:
#5856 

### What

Allocate `HtmlResponseWriter`'s two CDATA buffers on first use instead of in field initializers.

`cdataBuffer` (1KB) and `cdataTextBuffer` (128B) are only touched on the `writingCdata` path — `<script>`/`<style>`/CDATA and ajax-partial content. A plain page never enters it, so every writer was paying ~1.1KB of `char[]` for buffers it would not use, and several writers are built per render: the main one, the `WriteBehindStateWriter` clone, and one per `MenuRenderer` select (which clones the whole writer just to buffer its options for the size count).

`flushBuffer` still reads the field raw — that is safe because it returns early on `cdataBufferLength == 0`, so it only reaches the buffer after a write allocated it.

### Measurements

Tomcat, `test/perf` `flat-allattrs`, 8000 runs, JFR allocation profile of RENDER_RESPONSE:

| | before | after |
|---|--:|--:|
| `cloneWithWriter` | 155.4 MB | 27.0 MB |
| `char[]` | 406.7 MB | 251.1 MB |
| RENDER total | 2501 MB | 2379 MB |

Roughly 16KB per request. **Wall time is unchanged** — 4 interleaved jar-swapped reps put `flat-allattrs` render at +0.5% on `min_us` with a 2.5% spread, and the suite total at +0.1%. This reduces allocation pressure and GC frequency; it is not a throughput win and should not be read as one.

1289 impl unit tests green.

### Note

This is a backport of e43da38d04, which landed on master in July and never reached 4.0 or 4.1. The content is identical. Once this upmerges 4.0 → 4.1 → master, the master merge will collide with the commit already there; take either side, they are the same change.

### DIFF
--- a/impl/src/main/java/com/sun/faces/renderkit/html_basic/HtmlResponseWriter.java
+++ b/impl/src/main/java/com/sun/faces/renderkit/html_basic/HtmlResponseWriter.java
@@ -125,14 +125,15 @@ public class HtmlResponseWriter extends ResponseWriter {
     // Enables scripts to be included in attribute values
     private Boolean isScriptInAttributeValueEnabled;
 
-    // Internal buffer used when outputting properly escaped CData information.
-    //
+    // Internal buffers used when outputting properly escaped CData information. Lazily allocated on first
+    // CDATA/script/style write: a plain page never enters the writingCdata path, so eagerly allocating these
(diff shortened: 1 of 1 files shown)


BalusC:

PR #5936, opened 2026-08-10
UIValidateWholeBean avoid multiple copy of UIComponent children

pizzi80:
UIValidateWholeBean avoid multiple copy of UIComponent children

Util.reverse
 - return a reversed view of the List (no copy)
 - deprecated because is superseeded by Java 21 List.reversed

### DIFF
--- a/impl/src/main/java/com/sun/faces/util/Util.java
+++ b/impl/src/main/java/com/sun/faces/util/Util.java
@@ -56,6 +56,7 @@ import java.util.Enumeration;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.ListIterator;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


impl/src/main/java/com/sun/faces/util/Util.java
```diff
@@ -737,15 +738,18 @@ public static <T> T coalesce(T... objects) {
...
-        for (int i = length - 1; i >= 0; i--) {
-            result.add(list.get(i));
-        }
-
-        return result;
+    /**
+     * @deprecated use Java 21 List#reversed()
+     * @return a view of the passed {@link List} with all the elements in reverse order
+     */
+    @Deprecated(forRemoval = true, since = "5.0")
```
BalusC, line 745: 2 things: 
- since="5.0" should be put in 5.0 branch not 4.1 branch
- 5.0 uses Java 17 not 21

BalusC:
Superseded by https://github.com/eclipse-ee4j/mojarra/pull/5939

Thanks for bringing this up! This indeed warranted an improvement.

PR #5937, opened 2026-08-10
Util.split framework madness (phase A)

pizzi80:
the Util.split framework is a pre-Java 7 utility 
that try to pass every single call to string.split through a Pattern cache

1) this is a pre-Java7 problem for the majority of split use-cases

2) there are a lot of inefficient calls to these methods that retrieve 
    the FacesContext + ExternalContext + ApplicationMap 
    even inside loops only to split a string on a comma char ... 

3) the Util.split pattern cache use an inefficient synchronizedMap over an LRUMap
    with a 15 Patterns capacity! 

4) caching only 15 Patterns means that if we have 16 or 20 different Patterns inside mojarra
    every call to split will create a new Pattern on the fly after a blocking call to the Pattern cache

5) the Pattern cache lives in a ServletContext attribute, but the are 2 main entry points
    (the ServletContext and the ApplicationMap) but the lazy initialization is not synchronized
    so if you deploy a mojarra webapp under load you will initialize multiple instance of the cache
    and so a lot of split will have to compute a Pattern on the fly and store a new cache in the ServletContext

this is a first PR that at least will mitigate all these problems 

in my fork I've optimized a lot of other calls to these methods, 
so some more PRs will follow

### DIFF
--- a/impl/src/main/java/com/sun/faces/util/Util.java
+++ b/impl/src/main/java/com/sun/faces/util/Util.java
@@ -63,6 +63,8 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.Spliterator;
 import java.util.Spliterators;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


BalusC:
Upon closer analysis, the majority of patterns are indeed literal strings not regexes. We can just drop the pattern cache altogether. Thank you for bringing this up! I'll take a look at 4.0.

BalusC:
Superseded by https://github.com/eclipse-ee4j/mojarra/pull/5940. Thank you for bringing this up! It's 3x faster.

PR #5938, opened 2026-08-10
Record restored attributes where the setters would have put them

BalusC:
### The bug

A standard component's pass-through attributes are emitted in the order of its `attributesThatAreSet` list. #5926 removed the per-render `Collections.sort(setAttributes)` and made the writers keep the list in natural order instead — but only updated one of them. `ComponentStateHelper.handleAttribute`, which re-records the list when full state saving restores a component without running its setters, still appends.

So the two writers disagree. A component built from the view lists its attributes sorted; the same component restored from full state lists them in whatever order its state entries happened to be replayed in, and renders its HTML attributes in that order.

With `partialStateSaving=false`, an `h:outputText` carrying `dir`, `lang`, `style` and `title`:

| | rendered |
|---|---|
| initial request | `dir lang role style title` |
| postback | `role title style lang dir` |

Same view, same component, different markup.

Whether it bites is a coin flip: `restoreState` replays the saved entries in the state map's iteration order, so if the `attributesThatAreSet` entry happens to come first, every later `handleAttribute` finds its name already present and appends nothing. That is also why it went unnoticed.

### The fix

Mirror the insertion the setters use, so both writers record a name in the same place.

`HtmlComponentUtils.insertionPoint` is package-private to `jakarta.faces.component.html`, which `jakarta.faces.component` cannot see, so the rule is repeated rather than shared — the same wall that already forces the `ATTRIBUTES_THAT_ARE_SET` constant to be duplicated across those packages. The javadoc on each says the two must agree.

### Coverage

`ComponentStateHelperTest.restoringFullStateRecordsTheAttributesInTheOrderTheSettersWouldHave` builds a component through its setters, saves, restores, and asserts the tracked order matches.

It moves the saved `attributesThatAreSet` entry to the end first, so the properties are replayed before it — otherwise the test passes on unfixed code, depending on where the key happens to hash. Without the fix it fails with `[style, lang, dir, title]` against the expected `[dir, lang, style, title]`.

Full impl suite green: 1290 tests.

### Scope

Not in any release — #5926 is unreleased on 4.0, 4.1 and master, so nothing shipped is affected. This changes only where the restore path records a name; emission order is unchanged (still sorted), and no rendered output changes for the partial-state-saving default.

The end-to-end symptom above was captured on the 5.0 branch, which has no `test/perf` equivalent here to reproduce it on; the restore-mimic is byte-identical between the branches.

### DIFF
--- a/impl/src/main/java/jakarta/faces/component/ComponentStateHelper.java
+++ b/impl/src/main/java/jakarta/faces/component/ComponentStateHelper.java
@@ -372,7 +372,7 @@ class ComponentStateHelper implements StateHelper, TransientStateHelper {
 
     /*
      * Because our renderers optimize we need to make sure that upon restore we mimic the handleAttribute of our standard
-     * generated HTML components setter methods.
+     * generated HTML components setter methods -- including where in the list it records a name, not only that it does.
      */
     private void handleAttribute(String name, Object value) {
         List<String> setAttributes = (List<String>) component.getAttributes().get(UIComponentBase.ATTRIBUTES_THAT_ARE_SET);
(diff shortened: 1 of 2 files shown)


BalusC:

PR #5939, opened 2026-08-10
Walk the form by index in f:validateWholeBean's placement check

BalusC:
An alternative to #5936, which set out to stop `f:validateWholeBean`'s placement check from copying the form's children. It does, and this takes the same goal further by removing the need for the utility altogether.

### What the check was doing

`misplacedComponentCheck` walks the form backwards looking for a rendered input placed after `f:validateWholeBean`. Per form, per render, it copied the children into a new `ArrayList`, streamed that, built every child's client id to compare against its own, and ended the walk by throwing `BreakException` — which does not override `fillInStackTrace`, and which is the *normal* exit, since reaching itself is how the walk terminates.

### What this does

Iterates the children by index, compares by identity, and reports reaching itself as a return value. `Util.reverse` and `BreakException` are left without callers and are deleted.

| per form, per render | before | #5936 | this |
|---|---|---|---|
| `ArrayList` copy per level | ✗ | removed | removed |
| `Stream` + lambda per level | ✗ | removed | removed |
| `BreakException` with stack trace on the normal exit | ✗ | ✗ | removed |
| `getClientId()` per child walked | ✗ | ✗ | removed |

I have **not** benchmarked the relative sizes of those four. My reasoning is that the copy is the cheapest of them — one array allocation plus N reference writes — while `fillInStackTrace` on every successful walk and a naming-container walk plus string build per child are each larger. That ordering is mechanism reasoning, not measurement; this path isn't in the perf suite, so measuring it would need a purpose-built harness.

### Correctness

Beyond the allocation, this avoids `ListIterator` entirely. `UIComponentBase.ChildrenListIterator.hasPrevious()` returns `index > 1`, where the `ListIterator` contract requires `index > 0`, so a backwards walk through it never yields the child at index 0 and a single-child container yields nothing at all. That is a long-standing latent bug — nothing in the tree walked children backwards through `listIterator` before — and it is orthogonal to this change, which simply doesn't rely on that iterator.

### Tests

Seven cases covering: an input after the validator, inputs before it, an input nested in a container after it, a single-child container after it, an input as the very first child, an unrendered input after it, and a validator outside a form.

They run without a `FacesContext`, which is only possible because the walk no longer builds client ids — a side benefit of the change worth noting on its own.

Full impl suite green: 1296 tests.

### Trade-offs

- This deletes `Util.reverse` rather than keeping it. It had exactly one caller. If a second one ever appears the utility has to come back, and #5936's lazy version would have served it; with no other callers today I think removing it is the better trade, but it is a trade.
- Deleting a `public static` method from `com.sun.faces.util.Util` is a bigger binary-compatibility change than altering its return type. That only matters if the package is treated as API rather than implementation — if it is, this PR is the worse offender of the two and should keep the method as a deprecated delegate instead. Happy to do that if maintainers prefer.

### DIFF
--- a/impl/src/main/java/com/sun/faces/ext/component/UIValidateWholeBean.java
+++ b/impl/src/main/java/com/sun/faces/ext/component/UIValidateWholeBean.java
@@ -16,7 +16,6 @@
 
 package com.sun.faces.ext.component;
 
-import static com.sun.faces.util.Util.reverse;
 import static jakarta.faces.validator.BeanValidator.EMPTY_VALIDATION_GROUPS_PATTERN;
 import static jakarta.faces.validator.BeanValidator.ENABLE_VALIDATE_WHOLE_BEAN_PARAM_NAME;
 import static jakarta.faces.validator.BeanValidator.VALIDATION_GROUPS_DELIMITER;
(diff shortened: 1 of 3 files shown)


BalusC:

PR #5940, opened 2026-08-10
Split on a literal delimiter instead of through a Pattern cache

BalusC:
Same ground as #5937, taken to the point where the pattern cache is no longer needed at all. That PR's five complaints are the right diagnosis; this addresses all of them by removing the thing they are complaints about rather than mitigating it.

### What was happening

`Util.split` took a regular expression and looked up its compiled `Pattern` in an `LRUMap` kept in the application map or the servlet context. Of the 26 call sites, **22 pass one plain character** — `,` `:` `/` ` ` `;` `=` — for which none of that machinery pays: the cache lookup and the `Matcher` cost more than splitting by index does.

Only four call sites pass a real regular expression. Each now holds its own compiled constant, which is what a constant regular expression should have been in the first place:

| where | pattern |
|---|---|
| `NavigationHandlerImpl` | `&amp;\|&` |
| `PartialViewContextImpl` | `[ \t]+` |
| `WebFacesConfigResourceProvider` | `,\|;` |
| `AnnotationScanner` | `\s+` |

With those moved out there is nothing left to cache, so the cache goes, and with it the capacity, the `synchronizedMap` every read went through, the lazy-initialisation race, and the application map argument threaded through every caller.

The one `Pattern` that is genuinely per-application — the iterator-nested client id check, built around the configurable separator character — keeps application scope but is stored under its own application map key rather than in a map of patterns.

`getSeparatorRegex()` becomes `getSeparatorPattern()` returning a `Pattern`, and `WebConfiguration` gains a `Pattern`-taking `getOptionValue` overload for the one caller that needs one.

### Measured

Splitting `"javax.faces,jakarta.faces,com.sun.faces,org.glassfish,foo,bar,baz"` on a comma, best of 8 rounds of 300k after warmup, same harness for both:

| | ns/op |
|---|--:|
| before | 420.9 |
| after | **127.1** |

### Against the five points in #5937

1. *pre-Java-7 problem for the majority of use cases* — the majority no longer touch a regular expression at all.
2. *calls retrieve FacesContext + ExternalContext + ApplicationMap just to split on a comma* — the argument is gone from the signature, so 22 call sites stopped fetching it.
3. *inefficient `synchronizedMap` over an `LRUMap`* — gone.
4. *only 15 patterns cached* — there is no cache to size.
5. *unsynchronized lazy initialisation creates several caches under load* — nothing is lazily initialised any more.

### Tests

`UtilSplitTest` covers 32 cases against `Pattern.compile(Pattern.quote(delimiter)).split(...)` as the oracle — quoting being exactly what makes a delimiter literal — across the delimiters the implementation uses, regex metacharacters as ordinary delimiters, multi-character delimiters, an absent delimiter, and the limit semantics that decide whether trailing empty strings survive.

Two defects it caught during development, both of which would otherwise have shipped:

- empty input returned `[]` rather than `[""]`. `Pattern.split` short-circuits with *"if no match was found, return this"* **before** discarding trailing empty strings; the first implementation applied that pass unconditionally.
- an empty delimiter looped forever, since `indexOf("", offset)` returns `offset`. Now rejected explicitly.

Full impl suite green: 1321 tests.

### The one thing to weigh

This changes `Util.split`'s **semantics**, not only its speed: a delimiter such as `"."` used to match any character and now matches a dot. All 26 in-tree call sites were audited and none relies on the old behaviour, but `com.sun.faces.util.Util` is reachable by anything that puts it on the classpath, and for such a caller the change is silent rather than a compile or link error.

If that is a concern, the mitigation is to name the new method `splitLiteral` and leave `split` deleted, so an outside caller gets a compile error instead of different output. Happy to do that if maintainers prefer.

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/ApplicationAssociate.java
+++ b/impl/src/main/java/com/sun/faces/application/ApplicationAssociate.java
@@ -258,7 +258,7 @@ public class ApplicationAssociate {
             resourceCache = new ResourceCache();
         }
 
-        resourceManager = new ResourceManager(applicationMap, resourceCache);
+        resourceManager = new ResourceManager(resourceCache);
         namedEventManager = new NamedEventManager();
         applicationStateInfo = new ApplicationStateInfo();
 
@@ -331,8 +331,7 @@ public class ApplicationAssociate {
 
         FacesContext ctx = FacesContext.getCurrentInstance();
 
-        Map<String, Object> appMap = ctx.getExternalContext().getApplicationMap();
-        compiler = createCompiler(appMap, webConfig);
+        compiler = createCompiler(webConfig);
         faceletFactory = createFaceletFactory(ctx, compiler, webConfig);
     }
 
@@ -649,10 +648,10 @@ public class ApplicationAssociate {
         return toReturn;
     }
 
-    protected Compiler createCompiler(Map<String, Object> appMap, WebConfiguration webConfig) {
+    protected Compiler createCompiler(WebConfiguration webConfig) {
         Compiler newCompiler = new SAXCompiler();
 
-        loadDecorators(appMap, newCompiler);
+        loadDecorators(newCompiler);
 
         // Skip params?
         newCompiler.setTrimmingComments(webConfig.isOptionEnabled(FaceletsSkipComments));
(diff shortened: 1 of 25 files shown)


BalusC:

PR #5941, opened 2026-08-10
Split around a character, and stop treating literal delimiters as regular expressions

BalusC:
A further step on the ground #5937 opened and #5940 built on.

#5937 pointed out that `Util.split` was pushing every call through a regular expression when almost none of them needed one. #5940 stopped the delimiter being a regular expression at all and deleted the pattern cache. This finishes the thought: every caller passes a single character, and the delimiter is taken literally anyway, so the parameter becomes the `char` it always was.

### What changes

```java
-    public static String[] split(String toSplit, String delimiter, int splitLimit)
+    public static String[] split(String toSplit, char delimiter, int splitLimit)
```

`indexOf(char)` replaces `indexOf(String)`, so a `String` comparison per delimiter occurrence goes away. More usefully, handing the method something regex-shaped is no longer expressible — `split(value, ".")` at least *looked* like it might be a pattern; `split(value, '.')` cannot be one.

The separator constants become characters along with it:

| | |
|---|---|
| `UrlBuilder` | all four separators, and `nextSeparatorChar` now holds the `char` its name has always claimed |
| `RenderKitUtils` | `CONTENT_TYPE_DELIMITER`, `CONTENT_TYPE_SUBTYPE_DELIMITER` |
| `WebConfiguration` | the separator `getOptionValue` takes |

`UrlBuilder` is package-private, so its `public` constants were never reachable outside `com.sun.faces.context` and the type change carries no API consequence.

The `Pattern`-taking `getOptionValue` overload stays, for the one parameter (`AnnotationScanPackages`) whose values a single character cannot separate.

The empty-delimiter guard #5940 needed is gone too — a `char` cannot be empty, so the case stopped existing rather than needing rejecting.

### Measured

Splitting `"javax.faces,jakarta.faces,com.sun.faces,org.glassfish,foo,bar,baz"` on a comma, best of 8 rounds of 400k after warmup:

| | ns/op |
|---|--:|
| before #5940 (cached `Pattern`) | 420.9 |
| after #5940 (`String` delimiter) | 127.1 |
| this (`char` delimiter) | **112.9** |

The 14 ns is not the point at this scale — a handful of splits per request. The API no longer being able to take a pattern is.

### Tests

`UtilSplitTest` retargeted to `char`, still asserting against `Pattern.compile(Pattern.quote(...)).split(...)` as the oracle across the delimiters the implementation uses, characters that would be metacharacters if this were a regular expression, an absent delimiter, and the limit semantics. The multi-character and empty-delimiter cases are dropped because the signature makes them unrepresentable.

Full impl suite green: 1328 tests.

### Second commit: the rest of the literal-regex sites

Once `Util.split` takes a character, the same question applies to everywhere else in `impl` that hands the regex engine something fixed.

- **Nine `String.split("x")` calls** plus one `split("\\.")` go through `Util.split(s, ch)`. Each was compiling a `Pattern` per call.
- **Eleven `String.replaceAll` calls** whose pattern is a single literal character become `String.replace`, which replaces the same occurrences and compiles nothing. (`DevTools` ×6 on `"<"`, the config resource providers ×5 on `" "`.)
- **`AjaxBehavior` and `ResetValuesHandler`** each held a `Pattern` for one space, purely to split the ajax execute and render lists — on the per-ajax-request path. `AjaxBehavior` is a spec class, so it uses `String.split(" ")`, whose own single-character path skips the regex engine and needs nothing from the implementation; that matters because on 5.0 the API is a separate module that cannot see `com.sun.faces`. `ResetValuesHandler` uses `Util.split`.

  Splitting `"@form :form:input1 :form:input2 :form:messages @this"` on a space: **183.5 → 103.8 ns/op**. A cached `Pattern` still builds a `Matcher` per call; `String.split` does not.

- **`AjaxBehaviorRenderer`** replaced a leading `@this` via `replaceFirst`. The enclosing branch has already established `@this` is at the front, so it now takes the remainder directly. That also stops the client id being interpreted as a *replacement* string, where `$` and `\` carry meaning — latent rather than live, since client ids don't normally contain either.

Deliberately left alone: everything that is genuinely a regular expression, and `ExternalContextImpl`'s `replaceFirst("http", "ws")`, which depends on matching only the *first* occurrence — a `replace` would also rewrite a later `http` in a query string, so it is not an equivalent swap.

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/ApplicationAssociate.java
+++ b/impl/src/main/java/com/sun/faces/application/ApplicationAssociate.java
@@ -665,7 +665,7 @@ public class ApplicationAssociate {
         String decoratorsParamValue = webConfig.getOptionValue(FaceletsDecorators);
 
         if (decoratorsParamValue != null) {
-            for (String decorator : split(decoratorsParamValue.trim(), ";")) {
+            for (String decorator : split(decoratorsParamValue.trim(), ';')) {
                 try {
                     newCompiler
                             .addTagDecorator((TagDecorator) forName(decorator).getDeclaredConstructor().newInstance());
--- a/impl/src/main/java/com/sun/faces/application/ApplicationStateInfo.java
+++ b/impl/src/main/java/com/sun/faces/application/ApplicationStateInfo.java
@@ -43,7 +43,7 @@ public class ApplicationStateInfo {
         partialStateSaving = config.isOptionEnabled(PartialStateSaving);
 
         if (partialStateSaving) {
-            String[] viewIds = config.getOptionValue(FullStateSavingViewIds, ",");
+            String[] viewIds = config.getOptionValue(FullStateSavingViewIds, ',');
             fullStateViewIds = new HashSet<>(viewIds.length, 1.0f);
             fullStateViewIds.addAll(asList(viewIds));
         }
--- a/impl/src/main/java/com/sun/faces/application/NavigationHandlerImpl.java
+++ b/impl/src/main/java/com/sun/faces/application/NavigationHandlerImpl.java
@@ -867,7 +867,7 @@ public class NavigationHandlerImpl extends ConfigurableNavigationHandler {
             if (queryString != null && queryString.length() > 0) {
                 String[] queryElements = QUERY_STRING_SEPARATOR.split(queryString);
                 for (int i = 0, len = queryElements.length; i < len; i++) {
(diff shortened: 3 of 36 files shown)


BalusC:
cc: @pizzi80

pizzi80:
:top: 

you're on fire ;) ... I'll create other PRs so we can improve Mojarra

BalusC:

PR #5942, opened 2026-08-11
LoadBundleHandler optimization + modernization + cosmetics 

pizzi80:
- optimized inner Map wrapper to better use the native API
- properly init collection
- removed superseeded inner class for Entry

notes:
 - cosmetics was also necessary for code readibilty and future improvements
 - Util.calculateMapCapacity is fundamental for future backports from my fork

### DIFF
--- a/impl/src/main/java/com/sun/faces/facelets/tag/faces/core/LoadBundleHandler.java
+++ b/impl/src/main/java/com/sun/faces/facelets/tag/faces/core/LoadBundleHandler.java
@@ -74,12 +74,13 @@ public final class LoadBundleHandler extends TagHandlerImpl {
             markDynamicTransientBuild(ctx);
         }
 
+        final UIViewRoot root = ComponentSupport.getViewRoot(ctx, parent);
+
         final ResourceBundle bundle;
         try {
-            final String name = basename.getValue(ctx);
-            final ClassLoader cl = Thread.currentThread().getContextClassLoader();
-            final UIViewRoot root = ComponentSupport.getViewRoot(ctx, parent);
-            final Locale locale = root != null && root.getLocale() != null ? root.getLocale() : Locale.getDefault();
+            String name = basename.getValue(ctx);
+            ClassLoader cl = Thread.currentThread().getContextClassLoader();
+            Locale locale = root != null && root.getLocale() != null ? root.getLocale() : Locale.getDefault();
 
             bundle = ResourceBundle.getBundle(name, locale, cl);
         }
@@ -141,7 +142,7 @@ public final class LoadBundleHandler extends TagHandlerImpl {
 
         @Override
         public boolean isEmpty() {
-            return keySet().isEmpty();
+            return false;
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


impl/src/main/java/com/sun/faces/facelets/tag/faces/core/LoadBundleHandler.java
```diff
@@ -49,45 +51,55 @@
...
 
-            public ResourceEntry(String key, String value) {
-                this.key = key;
-                this.value = value;
-            }
+        final ResourceBundle bundle;
+        try {
+            final String name = basename.getValue(ctx);
+            final ClassLoader cl = Thread.currentThread().getContextClassLoader();
+            final UIViewRoot root = ComponentSupport.getViewRoot(ctx, parent);
```
BalusC: This shouldn't be inside try. Keep it outside like original.

impl/src/main/java/com/sun/faces/facelets/tag/faces/core/LoadBundleHandler.java
```diff
@@ -114,43 +121,36 @@ public boolean containsValue(Object value) {
...
-                return "???" + key + "???";
+            } catch (MissingResourceException mre) {
+                return MISSING_RESOURCE_PLACEHOLDER + key + MISSING_RESOURCE_PLACEHOLDER;
             }
         }
 
         @Override
         public boolean isEmpty() {
-            return false;
+            return keySet().isEmpty();
```
BalusC: `return !bundle.getKeys().hasMoreElements();` is more efficient as `keySet()` builds a `new HashSet` every time.

pizzi80: looking at the underlying implementation (PropertyResourceBundle from Java 21) also bundle.getKeys() use the internal "lookup" Map 

```
public Enumeration<String> getKeys() {
    ResourceBundle parent = this.parent;
    return new ResourceBundleEnumeration(lookup.keySet(),
            (parent != null) ? parent.getKeys() : null);
}
```

eventually we could revert to `false`, it should be a safe approximation for our needs, what do you think?

impl/src/main/java/com/sun/faces/facelets/tag/faces/core/LoadBundleHandler.java
pizzi80: ok

impl/src/main/java/com/sun/faces/facelets/tag/faces/core/LoadBundleHandler.java
BalusC: getKeys() copies nothing, it **reuses** internal lookup map. The `lookup.keySet()` doesn't reference `ResourceBundle#keySet()` but plain `Map#keySet()`.

Here's impl of `ResourceBundle#keySet()`, it builds HashSet and copies it.
```
    public Set<String> keySet() {
        Set<String> keys = new HashSet<>();
        for (ResourceBundle rb = this; rb != null; rb = rb.parent) {
            keys.addAll(rb.handleKeySet());
        }
        return keys;
    }
```

pizzi80: fixed

BalusC:

PR #5943, opened 2026-08-11
EventInfo: constructorCache must be a Cache to avoid CME under load

pizzi80:
HashMap is not thread safe and throws ConcurrentModificationException

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/applicationimpl/events/EventInfo.java
+++ b/impl/src/main/java/com/sun/faces/application/applicationimpl/events/EventInfo.java
@@ -77,7 +77,9 @@ public class EventInfo {
     // ----------------------------------------------------- Private Methods
 
     private Constructor<?> getCachedConstructor(Class<?> source) {
-        if (eventConstructor != null) return eventConstructor;
+        if (eventConstructor != null) {
+            return eventConstructor;
+        }
         return constructorCache.get(source);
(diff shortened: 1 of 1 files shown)


pizzi80:
discovered in production ... didn't save the stack trace ...

impl/src/main/java/com/sun/faces/application/applicationimpl/events/EventInfo.java
```diff
@@ -78,53 +72,40 @@ public SystemEvent createSystemEvent(Object source) {
...
-                c = getEventConstructor(source);
-                if (c != null) {
-                    constructorMap.put(source, c);
-                }
-            }
-            return c;
-        }
-
+    private Constructor<?> getCachedConstructor(Class<?> source) {
+        if (eventConstructor != null) return eventConstructor;
```
BalusC: This is not the standard coding convention in Mojarra source code.

pizzi80: should be fixed

BalusC:
> discovered in production ... didn't save the stack trace ...

According to Git archeology this bug existed since JSF 2.0 already :o

pizzi80:
"archeology" is wonderful :D

BalusC:

PR #5946, opened 2026-08-11
ResourceHandlerImpl optimization + WebConfiguration standard way to get an int option 

pizzi80:


BalusC:
WebConfiguration context param is reworked 5.0 via https://github.com/eclipse-ee4j/mojarra/pull/5925. I don't think we should put effort in changing this in 4.x as that would only cause merge conflicts upstream.

pizzi80:
ouch ... I create that method many years ago and I also use it inside ViewScopeManager and ClientWindowManager (future PRs) ... I hate copy/pasted code

BalusC:
I'll cherrypick the ResourceHandlerImpl opt into 4.0. No worries.

BalusC:
Superseded by https://github.com/eclipse-ee4j/mojarra/pull/5949

PR #5947, opened 2026-08-11
Remove three fixed per-request costs from RESTORE_VIEW

BalusC:
Three fixed-cost removals in `RESTORE_VIEW`, found by profiling the phase on the `test/perf` suite against MyFaces 4.1.4-SNAPSHOT on Tomcat.

Framing first, because the suite's own numbers are misleading here: the 32-scenario lifecycle suite reports the GET-path `RESTORE_VIEW` cells at +46…+119% against MyFaces, but an isolated 4-scenario GET run (20 000 requests each) puts the real gap at **22.3 vs 17.2 µs, ~5 µs**, with `index` at outright parity. The rest is cross-scenario cache interleaving. On a 16-35 µs cell the percentages rank essentially at random — `flat-readonly` (+119%) and `flat-text` (+106%) score worse than the scenarios that look worst in the report. These changes target the ~5 µs that survives isolation, plus one per-component cost on facet-bearing tables.

### 1. Resolve the `@View` facelet bean once per view id

`DefaultFaceletFactory.getMetadataFacelet`/`getFacelet` built a `View.Literal` and a qualifier `HashSet` on every call to probe CDI for a programmatic facelet. Hashing that key runs `AnnotationLiteral`'s reflective member walk, so it is far from free: **13.9% of `RESTORE_VIEW` on an initial request — two thirds of what `getMetadataFacelet` costs in total, i.e. more than resolving and applying the facelet it is looking for.** It runs once per request for the metadata facelet and again per `buildView`, so a postback pays it twice.

`CdiUtils.getViewFacelet(BeanManager, String)` now keys the resolved `Bean` on the view id, next to the existing converter, validator and behavior bean caches and cleared with them on shutdown. `getReference` stays per call, so a request-scoped `@View` facelet keeps its scope.

### 2. Memoize the `FactoryFinder` instance per web application

Every `FactoryFinder.getFactory` call rebuilt a `FactoryFinderCacheKey`, and building one reads the marker and the init-time ClassLoader out of the application map before the key can even be hashed. A request makes several such calls per phase. JFR: **4.3% of `RESTORE_VIEW`, against 0.13% for MyFaces**, which resolves its factories once.

`CurrentThreadToServletContext.getFactoryFinder` now answers from the web application's own application map once the instance is known. That map belongs to exactly one web application, which is precisely what the key's marker exists to disambiguate, so it is a finer key than the one it short-circuits. The fast path is gated on the same "we have a FacesContext with a ServletContext" condition `FactoryFinderCacheKey` uses to pick `initFromApplicationMap`, so init-time calls without a FacesContext — the ones the key's ClassLoader scan exists for — still take the full path, and `removeFactoryFinder` clears the entry.

### 3. Walk the facets map's own snapshot when publishing after-view events

`publishAfterViewEvents` copied `getFacets().values()` into a list to stay safe against a listener relocating a facet mid-walk. But `FacetsMapValuesIterator` already iterates `entrySetIterator()`, which is itself a snapshot — so this was a copy of a copy, taken per facet-bearing component, per walk, and the build, the state restore and the post-restore-state pass each walk the whole tree. The sibling walks `disconnectFromView` and `updateInView` already iterate the values directly.

This is what makes a facet cost more than an ordinary child. Isolated, `flat-table-nofacets` is at **parity** (410 vs 413 µs); the whole table deficit is facets, at 260 ns/facet. JFR puts `entrySetIterator` at 3.11% of `RESTORE_VIEW` (~28 µs/request on a 200-facet table) with leaves that are pure copying.

Refuted along the way, so nobody re-chases them: the `facelets.FACET_NAME` put/remove protocol is byte-identical in MyFaces, and Mojarra's marker is already field-backed via `UIComponentBase.facetName`, so neither the protocol nor the attributes-map cost is the facet gap.

## Measurements

Tomcat, interleaved jar-swap A/B, both arms built and run back-to-back in one session, 4 reps per arm.

| | base | fix | Δ |
|---|--:|--:|--:|
| `RESTORE_VIEW` min, mean over all 40 scenarios | 25.1 µs | 19.3 µs | **−22%** |
| GET cells (avg): index / repeat-readonly / table-readonly | 22 / 26 / 25 µs | 16 / 18 / 20 µs | −27% / −32% / −18% |
| `repeat-build` / `table-build` (avg) | 118 / 104 µs | 90 / 92 µs | −24% / −11% |
| `flat-table-facets` (isolated, 6 reps, avg / min) | 524 / 164 µs | 488 / 118 µs | −7.0% / −28.1% |
| `flat-table-nofacets` (isolated, same run) | 402 / 90 µs | 394 / 90 µs | −2.1% / −0.6% |

The per-block minimum is the GET request that bootstraps each scenario's block, so it is the GET path directly; it improves on all 40 scenarios. The two `flat-table` rows are the evidence for change 3: the views differ only in whether the same 200 components hang as facets or as children, and only the facet one moves.

Non-`RESTORE_VIEW` phases are flat (mean +0.5% across 156 cells, nothing outside noise).

Honest caveat: change 3's own magnitude is a point estimate against a ±8% session noise floor — the facets-vs-nofacets differential and the JFR attribution agree on ~28 µs, but a single run does not certify it.

## Verification

- Impl unit suite green on this branch: 1332 tests, 0 failures.
- 4 new tests in `CdiLookupCountTest` covering lookup count, caching across calls, per-view-id independence, and `getReference` being invoked per call so scope semantics hold.
- Rendered-HTML equivalence over the whole `test/perf` corpus: **40/40 pages byte-identical**, 825 872 bytes on both arms (run on the 4.1 pair, same sources).
- Full Faces TCK green (run against the 5.0 port of these changes).

Refs #5856

### DIFF
--- a/impl/src/main/java/com/sun/faces/cdi/CdiUtils.java
+++ b/impl/src/main/java/com/sun/faces/cdi/CdiUtils.java
@@ -54,6 +54,7 @@ import jakarta.enterprise.inject.spi.BeforeBeanDiscovery;
 import jakarta.enterprise.inject.spi.CDI;
 import jakarta.enterprise.inject.spi.InjectionPoint;
 import jakarta.enterprise.util.TypeLiteral;
+import jakarta.faces.annotation.View;
 import jakarta.faces.component.behavior.Behavior;
 import jakarta.faces.component.behavior.FacesBehavior;
 import jakarta.faces.context.FacesContext;
@@ -63,6 +64,7 @@ import jakarta.faces.model.DataModel;
 import jakarta.faces.model.FacesDataModel;
 import jakarta.faces.validator.FacesValidator;
 import jakarta.faces.validator.Validator;
+import jakarta.faces.view.facelets.Facelet;
 
 /**
  * A static utility class for CDI.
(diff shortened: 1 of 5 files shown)


BalusC:

PR #5949, opened 2026-08-11
Parse the resource buffer size once at init

BalusC:
The `ResourceHandlerImpl` part of #5946, on 4.0 so it upmerges rather than conflicting, and without the `WebConfiguration` refactor.

`allocateByteBuffer()` re-read `com.sun.faces.resourceBufferSize` from the config and re-parsed it on every resource request. The parsed value is now held on the handler next to the other init-time settings (`initExclusions`, `initMaxAge`), and `handleHeaders` takes the `ExternalContext` the caller already resolved instead of digging it back out of the `FacesContext`.

## What this is and is not worth

The throughput effect is not measurable, and the PR should not be read as claiming one. `getOptionValue` is an `EnumMap` lookup and the parse is `Integer.parseInt("2048")` — together roughly 25ns, against a method that allocates a 2KB `ByteBuffer`, opens the resource stream and copies it to the response. That is a floor of tens of microseconds, so the saving is on the order of 0.02% of the request.

What it does fix is concrete: with a misconfigured `com.sun.faces.resourceBufferSize`, the old code logged `faces.application.resource.invalid_resource_buffer_size` at WARNING **on every resource request**, and read the option three more times to build the message. It now warns once, at startup. Parsing configuration once at init rather than per request is the right shape regardless of whether it shows up in a profile.

## Differences from #5946

- **`WebConfiguration` is untouched.** #5946 introduces a generic `getOptionIntValueOrDefault` helper; the parse stays local here, in an `initBufferSize()` matching the class's existing `initExclusions()` / `initMaxAge()`. This also keeps the localized, keyed log message — the generic helper in #5946 replaces it with a hardcoded English string, which would be a regression for non-English deployments.
- **`Util.isNotEmpty` is not pulled in.** It would touch a second file to swap a non-capturing lambda for a static method reference, which is a wash.
- The javadoc typo fixes and the `ctx` → `context` rename are left out to keep the diff to the change being made.

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/resource/ResourceHandlerImpl.java
+++ b/impl/src/main/java/com/sun/faces/application/resource/ResourceHandlerImpl.java
@@ -89,13 +89,14 @@ public class ResourceHandlerImpl extends ResourceHandler {
      */
     public static final String DEFAULT_CSP_POLICY = "script-src 'self' 'nonce-#{nonce}' 'strict-dynamic'";
 
-    ResourceManager manager;
+    private final ResourceManager manager;
     private String[] excludedExtensions;
     private long creationTime;
     private long maxAge;
-    private boolean cspEnabled;
+    private final boolean cspEnabled;
     private SecureRandom secureRandom;
-    private WebConfiguration webconfig;
+    private final WebConfiguration webconfig;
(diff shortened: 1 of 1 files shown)


BalusC:

PR #5948, opened 2026-08-11
ZipDirectoryEntryScanner optimizations

pizzi80:

PR #5950, opened 2026-08-11
ZipDirectoryEntryScanner: backport of #5948 + locale-prefixed library lookup fix

BalusC:
Two commits: the backport of #5948, and a fix for a defect in the code it touches.

## Backport of #5948 (1572215a9c)

Clean cherry-pick, no adaptation — the file was byte-identical on both branches.

`ZipDirectoryEntryScanner` held its scanned library names in a `Map<String, Boolean>` mapped to `Boolean.TRUE` and populated it with a `containsKey`/`put` pair. It now uses a `Set<String>` backed by `ConcurrentHashMap.newKeySet()` — same backing structure, honest type — and populates it in one lookup instead of two. The field is `private final`; it has no references outside the class.

No behaviour change. The constructor runs once per application and is dominated by inflating every JAR under `WEB-INF/lib` through a `ZipInputStream`, so the saved lookups are not expected to be measurable; the value here is the type stating what the data actually is.

## Locale-prefixed libraries were never found

The scan and the lookup disagreed about what a key looks like. For a resource at `META-INF/resources/de/mylib/foo.js` the scan recorded `de/mylib`, then swept the set:

```java
if (cur.contains("/")) {
    iter.remove();
}
```

while `libraryExists` builds exactly that shape whenever a locale prefix is given:

```java
String key = localePrefix != null ? localePrefix + '/' + libraryName : libraryName;
```

So every locale-prefixed key was guaranteed absent, and `libraryExists(name, prefix)` could only ever return `false`.

Reachable from `ResourceManager`, which passes the real locale prefix and, on failure, retries with a null prefix. The effect is therefore silent rather than fatal: with `com.sun.faces.enableMissingResourceLibraryDetection` enabled (opt-in, default off), a localized resource library packaged inside a JAR degrades to its non-localized variant instead of being found. That, plus the default-off flag, is why it has gone unreported.

The filter now runs at insert rather than as a sweep afterwards, which also disposes of the iterator loop:

```java
/**
 * A key as built by {@link #libraryExists(String, String)} is either {@code libraryName} or
 * {@code localePrefix + "/" + libraryName}. Anything deeper is a directory within a library, not a library.
 */
private static boolean isLibraryKey(String entryName) {
    return entryName.indexOf('/') == entryName.lastIndexOf('/');
}
```

One deliberate limitation: `<library>/<subdir>` is indistinguishable from `<locale>/<library>` at this level, so a library subdirectory stays in the set. That direction only weakens missing-library *detection*; the reverse drops libraries that genuinely exist, which is the bug being fixed here.

`ZipDirectoryEntryScannerTest` covers all three cases against an in-memory JAR. Without the fix, `localePrefixedLibraryIsFoundByNameAndPrefix` fails and the other two pass.

The fix goes on 4.0 so it upmerges; 4.1 and master currently carry the refactor from #5948 without it.

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/resource/ZipDirectoryEntryScanner.java
+++ b/impl/src/main/java/com/sun/faces/application/resource/ZipDirectoryEntryScanner.java
@@ -19,7 +19,6 @@ package com.sun.faces.application.resource;
 import static java.util.logging.Level.SEVERE;
 
 import java.io.IOException;
-import java.util.Iterator;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Logger;
@@ -58,7 +57,9 @@ class ZipDirectoryEntryScanner {
                                 int i = entryName.lastIndexOf('/');
                                 if (-1 != i) {
                                     entryName = entryName.substring(0, i);
-                                    resourceLibraries.add(entryName);
+                                    if (isLibraryKey(entryName)) {
+                                        resourceLibraries.add(entryName);
+                                    }
                                 }
                             }
                         }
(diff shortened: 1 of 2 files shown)


BalusC:

PR #5951, opened 2026-08-11
LoadBundleHandler optimization + modernization + cosmetics (#5942)

BalusC:
Backport of #5942 to 4.0.

- optimized inner `Map` wrapper to better use the native `ResourceBundle` API (`containsKey`, `keySet`, `getObject`)
- properly init collections via the new `Util.calculateMapCapacity`
- removed superseded inner `ResourceEntry` class in favour of `Map.entry`
- generified `ResourceBundleMap` to `Map<String, Object>`

`LoadBundleHandler` is byte-identical to 4.1 after this. Two behaviour changes ride along from 4.1 as merged: `isEmpty()` was hardcoded `false` and now actually inspects the bundle, and `containsKey(null)` returns `false` instead of throwing NPE. `values()` returns a `List` instead of a `Set`, so duplicate values are no longer collapsed.

Builds on JDK 11 and the impl unit suite is green (1332 tests).

### DIFF
--- a/impl/src/main/java/com/sun/faces/facelets/tag/faces/core/LoadBundleHandler.java
+++ b/impl/src/main/java/com/sun/faces/facelets/tag/faces/core/LoadBundleHandler.java
@@ -17,9 +17,10 @@
 package com.sun.faces.facelets.tag.faces.core;
 
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


BalusC:

PR #5953, opened 2026-08-12
LRUMap optimization: collision and resize free

pizzi80:


impl/src/main/java/com/sun/faces/util/LRUMap.java
```diff
@@ -24,20 +27,29 @@
...
     // ------------------------------------------------------------ Constructors
 
+    /**
+     * Create a {@link LRUMap} with the passed maxCapacity
+     */
     public LRUMap(int maxCapacity) {
-        super(maxCapacity, 1.0f, true);
+        // 1) to avoid collisions we should keep the default load factor, which is 0.75f
+        // 2) to avoid the map rehash when inserting the maxCapacity+1 element before removing the eldest one, we use maxCapacity+1
+        super(calculateMapCapacity(maxCapacity+1) , 0.75f, true);
```
BalusC, line 43: 0.75f can resize beyond max capacity for nothing. E.g. maxcapacity of 15 (default of views per session) creates 32 entries instead of 16 entries. Waste of memory space. Better keep 1.0f as the original did.

0.75f is only correct when a map has no max capacity.

impl/src/main/java/com/sun/faces/util/LRUMap.java
```diff
@@ -24,20 +27,29 @@
  */
 public class LRUMap<K, V> extends LinkedHashMap<K, V> {
 
+    @Serial
```
BalusC, line 30: Any particular reason? It's a `long` already. No need to add annotation.

impl/src/main/java/com/sun/faces/util/LRUMap.java
BalusC: Also, `+` needs space around in order to be consistent with existing code style.

BalusC:
The rehashing bugfix on a map which is sized to maxcapacity is incompatible with `calculateMapCapacity`. The `@Serial` is unnecessary and out of scope.

This should have been the whole fix `super(maxCapacity + 1, 1.0f, true);`. I'll create PR for 4.0 (which also allows easier upmerging without need for a backport PR).

pizzi80:
not sure i've understood ...

0.75f is the default load factor ... use 1.0f is wrong

if you use 0.75f you need to properly init the Map size

"modern" jvm have statics initializers exactly to solve this confusion

pizzi80:
if you want maximum performance + no internal resize my version is correct 
is the same logic fixed in Java 21+  `LinkedHashMap.newLinkedHashMap()`

BalusC:
This is a bounded LRU map not an unbounded map. See new PR.

PR #5954, opened 2026-08-12
Skip view map for CSP nonce on stateless views (#5952)

### LINKED ISSUE #5952: Warnings about unsupported `@ViewScoped` when using CSP on a stateless view
When rendering a stateless view when CSP is enabled, the following warning is logged

```
@ViewScoped beans are not supported on stateless views
```

As far as I can tell, the warning is caused by the initialization of the view map when resolving the current CSP nonce for inclusion in the response header. See [here](https://github.com/eclipse-ee4j/mojarra/blob/072fa7667b9ef10606b8e0e8a753a3534bd99940/impl/src/main/java/org/glassfish/mojarra/application/resource/ResourceHandlerImpl.java#L405).

I didn't notice any problems other than the warning. Is it safe to ignore the warning? Can `ResourceHandlerImpl::getCurrentNonce` be changed to not create a view map when the view is stateless?

BalusC:
When CSP nonce support is enabled, `ResourceHandlerImpl#getCurrentNonce` called `getViewMap(true)` unconditionally, also on a transient view root. That publishes a `PostConstructViewMapEvent`, and `ViewScopeManager` then logs "@ViewScoped beans are not supported on stateless views" in Development stage, and — in any stage — acquires an HTTP session and registers a view map in `ACTIVE_VIEW_MAPS` which is unreachable on a stateless view, so a stateless view with CSP enabled acquired a session on every render.

The view map is transient state and cannot carry the nonce across requests on a stateless view anyway, so it is now resolved only for a non-transient view root. The nonce is still cached in the request map.

Integration tests are added to `test/csp`: a `<f:view transient="true">` view asserting the CSP header carries a nonce while no session is acquired, and one asserting behavior scripts still get a nonce.

Closes #5952

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/resource/ResourceHandlerImpl.java
+++ b/impl/src/main/java/com/sun/faces/application/resource/ResourceHandlerImpl.java
@@ -455,9 +455,10 @@ public class ResourceHandlerImpl extends ResourceHandler {
             var nonce = (String) requestMap.get(CURRENT_NONCE);
 
             if (nonce == null) {
-                var viewMap = context.getViewRoot().getViewMap(true);
+                var viewRoot = context.getViewRoot();
+                var viewMap = viewRoot.isTransient() ? null : viewRoot.getViewMap(true);
 
-                if (context.getPartialViewContext().isPartialRequest()) {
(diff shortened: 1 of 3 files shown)


BalusC:

PR #5957, opened 2026-08-13
Remove full state saving

BalusC:
Implementation half of jakartaee/faces#2226, which removes full state saving from the API in jakartaee/faces#2232. With `jakarta.faces.PARTIAL_STATE_SAVING` and `jakarta.faces.FULL_STATE_SAVING_VIEW_IDS` gone, there is no way to select it and nothing left behind the switch.

`FaceletPartialStateManagementStrategy` is renamed to `FaceletStateManagementStrategy` now that it is the only strategy, and `FaceletViewHandlingStrategy#getStateManagementStrategy` returns it unconditionally.

`StateContext#isPartialStateSaving` and the per-view decision it cached are gone. That method also detected a changing `UIViewRoot` and dropped the `AddRemoveListener` so the new view gets its own; since six call sites triggered that only as a side effect of asking about state saving, the detection moves to `startTrackViewModifications`, which is where the listener is created and subscribed. This is the only behavioral change in the diff.

What that leaves without a caller: `ApplicationStateInfo` (and its field, initialization and getter on `ApplicationAssociate`), `FaceletFullStateManagementStrategy`, both `FacesContextParam` constants, the no-op `AddRemoveListener` pair, the `RIConstants.VIEWID_KEY_NAME` stash in `ViewMetadataImpl` whose only reader was `isPartialStateSaving`, and the `StateContext`-taking overload of `ComponentSupport#suppressViewModificationEvents`, which existed to avoid a lookup that no longer happens.

`ComponentSupport#findChildByTagIdFullStateSaving` is renamed to `findChildByTagIdScan` — it is the plain scan next to the indexed and deep variants, and never had anything to do with the mode.

`CONTEXT-PARAMS.md` drops both rows, which `ContextParamsMdTest` enforces, and `ApplicationStateInfoTest` goes with the class it covered. Comments that explained code by naming the removed mode now state what the code does.

## Verification

`mvn clean install` green: 1308 tests, no failures. Full Jakarta Faces TCK green in 4:33 with this build overlaid, against the API from jakartaee/faces#2232.


BalusC:

PR #5959, opened 2026-08-13
Don't track view root postAddToView additions as dynamic components

### LINKED ISSUE #5958: Component added from a view-root postAddToView listener is needlessly saved in the state
A child added by a listener attached to the `UIViewRoot`'s `postAddToView` event is recorded as a dynamic component action, so it is carried in the saved state on every request — even though `buildView` reproduces it on every request by re-running the same listener. A listener attached to any other component does not have this problem.

The cost is threefold: the added subtree is serialized through `StateHolderSaver` as complete state rather than a delta, its components get `clearInitialState()` so they never participate in partial state saving, and the view's dynamic-action list is permanently non-empty, which disables the render-time skip in `canSkipTransientBuildRefresh` and forces a redundant facelet re-apply on every render for the rest of the view's life.

There is also a correctness consequence. Because the ADD is replayed from the state on the next request, a listener that adds conditionally sees the component reappear on a request where the condition no longer holds, instead of it simply not being built.

**Cause.** In `FaceletViewHandlingStrategy.buildView` the order is `startTrackViewModifications(ctx, view)` → `publishEvent(PostAddToViewEvent.class, UIViewRoot.class, view)` → `markInitialState(ctx, view)`. Tracking is therefore live when the view-root event fires. `AddRemoveListener.isListenerForSource` excludes `UIViewRoot`, but that only excludes the view root's own event: the child the listener adds publishes its own `PostAddToViewEvent` with itself as source, which reaches `handleAdd` and is recorded. Listeners on other components fire during `facelet.apply(...)`, before the tracking listener exists, and are correctly ignored.

**Fix.** Move `startTrackViewModifications(ctx, view)` after `markInitialState(ctx, view)`. The view-root listener's additions are then made while tracking is off and are marked with the rest of the tree, like every other build-time add; the tracking listener is still installed inside `buildView`, so genuine post-build mutation is tracked as before.

**How this surfaced.** While removing full state saving (jakartaee/faces#2226) the question came up whether components added at view build time (`postAddToView`) need to be saved at all, since the build reproduces them on every request. Reading `StateContext` suggested `postAddToView` adds are already free, but that the view-root ordering was an exception. A probe confirmed it: a page whose listener is attached to the view root reported one dynamic action on the initial render, the identical page with the listener on the container reported zero.

**Affected branches.** Present with the same ordering on `4.0` and `4.1` (`com.sun.faces.application.view.FaceletViewHandlingStrategy`, lines 385/389/390). Proposed for `master` only; 4.x apps can attach the listener to the target container instead.

**Verification.** Mojarra unit tests 1308 pass; full Jakarta Faces TCK green, 6373 integration tests, zero failures.

BalusC:
Fixes #5958. `buildView` started view modification tracking before publishing the view root's `PostAddToViewEvent`, so a child added by a listener on that event was recorded as a dynamic action and carried in the saved state on every request — although the next build re-runs the same listener and produces it again. Tracking now starts after `markInitialState`, which leaves those additions to the build like any other build-time add. Listeners on other components were already free, since they fire during the facelet apply, before the tracking listener exists.

Beyond the wasted state, the old order had a correctness consequence: the recorded ADD was replayed on a request where the listener would not have added anything, so a conditionally added component reappeared after its condition went away.

The new `test/impl` module is for integration tests of behavior the TCK cannot assert because it is implementation specific — here, what ends up in the saved state. It mirrors `perf`'s setup (war, `base` for the Arquillian config, Mojarra as `provided` so a probe can read internals) and its first test renders `StateContext.getDynamicActions().size()` and requires zero for both listener placements. That test fails on master with `expected: <0> but was: <1>` for the view-root page and passes for the container page, which is how the defect was pinned down.

Stacked on #5957 and targeting its branch; retarget to master once that merges.

Verification: unit tests 1308 pass, `test/impl` green, and the full Jakarta Faces TCK green with this build overlaid — 6373 integration tests, zero failures, including the new vendor-neutral test in jakartaee/faces#2233.

### DIFF
--- a/impl/src/main/java/org/glassfish/mojarra/application/view/FaceletViewHandlingStrategy.java
+++ b/impl/src/main/java/org/glassfish/mojarra/application/view/FaceletViewHandlingStrategy.java
@@ -367,13 +367,12 @@ public class FaceletViewHandlingStrategy extends ViewHandlingStrategy {
                 htmlDoctype.setSystem(doctype.getSystem());
                 view.setDoctype(htmlDoctype);
             }
-
-            startTrackViewModifications(ctx, view);
         } finally {
             ctx.getAttributes().remove(IS_BUILDING_INITIAL_STATE);
         }
         ctx.getApplication().publishEvent(ctx, PostAddToViewEvent.class, UIViewRoot.class, view);
         markInitialState(ctx, view);
+        startTrackViewModifications(ctx, view);
 
         setViewPopulated(ctx, view);
     }
--- a/test/impl/pom.xml
+++ b/test/impl/pom.xml
@@ -1,0 +1,56 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2026 Contributors to Eclipse Foundation.
+
+    This program and the accompanying materials are made available under the
+    terms of the Eclipse Public License v. 2.0, which is available at
+    http://www.eclipse.org/legal/epl-2.0.
+
+    This Source Code may also be made available under the following Secondary
+    Licenses when the conditions for such availability set forth in the
+    Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+    version 2 with the GNU Classpath Exception, which is available at
+    https://www.gnu.org/software/classpath/license.html.
+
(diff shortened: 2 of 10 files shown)


BalusC:

PR #5960, opened 2026-08-14
Skip the render-time re-apply when the build's decisions still hold

BalusC:
A view holding any build-time-dynamic tag re-applies its whole facelet before rendering, because the skip was denied on the mere presence of such a tag. One `c:if` governing 0.25% of the tree therefore re-applied 100% of it, at ~75% of a full initial build.

Such a handler now records what it decided, and the skip goes ahead when every recorded decision still holds: the same handlers then take the same branches, create the same components and apply the same values, so re-applying would reproduce the identical view. That is N evaluations instead of a walk over N components. A decision that yields another value, or cannot be evaluated at all (as one that only holds under the branch that guarded it), falls back to the full re-apply.

A decision is either an expression that must still evaluate to its recorded value, or a supplier that must still yield it. The latter answers what an expression cannot: whether the value a handler applied is still the one in effect. Under partial state saving the saved state is applied over the tree the restore-time build produced, so a `UIViewRoot` property that build set is replaced by the one the previous request saved, and the re-apply is what asserts it again. The resource library contracts need the same for another reason: restoring a view recalculates them from the configured mappings once the build that applied them has finished.

A decision is re-evaluated under the components it was recorded under. An expression reaching `cc` or `component` resolves them from the component stack rather than from a variable, and that stack is empty outside a build: `cc` would resolve to null, and an operator over null yields a value of its own rather than failing, which a recorded value can match.

| tag | decision |
| - | - |
| `c:if`, `c:choose` | the test they evaluated |
| `f:view` | each non-literal attribute twice: the value it evaluated to and the value it applied |
| `f:loadBundle` | basename and var, and the view locale even for a literal basename, which a changed locale resolves a different bundle from |
| `c:forEach` | its range and, element-wise, its item sequence, for a positionally indexable source |
| `ui:include`, `ui:composition`, `ui:decorate` | the path they applied; the facelet at that path recorded its own decisions into the same list while it was being applied |
| `c:set` | the value it wrote and that value still being the one at the location it wrote it to; its var-only form maps the unevaluated expression and needs no decision at all |

The attributes naming what a handler binds to — a variable, a scope, a property, a facet — are recorded like any other, the tag libraries declaring them as expressions rather than as literals.

## Handlers this implementation does not know

Such a handler was granted the skip silently, since nothing marked the build — a third-party tag that branches on its attributes at build time has been skipped since 4.0.19. The compiler now classifies every tag handler it creates, once per facelet compile rather than per build, and a facelet holding one whose contribution is unaccounted for is built as unreproducible. A handler is accounted for when:

1. it attaches a converter, validator or behavior to its parent — re-executable during a render rather than defining what is built, and acting only while the parent is new to the tree (this is what keeps `o:converter` and `o:validator` on the fast path despite taking over `apply`);
2. it leaves `apply` to the delegate of `ComponentHandler` and its siblings — nearly every component tag of every component library;
3. it is one of this implementation's own, each of which `BuildTimeDecisionsTest` holds to having been audited;
4. or its tag carries no expression at all, which leaves nothing the lifecycle could have changed since the last build — both builds are performed within the same request, so everything the handler reads from that request is the same for both.

Nothing is required of a third-party library: the safe verdict is the default, and no mojarra-specific annotation or dependency is involved.

## What still denies the skip

`c:catch`, whose body having thrown is knowable only by applying it, and a `c:forEach` over a Map or other non-indexable Collection, which iterates over a snapshotted entry with no positional index to compare or read live.

`f:attribute` and `f:passThroughAttribute` need no decision, being applied only while their parent is new to the tree.

`RIConstants.DYNAMIC_TRANSIENT_BUILD` is dropped — the record answers what the flag was consulted for, and `c:forEach` compares the record's size around its body where it previously cleared and re-set that flag.

## Measurements

Second `buildView` of the same request, GlassFish, ~815 components, 1500 iterations:

| page | before | after |
| - | - | - |
| no build-time-dynamic tag | 832 ns | ~1 us (existing skip) |
| one `c:if`, 4 levels deep | 489,955 ns | 7,843 ns |
| same with `c:choose` | 457,426 ns | 4,367 ns |

## Testing

The 5.0 equivalent of this change (identical logic, `org.glassfish.mojarra` packages) is green against the full TCK. That includes the guards which exercise a decision changing mid-request — `Issue4086IT` (`f:view locale` switched across five postbacks), `LoadBundleIT`, `Issue3982IT` and `Issue1849IT` (`c:if` flipped by an action), `Issue2631IT`/`Issue2896IT` (`c:forEach` mutated on postback) — plus `IfPostbackIT`, `ChoosePostbackIT` and `CompositeIfPostbackIT` in jakartaee/faces#2236, the last of which fails without the component context restored around a re-evaluated decision.

On this branch: 1342 unit tests green at source level 11, covering the record itself (a decision holding, changing, failing to evaluate, being overridden by an unreproducible mark, and being discarded by a reset) and the classification of a handler from another tag library.

Part of #5856

### DIFF
--- a/impl/src/main/java/com/sun/faces/RIConstants.java
+++ b/impl/src/main/java/com/sun/faces/RIConstants.java
@@ -41,14 +41,6 @@ public class RIConstants {
     public static final String SAVED_STATE = FACES_PREFIX + "savedState";
 
     /**
-     * Request-scoped flag set during a view build when a build-time-dynamic handler (a JSTL conditional/iteration or a
-     * dynamic ui:include/ui:decorate/ui:composition) participates, marking the view as one whose facelet must be
-     * re-applied on every (re)build. Read by {@code FaceletViewHandlingStrategy.buildView} to decide whether the
-     * redundant render-time re-apply may be skipped (see {@code refreshTransientBuildOnPSS}).
-     */
-    public static final String DYNAMIC_TRANSIENT_BUILD = FACES_PREFIX + "dynamicTransientBuild";
-
-    /**
      * Request-scoped flag recording whether the render-time {@code buildView} re-applied the facelet ({@code TRUE}) or
      * skipped the re-apply for an already-populated static view ({@code FALSE}, see {@code refreshTransientBuildOnPSS}).
      * Since {@code buildView} runs immediately before {@code renderView}, this reflects whether the tree the state
--- a/impl/src/main/java/com/sun/faces/application/view/FaceletViewHandlingStrategy.java
+++ b/impl/src/main/java/com/sun/faces/application/view/FaceletViewHandlingStrategy.java
@@ -17,7 +17,6 @@
 package com.sun.faces.application.view;
 
 import static com.sun.faces.RIConstants.DYNAMIC_COMPONENT;
-import static com.sun.faces.RIConstants.DYNAMIC_TRANSIENT_BUILD;
 import static com.sun.faces.RIConstants.VIEW_REBUILT_AT_RENDER;
 import static com.sun.faces.RIConstants.FACELETS_ENCODING_KEY;
 import static com.sun.faces.RIConstants.FLOW_DEFINITION_ID_SUFFIX;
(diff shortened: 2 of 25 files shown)


BalusC:

PR #5963, opened 2026-08-14
Reject a value expression in ui:repeat var and varStatus

### LINKED ISSUE #5962: ui:repeat var and varStatus silently ignore a value expression
`<ui:repeat var="#{bean.name}">` is accepted and then does nothing: the loop variable is never set, so the body cannot resolve it.

`RepeatHandler` uses the standard component meta ruleset, so a non-literal attribute is applied through `ComponentRule.ValueExpressionMetadata`, which calls `setValueExpression("var", …)`. `UIRepeat` keeps `var` and `varStatus` in plain fields with plain getters, so the stored expression is never consulted. A literal works only because the property-aware attributes map routes it to `setVar`.

Probed against 4.0.23-SNAPSHOT:

```java
UIRepeat repeat = new UIRepeat();
repeat.setValueExpression("var", expression);
repeat.getVar(); // null
```

Every other attribute of this kind rejects an expression instead: `UIData.setValueExpression` throws `IllegalArgumentException` for `var` and `rowIndex`, and `UISelectItems`, `UIImportConstants` and `UIWebsocket` do the same for `var` and for `channel`/`scope`. `h:dataTable var` and `ui:repeat var` have the same semantics and should behave the same way.

Expected: `UIRepeat.setValueExpression` throws `IllegalArgumentException` for `var` and `varStatus`, so a page using an expression there fails where it is written rather than rendering an empty body.

For 4.0, to be upmerged into 4.1 and master.

The documentation side is jakartaee/faces#2238: the tag library declares both attributes with a `<type>`, which documents them as value expressions, unlike every comparable attribute.

BalusC:
`<ui:repeat var="#{bean.name}">` was accepted and then did nothing: the loop variable was never set, so the body could resolve nothing.

`RepeatHandler` uses the standard component meta ruleset, so a non-literal attribute is applied through `ComponentRule.ValueExpressionMetadata`, which calls `setValueExpression("var", …)`. `UIRepeat` keeps `var` and `varStatus` in plain fields with plain getters, so the stored expression was never consulted. A literal works only because the property-aware attributes map routes it to `setVar`.

`UIRepeat` now rejects both with `IllegalArgumentException`, exactly as `UIData` does for `var` and `rowIndex`, and `UISelectItems`, `UIImportConstants` and `UIWebsocket` do for theirs. The body of the tag resolves the name as it is written, so nothing can name it that is not known when the page is.

`value`, `offset`, `size`, `begin`, `end` and `step` are untouched and still take expressions.

The tag library side of the same defect is jakartaee/faces#2239, which stops documenting these attributes as value expressions.

## Testing

`UIRepeatTest` asserts both rejections. The full `impl` unit suite is green (1339). No page in the TCK passes an expression to either attribute — 164 `ui:repeat` usages, none of them dynamic in `var` or `varStatus` — so nothing existing trips the new throw.

Closes #5962

### DIFF
--- a/impl/src/main/java/com/sun/faces/facelets/component/UIRepeat.java
+++ b/impl/src/main/java/com/sun/faces/facelets/component/UIRepeat.java
@@ -183,6 +183,21 @@ public class UIRepeat extends UINamingContainer {
 
     }
 
+    /**
+     * A value expression cannot name the variable this component exports, since the body of the tag resolves that name
+     * as it is written. One is therefore rejected rather than accepted and then ignored.
+     *
+     * @throws IllegalArgumentException when the given name is {@code var} or {@code varStatus}
+     */
+    @Override
+    public void setValueExpression(String name, ValueExpression binding) {
+        if ("var".equals(name) || "varStatus".equals(name)) {
+            throw new IllegalArgumentException(name);
+        }
+
+        super.setValueExpression(name, binding);
+    }
+
     public String getVar() {
         return var;
     }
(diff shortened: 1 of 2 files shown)


BalusC:

PR #5964, opened 2026-08-14
Report where the restore rebuilds another view than was rendered

### LINKED ISSUE #5961: PSS does not fully restore view at end of RESTORE_VIEW as FSS does
**Describe the bug**

PSS does not fully restore view at end of RESTORE_VIEW as FSS does

**Expected behavior**

At the end of RESTORE_VIEW the view build by PSS should be identical to FSS

**Additional context**

FSS — tree fully materialized from state, no Facelet re‑execution
* FaceletViewHandlingStrategy.restoreView (line ~213): FSS takes no buildView branch — it falls straight through to super.restoreView (line ~268).
* FaceletFullStateManagementStrategy.restoreView (line ~520) → restoreTree (def ~476, called ~535): rebuilds the whole structure from serialized TreeNodes via newInstance, then restoreComponentState (~551) applies full state.
* saveView (~649): captureChild (def ~126, called ~676) serializes the structure; saveComponentState (~580) saves full per‑component state.
  
PSS — structure rebuilt by re‑running the template, finalized at render
* FaceletViewHandlingStrategy.restoreView, PSS branch (lines ~241–266): calls vdl.buildView(...) at line ~262 — i.e. it re‑executes the Facelet against the current(pre‑Update‑Model) model to recreate structure.
* FaceletPartialStateManagementStrategy.restoreView (~318): overlays only deltas via visitTree → restoreState (~366), and crucially only when a live target exists (guard at ~358–366); then restoreDynamicActions (~376).
* FaceletViewHandlingStrategy.renderView (~410) → buildView (~427): the second apply at Render Response.
buildView skip guard canSkipTransientBuildRefresh (~397) + VIEW_REBUILT_AT_RENDER(~308/314): the second apply is skipped only for static views and forced when DYNAMIC_TRANSIENT_BUILD/dynamic actions are present.

BalusC:
Proposal (1) of #5961, taken on its own: a Development mode report, no behavior change, no agreement needed about the semantics that (2) turns on.

## What it reports

A postback rebuilds its view from the markup, so a build time condition evaluating to another value than it did while the response was rendered produces another view than the one that was submitted. Under `ProjectStage.Development` the tag each client id was built from is saved with the state, and the restore compares it against the view it rebuilt, logging two things at `WARNING`:

- **Not rebuilt** — client ids the rendered view held which the rebuilt one does not, so their state is restored into nothing. The ones for which a request parameter of that exact name was submitted are counted separately: those are the silently dropped values.
- **Rebuilt from another tag** — client ids another tag now occupies than the one that rendered them, so the state saved by one tag is restored into the component of another.

Both name what resolves it: the value the condition depends on has to survive the postback, so hold it in a `@ViewScoped` bean or recompute it from the relevant request parameters in the `@PostConstruct` of the request scoped one.

## Why the tag, and not the state map or the component type

The obvious implementation — diff `state.keySet()` against the client ids the restore walk visited — does not catch the reproducer in the issue. `saveView` only puts an entry in the map `if (stateObj != null)`, so an `h:inputText` created by the render which produced the response, and never touched after `markInitialState`, has a null delta and **no entry at all**. There is no orphan key to find. The comparison therefore has to be against what the previous render actually built, which is why the saved side is collected in `saveView`'s existing walk rather than derived from the state map.

The component type is not a sufficient discriminator either. Two tags may carry the same component id as long as only one of them is ever built, which is exactly what the branches of a `c:choose` and sibling `c:if` tags do:

```xhtml
<c:if test="#{bean.a}"><h:inputText id="value" value="#{bean.x}" /></c:if>
<c:if test="#{not bean.a}"><h:inputText id="value" value="#{bean.y}" /></c:if>
```

Same id, same class, different tag. When the decision flips, the first input's `submittedValue`/`localValue`/`valid` are restored into the second, which binds to a different property. Nothing throws, because the state shape is identical. That is the quietest variant of the family, and only the tag tells the two apart. Facelets already stamps it as `ComponentSupport.MARK_CREATED`, derived from the tag's compile time position, so it is stable per tag across rebuilds and costs nothing to collect. A component no facelet built — the view root, anything added programmatically — falls back to its type, which the restore reproduces as well.

The both-tests-true variant needs nothing: that produces a genuine duplicate id and `Util.checkIdUniqueness` already errors on it.

## Scope

This reports the `c:if` / `c:choose` / variable `ui:include` class only. A flipped `rendered` leaves the component in the tree and is re-evaluated at decode identically under both state saving strategies, so it stays quiet there — the two are different axes, as established in the issue thread.

## No effect outside Development

`saveView` collects nothing, the state map gains no entry, and the restore compares nothing. The one non-gated addition is `isViewRootOnlyState` excluding the new key from its entry count, which is an `O(1)` map lookup per restore and keeps its fast path intact in Development. `getClientId()` is still only called when `saveState()` returned non-null, as before.

The state format gains a key in Development only; state saved before this change simply lacks it and the check no-ops.

## Reproducer

One page exercises both reports in a single postback, with a `@Named @RequestScoped` bean whose `shown` flag `show()` sets:

```xhtml
<h:form id="form">
    <h:commandButton id="show" value="show" action="#{issue5961.show}" />

    <c:if test="#{issue5961.shown}">
        <h:inputText id="guarded" value="#{issue5961.guarded}" />
    </c:if>

    <c:if test="#{issue5961.shown}">
        <h:inputText id="shared" value="#{issue5961.whenShown}" />
    </c:if>
    <c:if test="#{not issue5961.shown}">
        <h:inputText id="shared" value="#{issue5961.whenHidden}" />
    </c:if>

    <h:commandButton id="submit" value="submit" />
</h:form>
```

GET, click `show`, fill out both inputs, click `submit`. The flag is true during the render which puts them in the response and false again during the restore of the postback which submits them. Before this change `form:guarded` is dropped silently and what was typed into `form:shared` lands in `whenHidden` rather than `whenShown`, with nothing logged. After it, that postback reports `form:guarded` as not rebuilt and as carrying a submitted value, and `form:shared` as built from another tag.

## Tests

`FaceletPartialStateManagementStrategyTest` covers the two comparisons in both directions, the tag identity and its fallback, the report truncation boundary, and `isViewRootOnlyState` for each bookkeeping key in isolation and together. Both reports are `MessageFormat` patterns, so a further test pins that neither holds a lone apostrophe — one would quote the text after it and silently swallow what follows rather than fail.

Closes #5961 partially; (2) remains open and targets 5.0.

### DIFF
--- a/impl/src/main/java/com/sun/faces/RIConstants.java
+++ b/impl/src/main/java/com/sun/faces/RIConstants.java
@@ -87,6 +87,14 @@ public class RIConstants {
     public static final String DYNAMIC_ACTIONS = FACES_PREFIX + "DynamicActions";
 
     /**
+     * Marker used when saving, under {@link jakarta.faces.application.ProjectStage#Development} only, the tag each
+     * client id was built from, so that the build which restores the view can be told where it produced another
+     * component than the one that was rendered. It holds every component rather than only those carrying a delta: a
+     * component the restore fails to produce is the one to report, and it need never have held state of its own.
+     */
+    public static final String RENDERED_TAGS = FACES_PREFIX + "RenderedTags";
+
+    /**
      * Marker attached to a component that has dynamic children.
      */
     public static final String DYNAMIC_CHILD_COUNT = FACES_PREFIX + "DynamicChildCount";
--- a/impl/src/main/java/com/sun/faces/application/view/FaceletPartialStateManagementStrategy.java
+++ b/impl/src/main/java/com/sun/faces/application/view/FaceletPartialStateManagementStrategy.java
@@ -18,6 +18,7 @@ package com.sun.faces.application.view;
 
 import static com.sun.faces.RIConstants.DYNAMIC_ACTIONS;
 import static com.sun.faces.RIConstants.DYNAMIC_COMPONENT;
+import static com.sun.faces.RIConstants.RENDERED_TAGS;
 import static com.sun.faces.RIConstants.VIEW_REBUILT_AT_RENDER;
 import static com.sun.faces.util.ComponentStruct.ADD;
 import static com.sun.faces.util.ComponentStruct.REMOVE;
@@ -37,6 +38,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import com.sun.faces.context.StateContext;
+import com.sun.faces.facelets.tag.faces.ComponentSupport;
 import com.sun.faces.renderkit.RenderKitUtils;
 import com.sun.faces.util.ComponentStruct;
 import com.sun.faces.util.FacesLogger;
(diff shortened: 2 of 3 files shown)


BalusC:

PR #5965, opened 2026-08-15
Replay the build time decisions of the rendered view

BalusC:
Proposal (2) of #5961, opt in: the build which restores a postback replays the build time decisions the response was rendered under instead of evaluating them against the current state of the model, so a value submitted for a component one of them guarded is decoded rather than dropped.

## What it does

A postback rebuilds its view from the markup, so a build time decision evaluating to another value than it did while the response was rendered produces another view than the one that was submitted. The state of a component the rebuild does not produce is restored into nothing, and a value submitted for it is decoded by nothing. #5964 reports that under Development; this fixes it.

Under `com.sun.faces.restoreBuildTimeDecisions` the value every build time decision evaluated to is saved with the state and replayed by the build which restores the postback:

| tag | what is saved |
| --- | --- |
| `<c:if>` | the test |
| `<c:choose>` | the index of the branch taken |
| `<ui:include>`, `<ui:decorate>`, `<ui:composition>` | the resolved path |
| `<c:forEach>` | the range, the row count and, over a map, the key of each row |
| any component with an expression for an `id` | the id it was rendered under |

The re-apply of the facelet which precedes rendering evaluates everything again, so the response still follows the model, one phase later, and a component whose decision no longer holds is pruned by that refresh exactly as before. The reproducer of the issue therefore renders no input after the submit, and its value reaches the model.

## What is not saved: the elements of an iteration

Each row of a `<c:forEach>` reads its own element live — by position (`IndexedValueExpression`, `IteratedValueExpression`) or by key (`MappedValueExpression`, which holds the key and resolves `src.get(key)` on every read) — so reproducing the rows needs the range, the count and, over a map, the keys, never the elements themselves. A row therefore edits what the model holds now rather than a copy of what it held.

One state entry covers a positional iteration however many rows it produces: **+88 characters of client state for a 50 row iteration, against +3048 for the same view on MyFaces**, which saves `{base, value, index}` per row. A map pays one key per row, because an entry is reached by key rather than by position and neither the count nor the range names the one a row belongs to.

## A row whose element is gone

The rows are reproduced whole, including the ones over items which no longer hold their elements. Such a row reads a stand-in for its element — `null` for every property, swallowing what is written to it, and carrying the key it was rendered for where the iteration is over a map — rather than resolving its var to `null`, which fails the postback outright on the first property any expression reads. That is what full state saving does today: `PropertyNotFoundException: Target Unreachable, identifier [row] resolved to null`.

One warning per iteration reports how many of its rows that happened to, and names what resolves it. Nothing in the saved state can repair an element the model no longer holds, so this makes it visible instead of silent.

Saving the rows would not repair it either, which is worth stating since MyFaces does save them. Rendering two rows from a session scoped list, then posting back with the `items` expression evaluating to an empty list, on one Tomcat, same webapp:

| | outcome |
| --- | --- |
| this branch, parameter off | values dropped, postback completes |
| this branch, parameter on | rows reproduced, values decoded into the stand-in, warning logged, postback completes |
| **full state saving** | **HTTP 500**, `PropertyNotFoundException: Target Unreachable, identifier [row] resolved to null` |
| MyFaces, which saves the rows | values dropped, postback completes |
| MyFaces full state saving | the input is never re-rendered at all, values dropped |

MyFaces saves `{base, value, index}` per row and drops the value here all the same, because for a non transient `items` its saved values drive the row count, the per row id base and `varStatus` while the var itself stays bound by index. The one arm which does materialize the rendered rows is full state saving, and it then fails on the model which no longer holds them: restoring the input does not restore what it was bound to.

## Design

`SavedBuildTimeDecisions` carries a `HashMap` in the state map under `RIConstants.BUILD_TIME_DECISIONS`. `FaceletViewHandlingStrategy.restoreView` starts and stops replaying around its `vdl.buildView` — the state map was already unwrapped there for the view scope state, so nothing had to be reordered. A decision is keyed by the id its tag generates for the build (`ctx.generateUniqueId(tagId)`), which is unique per tag and stable across the builds of one view, and which an iteration generates anew per round so a nested condition replays the value it had in its own round.

A decision is read back only where it is of the type the handler saves, and an iteration's decision also carries the place the tag is written, so a facelet edited since the response was rendered replays nothing rather than another tag's decision. A handler which cannot decide differently between the builds of a view neither saves nor reads one, and the build which composes the metadata of a view records nothing at all, since the view it produces is thrown away.

Restore view does not throw. Everything the replay can be handed — items which are gone, items which have become another kind, a range whose expression only evaluated under what guarded it, a key a map refuses to answer for, a decision of the wrong shape — degrades to building what the model asks for now, and what the model can no longer answer for surfaces in a later phase where it belongs.

This composes with #5960 rather than fighting it: `BuildTimeDecisions.reproducesBuild` re-evaluates each recorded decision against the current model, so a replayed decision which no longer holds denies the render time skip, which is what performs the re-apply that updates the structure.

## What it does not close

A `TagHandler` outside this implementation which decides its structure at build time still rebuilds differently at restore. The save and replay entry points are impl packages here, as they are in MyFaces, so no third party handler can participate portably. Closing that needs the hook in `jakarta.faces.view.facelets`, which is raised at jakartaee/faces#2240.

## Tests

IT module `test/restoreBuildTimeDecisions`, 17 tests: `c:if`, ajax `c:if`, `c:choose`, `c:forEach` over a range, over a map, over a map keyed by the very type an index has, over a set with a range which begins past the first element, over a step which does not advance, over items which are gone, over entries which are gone, over items which switched between map and list in either direction, over keys which cannot be saved, a `c:if` nested in a `c:forEach`, `ui:include`, and an id an expression decided. Full IT suite green.


BalusC:
> 5.0 is where flipping the default belongs

~I'd personally leave it off in 5.0. Better performance and more natural behavior. Still complaint with updated spec. WDYT? The only argument would be Mojarra vs MyFaces divergence.~

Update: It will be default on in 5.0 as per https://github.com/jakartaee/faces/pull/2235#issuecomment-5346391970

mnriem:
@BalusC When I did the implementation I was able to actually get FSS and PSS both to work by introducing a UIIterate that was a stand-in for the c:forEach as well as similar for the others. So the entire tree was saved and available to be fully restored on postback. Thoughts?

BalusC:
> @BalusC When I did the implementation I was able to actually get FSS and PSS both to work by introducing a UIIterate that was a stand-in for the c:forEach as well as similar for the others. So the entire tree was saved and available to be fully restored on postback. Thoughts?

I agree there should be an UIIterate/UIIterable as that simplifies a lot wrt UIData/UIRepeat/UIWhateverCustomIterable (e.g. PrimeFaces TabView, AccordionPanel, etc which was indeed forced to needlessly copypaste many parts of Faces impl internals) but that's beyond scope now. It's at least tracked via https://github.com/jakartaee/faces/issues/963. It was deferred for so long time because of the complexity. With AI assisted development these days it should be easy for 5.1.

Though seeing c:forEach as an UIComponent is kind of strange. It is a TagHandler. It has a different lifecycle. And "the entire tree was saved" is really a bad idea as it violates philosophy of PSS. Continue here please: https://github.com/jakartaee/faces/issues/2240.

BalusC:

BalusC:

PR #5966, opened 2026-08-19
Render an event attribute only for an event the component supports

BalusC:
Two commits, both following jakartaee/faces#2241.

**Follow the three HTML event handler groups of the API.** `HtmlEvents` now names its enums after what they group — `HtmlElementEvent`, `HtmlBodyEvent`, `HtmlWindowEvent` — with `getHtmlElementEventNames()` for every component and `getHtmlBodyEventNames()` for the one rendering the body element. Rename ripple only, across `RenderKitUtils`, eleven renderers and `PassthroughElement`; `jsf:element` keeps the all-element set, matching MyFaces.

**Render an event attribute only for an event the component supports.** The standard HTML render kit ties an `on*` attribute to a client behavior event of the same name when the substring after `on` is contained in `ClientBehaviorHolder#getEventNames()`. Pass-through rendering honored the prefix alone, so an attribute naming an event the component does not support — a typo such as `onclck`, or `onresize` on a component whose element does not fire it — still became a client event listener, while MyFaces renders nothing for it. The gate is one helper consulted from `renderPassThruAttributesOptimized` and `renderPassThruAttributesUnoptimized`. Behaviors themselves need no gate: `UIComponentBase#addClientBehavior` already rejects unsupported event names.

A known renderer attribute that happens to be `on*` but is not in `getEventNames()` now falls to the plain `writeAttribute` branch instead of the behavior branch — same output, since an unsupported event cannot carry behaviors.

Scope call worth reviewing: the gate applies only to `ClientBehaviorHolder` components, which is exactly the premise of the render kit rule. A non-holder — `h:outputText`, `h:outputFormat`, `h:message` — has no event-name list to consult, so `<h:outputText onclick="...">` keeps working as before. MyFaces drops those too, so a sliver of divergence remains, but there it is "component has no event list at all" rather than "event is not in the list". Tightening that further would silently kill a common idiom, so it belongs in the spec discussion on jakartaee/faces#2241 rather than here.

New `BehaviorEventAttributeTest` asserts on the listener path rather than on rendered attributes, since master parks handlers via `addEventListener` and emits `mojarra.ael(...)` on flush: a supported event attaches, a typo attaches nothing, `onpagehide` attaches on `h:body` but not on `h:panelGroup`, and a non-holder still attaches. Full impl suite green at 1334 tests, and the full TCK is green.

Merging is blocked on jakartaee/faces#2241: the faces submodule pin has to be bumped to that merge commit and amended into the first commit here, so the builds will fail until 2241 is merged and PR is updated with bumped submodule pin.

Refs jakartaee/faces#1507

### DIFF
--- a/impl/src/main/java/org/glassfish/mojarra/component/PassthroughElement.java
+++ b/impl/src/main/java/org/glassfish/mojarra/component/PassthroughElement.java
@@ -16,12 +16,12 @@
 
 package org.glassfish.mojarra.component;
 
-import static jakarta.faces.component.html.HtmlEvents.getHtmlBodyElementEventNames;
+import static jakarta.faces.component.html.HtmlEvents.getHtmlElementEventNames;
 
 import java.util.Collection;
 
 import jakarta.faces.component.behavior.ClientBehaviorHolder;
-import jakarta.faces.component.html.HtmlEvents.HtmlDocumentElementEvent;
+import jakarta.faces.component.html.HtmlEvents.HtmlElementEvent;
 
 /**
  * <p>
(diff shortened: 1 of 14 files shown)


BalusC:

Issue #5981, opened 2026-08-31
DefaultFaceletContext.generateUniqueId ArrayIndexOutOfBoundsException under load


pizzi80:
```
java.lang.ArrayIndexOutOfBoundsException: Index 10 out of bounds for length 10
	at com.sun.faces.facelets.impl.DefaultFaceletContext.generateUniqueId(DefaultFaceletContext.java:283)
	at com.sun.faces.facelets.UniqueIdSlot.generateUniqueId(UniqueIdSlot.java:65)
	at com.sun.faces.facelets.tag.faces.ComponentTagHandlerDelegateImpl.generateUniqueId(ComponentTagHandlerDelegateImpl.java:500)
	at com.sun.faces.facelets.tag.faces.ComponentTagHandlerDelegateImpl.apply(ComponentTagHandlerDelegateImpl.java:138)
	at jakarta.faces.view.facelets.DelegatingMetaTagHandler.apply(DelegatingMetaTagHandler.java:139)
	at jakarta.faces.view.facelets.CompositeFaceletHandler.apply(CompositeFaceletHandler.java:57)
	at com.sun.faces.facelets.tag.jstl.core.IfHandler.apply(IfHandler.java:61)
	at jakarta.faces.view.facelets.CompositeFaceletHandler.apply(CompositeFaceletHandler.java:57)
	at jakarta.faces.view.facelets.DelegatingMetaTagHandler.applyNextHandler(DelegatingMetaTagHandler.java:163)
	at com.sun.faces.facelets.tag.faces.ComponentTagHandlerDelegateImpl.apply(ComponentTagHandlerDelegateImpl.java:204)
	at jakarta.faces.view.facelets.DelegatingMetaTagHandler.apply(DelegatingMetaTagHandler.java:139)
	at jakarta.faces.view.facelets.CompositeFaceletHandler.apply(CompositeFaceletHandler.java:57)
	at com.sun.faces.facelets.tag.faces.core.ViewHandler.apply(ViewHandler.java:177)
	at jakarta.faces.view.facelets.CompositeFaceletHandler.apply(CompositeFaceletHandler.java:57)
	at com.sun.faces.facelets.compiler.NamespaceHandler.apply(NamespaceHandler.java:52)
	at com.sun.faces.facelets.compiler.EncodingHandler.apply(EncodingHandler.java:56)
	at com.sun.faces.facelets.impl.DefaultFacelet.include(DefaultFacelet.java:320)
[...]

PR #5982, opened 2026-08-31
DefaultFacelet: fix for issue 5981

pizzi80:
https://github.com/eclipse-ee4j/mojarra/issues/5981

### DIFF
--- a/impl/src/main/java/com/sun/faces/facelets/impl/DefaultFacelet.java
+++ b/impl/src/main/java/com/sun/faces/facelets/impl/DefaultFacelet.java
@@ -29,6 +29,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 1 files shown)


pizzi80:
note that "mvn clean verify" of the 4.0 branch on my PC fails on test Issue5596IT

I've wasted 20 mins only to understand what was the problem with the IT 
only ot find that Issue5596IT is not aligned with 4.1 (now renamed in main) branch

also if I backport the 4.1 Issue5596IT into 4.0
there are still some problems with the chromewebdriver 

imho the 4.0 branch is too old to receive massive improvements,
so in the future I will contribute only to the "main" branch

BalusC:
> only to find that Issue5596IT is not aligned with 4.1 (now renamed in main) branch

Indeed, the PR with the fix (#5670) was targeted to 4.1 instead of 4.0 and never backported.

BalusC:
Backported now https://github.com/eclipse-ee4j/mojarra/commit/c34554830118d08ad7cbbd6c606da92e4cc9fddb (plus unit test for slot ID to prevent this from happening in the future)

PR #5985, opened 2026-09-02
Cumulative code improvements from my fork 

pizzi80:
I'm trying to reduce the diff from official Mojarra...

This is a first PR of cumulative code improvements from my fork 
of things written over years... in production from years...

I've cherry picked minor optimizations and cosmetics and done 
a separate commit for every class (a very time consuming work)

- minor optimizations 
- cosmetics and code clarifications
- documentation fix & typos
- optimized constants initialization
- reduced memory allocation
- properly sized collections
- final on effective final fields
- useless casts
- useless SuppressWarnings
- useless toString() call

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/ActionListenerImpl.java
+++ b/impl/src/main/java/com/sun/faces/application/ActionListenerImpl.java
@@ -44,7 +44,6 @@ import jakarta.faces.event.ActionListener;
  */
 public class ActionListenerImpl implements ActionListener {
 
-    // Log instance for this class
     private static final Logger LOGGER = FacesLogger.APPLICATION.getLogger();
 
     // --------------------------------------------- Methods From ActionListener
@@ -64,7 +63,9 @@ public class ActionListenerImpl implements ActionListener {
         context.renderResponse();
     }
 
-    private String getNavigationOutcome(FacesContext context, MethodExpression expression) {
+    // Utils -----------------------------------------------------------------------------------
+
+    private static String getNavigationOutcome(FacesContext context, MethodExpression expression) {
         if (expression == null) {
             return null;
         }
@@ -83,7 +84,7 @@ public class ActionListenerImpl implements ActionListener {
         }
     }
 
-    private void invokeNavigationHandling(FacesContext context, UIComponent source, MethodExpression expression, String outcome) {
+    private static void invokeNavigationHandling(FacesContext context, UIComponent source, MethodExpression expression, String outcome) {
         NavigationHandler navHandler = context.getApplication().getNavigationHandler();
 
         String toFlowDocumentId = (String) source.getAttributes().get(TO_FLOW_DOCUMENT_ID_ATTR_NAME);
--- a/impl/src/main/java/com/sun/faces/application/ApplicationAssociate.java
+++ b/impl/src/main/java/com/sun/faces/application/ApplicationAssociate.java
@@ -173,7 +173,7 @@ public class ApplicationAssociate {
 
     private Map<String, List<String>> resourceLibraryContracts;
 
-    Map<String, ApplicationResourceBundle> resourceBundles = new HashMap<>();
+    private final Map<String, ApplicationResourceBundle> resourceBundles = new HashMap<>();
 
     public static void setCurrentInstance(ApplicationAssociate associate) {
         if (associate == null) {
@@ -384,7 +384,7 @@ public class ApplicationAssociate {
         Map<String, Object> applicationMap = externalContext.getApplicationMap();
         ApplicationAssociate me = (ApplicationAssociate) applicationMap.get(ASSOCIATE_KEY);
 
-        if (me != null && me.resourceBundles != null) {
+        if (me != null) {
             me.resourceBundles.clear();
         }
 
@@ -394,7 +394,7 @@ public class ApplicationAssociate {
     public static void clearInstance(ServletContext servletContext) {
         ApplicationAssociate me = (ApplicationAssociate) servletContext.getAttribute(ASSOCIATE_KEY);
 
-        if (me != null && me.resourceBundles != null) {
+        if (me != null) {
             me.resourceBundles.clear();
         }
 
@@ -416,7 +416,7 @@ public class ApplicationAssociate {
     /**
      * Obtain the PropertyEditorHelper instance for this app.
      *
-     * @return The PropertyEditorHeler instance for this app.
+     * @return The PropertyEditorHelper instance for this app.
      */
     public PropertyEditorHelper getPropertyEditorHelper() {
         return propertyEditorHelper;
@@ -517,7 +517,7 @@ public class ApplicationAssociate {
         // If there already is a case existing for the fromviewid/fromaction.fromoutcome
         // combination,
         // replace it ... (last one wins).
-        navigationMap.computeIfAbsent(navigationCase.getFromViewId(), k -> new LinkedHashSet<>()).add(navigationCase);
+        navigationMap.computeIfAbsent(navigationCase.getFromViewId(), $ -> new LinkedHashSet<>()).add(navigationCase);
     }
 
     public NamedEventManager getNamedEventManager() {
--- a/impl/src/main/java/com/sun/faces/application/ApplicationResourceBundle.java
+++ b/impl/src/main/java/com/sun/faces/application/ApplicationResourceBundle.java
@@ -24,6 +24,8 @@ import java.util.Map;
 import java.util.ResourceBundle;
 import java.util.concurrent.ConcurrentHashMap;
 
+import com.sun.faces.RIConstants;
+
 /**
  * <p>
  * Contains an application level resource bundle name and its associated descriptions, if any.
@@ -100,7 +102,7 @@ public class ApplicationResourceBundle {
             displayName = queryMap(locale, displayNames);
         }
 
-        return coalesce(displayName, "");
+        return coalesce(displayName, RIConstants.EMPTY_STRING);
     }
 
     /**
@@ -113,7 +115,7 @@ public class ApplicationResourceBundle {
             description = queryMap(locale, descriptions);
         }
 
-        return coalesce(description, "");
+        return coalesce(description, RIConstants.EMPTY_STRING);
     }
 
     // --------------------------------------------------------- Private Methods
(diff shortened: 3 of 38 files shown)


BalusC:
Reviewed at 4db99b1444 against main 2e075aea61.

### Verification

Built `impl` at PR head and at base. The javac warning set is identical: 100 warnings on both sides, differing only in one line number. No new lint. All 1401 unit tests pass.

Not run: TCK, integration tests. This PR touches state saving, navigation and resource resolution, so a TCK pass is needed before merge.

### Verified behavior-neutral

These look risky in the diff. Each one checked out:

- `NavigationMap` keySet/values/entrySet rewrite. `size()` already delegated to `mapToLookForNavCase`, and the old anonymous iterators threw `UnsupportedOperationException` on `remove()`, exactly like `Collections.unmodifiable*`.
- Moving `getDynamicActions()` / `getDynamicComponents()` inside the guard. Both are pure lazy getters and their backing field is only reachable through the getter, so this correctly saves one `ArrayList` / `HashMap` per request.
- `isResetFlowHandlerState` switching from `getCurrentInstance()` to the parameter. The sole caller passes exactly that context.
- `ViewScopeContextObject` final fields. `Serializable` is fine with final fields, and no setter callers exist.
- `Map.of` for `PRIM_MAP` and `STANDARD_CONV_ID_TO_TYPE_MAP`. `Map.of` throws on a null key where `HashMap` returns null, but both `get()` sites are null-guarded, so that path is unreachable.
- `supportedLocales` read twice in `getSupportedLocales`. The field only ever goes from null to non-null.
- `Collections.singletonList` for resource library contracts. The list is never mutated downstream.

### Major

**M1. `Set.of` on `FULL_STATE_SAVING_VIEW_IDS` turns a duplicated view id into a startup failure.**

`ApplicationStateInfo.java:51` reads `Set.of(config.getOptionValue(FullStateSavingViewIds, ','))`. `WebConfiguration.getOptionValue(param, char)` (`WebConfiguration.java:293`) trims entries and drops empty ones, but does not dedupe. `Set.of` throws on duplicates. So `jakarta.faces.FULL_STATE_SAVING_VIEW_IDS = /a.xhtml,/a.xhtml` makes the constructor throw `IllegalArgumentException: duplicate element: /a.xhtml` while `ApplicationAssociate` is being built, and the application fails to deploy. The old `new HashSet<>(...)` plus `addAll` deduped silently.

Reporting the config mistake is right. Failing the deployment over it is not. The exception also names neither the parameter nor the fact that it came from a context param, so it does not tell the deployer what to fix.

**Fix:** dedupe and warn.

```java
String[] viewIds = config.getOptionValue(FullStateSavingViewIds, ',');
fullStateViewIds = Set.copyOf(asList(viewIds));

if (fullStateViewIds.size() != viewIds.length) {
    LOGGER.warning("Duplicate view IDs in '" + FullStateSavingViewIds.getQualifiedName() + "' are ignored.");
}
```

`Set.copyOf` keeps the immutability the change was after and ignores duplicates. `FullStateSavingViewIds.getQualifiedName()` avoids `StateManager.FULL_STATE_SAVING_VIEW_IDS_PARAM_NAME`, which is deprecated and marked for removal. The `asList` static import has to come back.

### Minor

**m2. `PersistenceContextHandler.applyToField` keeps the `trim()` the commit set out to remove.**

`PersistenceContextHandler.java:67` reads `!unit.name().trim().isBlank()`. The three sibling call sites became plain `isBlank()`: `PersistenceContextHandler.java:58`, `PersistenceUnitHandler.java:57` and `PersistenceUnitHandler.java:66`. This one allocates a trimmed copy for nothing, and it reads as if the trim were load-bearing.

**Fix:** drop the `.trim()`.

**m3. Five imports left dangling by the PR's own edits.**

`WebappLifecycleListener.java:23` (`java.util.Arrays.asList`), `ConverterConfigHandler.java:20` (`ArrayList`), `ConverterConfigHandler.java:22` (`Collections`), `ResourceManager.java:22` (`Arrays`), `FaceletFullStateManagementStrategy.java:43` (`Iterator`). Each is the only occurrence of its simple name in the file after the change.

**Fix:** remove all five.

**m4. The `unmodifiableMap` wrap sits one brace too shallow.**

`InstanceFactory.java:623`. The wrap moved into the `synchronized` block but stayed outside the inner `if (defaultValidatorInfo == null)`. A thread that passes the outer check while another thread is still building re-wraps an already-wrapped map, nesting `UnmodifiableMap` layers. That much is harmless.

It also leaves a pre-existing NPE window open. `addDefaultValidatorId` (`InstanceFactory.java:594`) nulls the field without holding the monitor, so a thread reaching line 623 after that null-out calls `unmodifiableMap(null)`. The old code had the same window, only wider: the wrap was outside the lock entirely.

**Fix:** move `defaultValidatorInfo = unmodifiableMap(defaultValidatorInfo);` inside the inner `if`, as the last statement after the fill loop. That kills the double wrap and closes the NPE window in one move.

**m5. Making `classMap` final leaves two unreachable null checks in place.**

`FaceletFullStateManagementStrategy.java:238` reads `classMap != null ? classMap.get(...) : null` and line 241 reads `isAnyNull(clazz, classMap)`. `classMap` is assigned unconditionally at line 112, so both `classMap` operands are provably non-null. The `isAnyNull` call now reads as if `classMap` could vanish mid-restore.

**Fix:** use `Class<?> clazz = classMap.get(treeNode.componentType);` and `if (clazz != null)`. This is the payoff of the `final` the commit added; without it the `final` buys nothing.

**m6. `SystemEventInfo`'s cache lambda captures `this` from inside the constructor.**

`SystemEventInfo.java:38` reads `new Cache<>(sourceClass -> new EventInfo(this.systemEvent, sourceClass))`. The explicit `this.` forces the lambda to capture the partially constructed instance and hand it to `Cache`. It is safe today, because the factory only runs from `getEventInfo`, long after construction, and javac's `this-escape` lint is not enabled here so nothing fires. The escape is still gratuitous.

**Fix:** drop the `this.`, so the lambda captures the constructor parameter instead: `new Cache<>(sourceClass -> new EventInfo(systemEvent, sourceClass))`. Same semantics, no capture of `this`.

**m7. `NavigationMap` still inherits `AbstractMap.get()`, so every navigation lookup is a linear scan.**

`NavigationHandlerImpl.java:1459` overrides `size`, `isEmpty`, `put`, `putAll`, `keySet`, `values` and `entrySet`, but not `get` or `containsKey`. `AbstractMap.get` walks `entrySet()`. `getViewId` calls `navMap.get(...)` at lines 657, 715 and 756, so that runs on every navigation, once per rule in the app.

This is pre-existing, not introduced here. But this PR rewrites exactly those three view methods, and the O(n) lookup is worth an order of magnitude more than the view-allocation saving.

**Fix:** add `@Override public Set<NavigationCase> get(Object key) { return mapToLookForNavCase.get(key); }` and the matching `containsKey`. To keep this PR strictly cosmetic, split it into its own issue rather than dropping it.

**m8. `SESSION_LISTENERS` holds attribute names, not listeners.**

`WebappLifecycleListener.java:57` declares `private static final String[] SESSION_LISTENERS = {VIEW_SCOPE_MANAGER, CLIENT_WINDOW_SCOPE_MANAGER};` and line 144 consumes it as `servletContext.getAttribute(listenerName)`. The hoist itself is a real win: it drops a `List` plus an array per session destroy. Only the name misleads.

**Fix:** rename to `SESSION_LISTENER_ATTRIBUTE_NAMES`.

**m9. The `// can't be null` trailing comments restate the code.**

`SystemEventHelper.java:42-43`. Removing the null check is correct: `Cache.get` (`Cache.java:74`) only returns null if the factory does, and `SystemEventInfo::new` cannot. But the comments explain what the code does. They do not state a constraint the next editor could violate.

The javadoc at line 39 reads `@return get the EventInfo (eventually creating a new one)`. "Eventually" here means "if necessary", which is a false friend, and a `@return` tag should not start with "get".

**Fix:** delete both trailing comments. Reword the javadoc to "@return the `EventInfo` for the given system event and source class, creating and caching it if absent."

### Nits

**n10. `if ( cond )` spacing.** `ConverterConfigHandler.java:70` and `FaceletViewHandlingStrategy.java:363` add spaces inside the parens. Nothing else in either file does that. **Fix:** write `if (cond)`.

**n11. `Level.FINE` next to the statically imported `FINE`.** `FaceletViewHandlingStrategy.java:363` writes `LOGGER.isLoggable(Level.FINE)` and line 364 writes `LOGGER.log(FINE, ...)`. The switch away from the `Supplier` overload is worth keeping, as it drops a capturing lambda per `buildView`, so per request. **Fix:** use the imported `FINE` on both lines.

**n12. Redundant local.** `NavigationHandlerImpl.java:91-92` assigns `obtainingNavigationCase` and returns it on the next line. **Fix:** `return Boolean.TRUE.equals(facesContext.getAttributes().get(RESET_FLOW_HANDLER_STATE_KEY));`

### Follow-up: generalize M1's check on 5.0

M1 fixes one parameter. Every other multi-value context param has the same blind spot: duplicates pass unreported. On 5.0 that can be caught in one place.

The right choke point is `WebConfiguration.resolve(ContextParam)` at `org/glassfish/mojarra/config/WebConfiguration.java:379`. It holds both `param.getName()` and the raw declared value, it already reports resolved params and warns about unusable ones (`faces.config.webconfig.boolconfig.invalidvalue`), and its results land in `resolvedValues` (line 96), so it runs once per param per application and cannot spam. It covers every `String[]`-typed param at once.

Two places look right but are not. `ContextParam.Separator.split` (`ContextParam.java:362`) does the splitting but has no param name and no logger. `ContextParam.toValue` (line 287) has the name but is also called from `resolveUnreported`, which is deliberately silent.

`Separator.split` already trims entries and drops empty ones, so duplicates are the only anomaly left to report. Shape: after `param.toValue(value)` returns, if the param type is `String[].class` and the resolved array holds fewer distinct elements than entries, warn naming the param and the duplicated values. Test home is `impl/src/test/java/org/glassfish/mojarra/config/ContextParamResolutionTest.java`, with `UnrecognizedContextParamTest` as the precedent for asserting a config warning.

### Meta

Behavior-affecting changes (the `Set.of` switch, M1) are interleaved with pure cosmetics inside the same commits. The one-commit-per-class split does help. A follow-up PR would land easier if the few behavior-affecting commits were separated from the whitespace and typo ones, because those can go in on sight.

BalusC:
To be clear, I can also ask Claude to address all of them. No problem.

pizzi80:
If you prefer to keep the Claude style also for standardization, 
I'm fine! so we can move on ;)

BalusC:
Ok.

It found 19 things btw, I already dropped 7 of them on review because they were nonsense and then I added the 5.0 proposal to generalize the M1 check on all other multi-value context params.

pizzi80:
yes maybe in the cherry-picking process I made some mistakes ... 

in some situation I had to manually copy/paste things because I have also other changes
in the same class but they will be part of future PRs ...

PR #5986, opened 2026-09-02
FastStringWriter improvements + documentation + cosmetics

pizzi80:
implements all the Writer methods 
invoking the right native StringBuilder methods to improve performance

### DIFF
--- a/impl/src/main/java/com/sun/faces/io/FastStringWriter.java
+++ b/impl/src/main/java/com/sun/faces/io/FastStringWriter.java
@@ -21,7 +21,7 @@ import java.io.Writer;
 
 /**
  * <p>
- * This is based on {@link java.io.StringWriter} but backed by a {@link StringBuilder} instead.
+ * A {@link Writer} implementation backed by a {@link StringBuilder}.
  * </p>
  * 
  * <p>
@@ -30,7 +30,7 @@ import java.io.Writer;
  */
 public class FastStringWriter extends Writer {
 
-    protected StringBuilder builder;
+    protected final StringBuilder builder;
 
     // ------------------------------------------------------------ Constructors
 
@@ -59,27 +59,36 @@ public class FastStringWriter extends Writer {
         builder = new StringBuilder(initialCapacity);
     }
 
+    /**
+     * <p>
+     * Constructs a new <code>FastStringWriter</code> instance using the specified <code>builder</code>.
+     * </p>
+     * @param builder the builder to use as internal buffer
+     */
+    public FastStringWriter(StringBuilder builder) {
+        this.builder = builder;
+    }
+
     // ----------------------------------------------------- Methods from Writer
 
     /**
      * <p>
      * Write a portion of an array of characters.
      * </p>
-     *
-     * @param cbuf Array of characters
-     * @param off Offset from which to start writing characters
-     * @param len Number of characters to write
-     *
+     * @param str Array of characters
+     * @param off  Offset from which to start writing characters
+     * @param len  Number of characters to write
      * @throws IOException
      */
     @Override
-    public void write(char cbuf[], int off, int len) throws IOException {
-        if (off < 0 || off > cbuf.length || len < 0 || off + len > cbuf.length || off + len < 0) {
+    public void write(char[] str, int off, int len) throws IOException {
+        // this check it's implemented also in the StringBuilder class ... probably can be removed
+        if (off < 0 || off > str.length || len < 0 || off + len > str.length || off + len < 0) {
             throw new IndexOutOfBoundsException();
         } else if (len == 0) {
             return;
         }
-        builder.append(cbuf, off, len);
+        builder.append(str, off, len);
     }
 
     /**
@@ -107,12 +116,26 @@ public class FastStringWriter extends Writer {
     // ---------------------------------------------------------- Public Methods
 
     /**
+     * Write a single character.
+     * @param c the String to be written
+     */
+    @Override
+    public void write(int c) throws IOException {
+        builder.append((char)c);
+    }
+
+    /**
      * Write a string.
      *
      * @param str String to be written
      */
     @Override
     public void write(String str) {
+        builder.append(str);
+    }
+
+    @Override
+    public void write(char[] str) throws IOException {
         builder.append(str);
     }
 
@@ -125,7 +148,7 @@ public class FastStringWriter extends Writer {
      */
     @Override
     public void write(String str, int off, int len) {
-        builder.append(str.substring(off, off + len));
+        builder.append(str, off, off + len);
     }
 
     /**
@@ -145,6 +168,27 @@ public class FastStringWriter extends Writer {
 
     public void reset() {
         builder.setLength(0);
+    }
+
+
+    // ------------------------------------------------- Append Methods
+
+    @Override
+    public Writer append(CharSequence s) throws IOException {
+        builder.append(s);
+        return this;
+    }
+
+    @Override
+    public Writer append(CharSequence s, int start, int end) throws IOException {
+        builder.append(s, start, end);
+        return this;
+    }
+
+    @Override
+    public Writer append(char c) throws IOException {
+        builder.append(c);
+        return this;
     }
 
 }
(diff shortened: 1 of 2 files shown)


melloware:
@pizzi80 does PF need this too?  https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/java/org/primefaces/util/FastStringWriter.java

BalusC:
Reviewed at 26eb570382 against main 2e075aea61.

### Verification

Built `impl` at PR head. No warnings on the changed file. 1401 unit tests pass, 0 failures and 4 skipped. Jest: 356 passed.

I also ran all five new overrides against the `java.io.Writer` contract.

### Contract check

All five are correct.

| call | result | `Writer` spec |
|---|---|---|
| `write(0xFFFF0041)` | `A` | low 16 bits written, high bits ignored |
| `write("abcdef", 1, 3)` | `bcd` | off/len to start/end conversion |
| `write("abc", 1, 5)` | `IndexOutOfBoundsException` | as specified |
| `append(null)` | `null` | "the four characters null" |
| `append(null, 1, 3)` | `ul` | as if csq contained "null" |
| `append('x')` | returns `this` | as specified |

The `off, off + len` conversion in `write(String, int, int)` is the one place this could have silently inverted. It did not.

### The wins, sized

`write(int)` is the one that matters. The base `Writer.write(int)` enters `synchronized (lock)`, lazily allocates `writeBuffer = new char[1024]` (2 KB per writer instance), then dispatches to `write(char[], 0, 1)`. `HtmlResponseWriter` calls `writer.write('<')`, `write('>')`, `write(' ')` and `write('"')` on every element and every attribute (`HtmlResponseWriter.java:513, 519, 597, 715, 719, 722, 729`), and `cloneWithWriter` hands the `FastStringWriter` straight to it with nothing buffering in between (`HtmlResponseWriter.java:226`). Synthetic loop of 6 char writes plus 3 string writes per iteration: 130 ms to 42 ms, so 3.0x. Real renders do more work per character, so the end-to-end share is much smaller, but the mechanism is real.

The rest: dropping the `substring` in `write(String, int, int)` removes one `String` copy per call; the base `append(CharSequence)` goes through `String.valueOf(csq)` and the base `append(CharSequence, int, int)` through `csq.subSequence(...).toString()`, so both allocated a full copy; `write(char[])` and `append(char)` are small.

### Major

**M1. Five new public overrides and a new constructor, zero tests.**

There is no `FastStringWriterTest` anywhere in `impl/src/test`, and this PR adds `write(int)`, `write(char[])`, three `append` methods and a constructor without one. The project has test setup, 1401 tests. The `off, off + len` conversion at `FastStringWriter.java:151` and the `(char) c` truncation at `:124` are exactly what a test pins for free. Both had to be verified by hand, and nothing keeps them correct after the next edit.

**Fix:** add `impl/src/test/java/com/sun/faces/io/FastStringWriterTest.java` asserting the six rows in the contract table above. That is the whole test, and it is worth more than the rest of the findings combined.

Say the word and I write it and push it to this branch, so you do not have to.

### Minor

**m2. `FastStringWriter(StringBuilder)` is unused, aliases the caller's buffer, and accepts null.**

`FastStringWriter.java:68`. Nothing in the repo calls it. The caller keeps a reference to the buffer, so the writer's content changes underneath it: seeding with `new StringBuilder("seed:")`, writing `x`, then appending `!aliased` to the original yields `seed:x!aliased` from `toString()`. `new FastStringWriter(null)` also constructs fine and defers the NPE to the first write.

**Fix:** drop the constructor. It is scope creep in a performance PR, no caller needs it, and `getBuffer()` already covers the read side. If it stays because a caller is coming, add `Util.notNull("builder", builder)`.

**m3. A speculative comment shipped in the source.**

`FastStringWriter.java:85` reads `// this check it's implemented also in the StringBuilder class ... probably can be removed`. The claim is right: `AbstractStringBuilder.append(char[], int, int)` calls `checkRange(off, off + len, str.length)`, which covers every case the explicit check does, including the `off + len` overflow. But "probably can be removed" is a note to self, not a statement about the code.

**Fix:** delete the check and the comment together. `StringBuilder` throws `StringIndexOutOfBoundsException`, a subclass of the `IndexOutOfBoundsException` the `Writer` contract names, and its message is better than the current bare throw. Keeping the check and deleting only the comment is the safer half-measure if you would rather not touch behavior in this PR.

**m4. `write(int)` javadoc describes the wrong parameter.**

`FastStringWriter.java:120` reads `@param c the String to be written`. It is a character, passed as an int.

**Fix:** write `@param c the character to write; the 16 high-order bits are ignored`. That second clause is the part worth documenting, because `(char) c` truncates silently.

**m5. The `char[]` parameter was renamed `cbuf` to `str`.**

`FastStringWriter.java:84`. `Writer.write(char[] cbuf, int off, int len)` names it `cbuf`, and `str` reads as a `String` in a method that takes an array. `write(char[] str)` at `:138` inherits the same name.

**Fix:** keep `cbuf` in both.

**m6. The new methods have no javadoc, in a file that documents everything else.**

`write(char[])` at `:137` and the three `append` methods at `:176`, `:182` and `:188`.

**Fix:** one line each, or state on the class that the `append` methods follow `Writer` and add nothing.

### @melloware: does PrimeFaces need this too?

Yes, for one of the six changes, and it is the one that matters.

`primefaces/src/main/java/org/primefaces/util/FastStringWriter.java` already has the `write(String, int, int)` fix. Line 74 is `builder.append(str, off, off + len)`, no substring. So the biggest allocation win is already in.

What PrimeFaces does not have is `write(int)`. PrimeFaces wraps its `FastStringWriter` in a cloned `ResponseWriter` (`ComponentUtils.java:590` and `:653`) and encodes whole components into it, so every write comes from the impl's `HtmlResponseWriter`. Both impls write single characters constantly: Mojarra at `HtmlResponseWriter.java:513, 519, 597, 715, 719, 722, 729`, MyFaces at 15 sites in `HtmlResponseWriterImpl.java` (`write('<')`, `write('>')`, `write(' ')`, `write('"')`). Every one of those lands in `java.io.Writer.write(int)`, which takes a monitor and lazily allocates a 1024-char buffer per writer before delegating. In the synthetic loop that path was 3.0x slower than the override.

The minimal patch for PrimeFaces:

```java
@Override
public void write(int c) {
    builder.append((char) c);
}

@Override
public Writer append(char c) {
    builder.append(c);
    return this;
}
```

`append(char)` is included because the base implementation routes straight into `write(int)`.

The other two `append` overrides (`CharSequence` and `CharSequence, int, int`) avoid a `toString()` copy each, but only pay off if something appends a non-`String` `CharSequence`, which the `ResponseWriter` path does not. `write(char[])` saves a single delegating call. Both are cheap to add and harmless, but neither is why you would do this.

One caveat on the 3.0x: my benchmark writes characters and nothing else, so the `synchronized` block is almost the whole cost there. In a real render, character writes are a small slice of the work per component, so the same absolute saving is a much smaller share of the total. Expect a low single-digit percentage on component capture, not 3x.

pizzi80:
in the PF implementation the method

```
public void write(String str, int off, int len) {
    builder.append(str, off, off + len);
}
```

is implemented correctly, while the original Mojarra  
was wrong and was creating a copy of the passend String...

you could declare the builder `final` and you should be on par ...

I've also implemented the `append` methods to be aligned with the StringBuilder implementation 
but is questionable if they are micro-faster or not than the original one ... 

all the append method inside Writer class call the method write, 
so it should be equivalent ...

pizzi80:
Lol Claude was faster than me ... ;)

about the cosmetics... 

I prefer to give proper name to things and (not a C copy/pasted ugly name char cbuf[])

I have aligned the name of the variable to the name of `StringBuilder` methods
because we are using a `StringBuilder` internally

BalusC:
Just let me know if I should let Claude address all of them. No problem.

melloware:
Let Claude do the work!

pizzi80:
the saddest part of the story is that Java is really really really late 
on implementing modern version of Input/Output classes 
and on every real Java project the are copies of this unsynchronized 
implementations which duplicate code and grows the CodeCache and Metaspace in the jvm at runtime
(not to mention the potential bugs and bugfix work)

pizzi80:
> Just let me know if I should let Claude address all of them. No problem.

ok! no problem!

PR #5988, opened 2026-09-02
HtmlUtils optimizations & cosmetics

pizzi80:
final + optimizations + code standardizations + FastByteArrayOutputStream unsynchronized + cosmetics

### DIFF
--- a/impl/src/main/java/com/sun/faces/util/HtmlUtils.java
+++ b/impl/src/main/java/com/sun/faces/util/HtmlUtils.java
@@ -19,7 +19,6 @@ package com.sun.faces.util;
 import static java.lang.Character.isHighSurrogate;
 import static java.lang.Character.isLowSurrogate;
 
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
@@ -40,8 +39,10 @@ import com.sun.faces.RIConstants;
 public final class HtmlUtils {
     private HtmlUtils() {}
 
+    private static final String ISO_8859_1 = StandardCharsets.ISO_8859_1.name();
+
     private static final Set<String> UTF_CHARSET = Set.of("UTF-8", "UTF-16", "UTF-16BE", "UTF-16LE", "UTF-32", "UTF-32BE", "UTF-32LE",
-            "x-UTF-16LE-BOM", "X-UTF-32BE-BOM", "X-UTF-32LE-BOM", "");
+            "x-UTF-16LE-BOM", "X-UTF-16LE-BOM", "X-UTF-32BE-BOM", "X-UTF-32LE-BOM", "");
 
     // -------------------------------------------------
     // The following methods include the handling of
@@ -479,7 +480,7 @@ public final class HtmlUtils {
      * percent-encoding is lazily allocated on first need and shared across all encode calls in
      * a single invocation -- matching the original allocation amortization.
      */
-    static private void encodeURIString(Writer out, char[] textBuff, String encoding, int start, int end) throws IOException {
+    private static void encodeURIString(Writer out, char[] textBuff, String encoding, int start, int end) throws IOException {
         int runStart = start;
         boolean fragment = false;
         FastByteArrayOutputStream buf = null;
@@ -659,7 +660,7 @@ public final class HtmlUtils {
     // ----------------------------------------------------------
     //
     public static boolean isISO8859_1encoding(String encoding) {
-        return StandardCharsets.ISO_8859_1.name().equals(encoding);
+        return ISO_8859_1.equals(encoding);
     }
 
     // ----------------------------------------------------------
@@ -667,7 +668,7 @@ public final class HtmlUtils {
     // ----------------------------------------------------------
     //
     public static boolean isUTFencoding(String encoding) {
-        return UTF_CHARSET.contains(encoding);
+        return encoding != null && UTF_CHARSET.contains(encoding);
     }
 
     // ----------------------------------------------------------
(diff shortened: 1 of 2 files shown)


pizzi80:
a message for Claude about cosmetics... :D
we are human and we are slow to read code if is written out of "standard"

BalusC:
Reviewed against a local apply of this PR onto 53b17d3006. Build plus the full impl unit suite is green on JDK 21: 1376 tests, 0 failures. That includes the 67 `HtmlUtilsTest` cases which cover the `writeURL` / `encodeURIString` percent-encoding path that the new `FastByteArrayOutputStream` sits under. So the swap itself is functionally proven.

The `static public` → `public static` sweep, the missing `final` on the entity and name tables, and the class-level `final` plus constructor hoist are all straight wins. Findings below.

## Major

**M1. `Set.of` turns `isUTFencoding(null)` from `false` into an NPE.**

`HtmlUtils.java:41`. `ImmutableCollections.SetN.contains` starts with `Objects.requireNonNull(o)`, while `HashSet.contains(null)` returns `false`. Verified on JDK 21: `Set.of("UTF-8", "").contains(null)` throws `NullPointerException`. `isUTFencoding` is a public method. There is a mitigation, which is why this is not critical: the only caller in the tree is `HtmlResponseWriter.java:284` and `:286`, and the value it passes already survived `HtmlUtils.validateEncoding(encoding)` → `Charset.isSupported`, which throws on null first. So no current path reaches it. Neither method has a test.

**Fix:** keep `Set.of`, it is the better collection here, and null-guard the method: `return encoding != null && UTF_CHARSET.contains(encoding);`. Going back to `HashSet` would also fix it, but then you lose the immutability this PR is after. Please add a test for `isUTFencoding(null)`.

## Minor

**m2. `FastByteArrayOutputStream` is `public` but has no caller outside `HtmlUtils`.**

`HtmlUtils.java:748`. Every use is in the same class: lines 404, 430, 485, 519 and 537. A grep over `impl/src` finds nothing else. `MyByteArrayOutputStream` was `private`. Making it public inside a public class adds it to the impl surface for no gain.

**Fix:** make it `private static final class`.

**m3. Three members of the new class are dead.**

The no-arg constructor is never called, both sites use `new FastByteArrayOutputStream(MAX_BYTES_PER_CHAR)`. `writeTo(OutputStream)` has no caller anywhere in the repository. The `close()` override reimplements the no-op it already inherits from `OutputStream`. The used surface is exactly `write(int)`, `write(byte[], int, int)`, `reset()`, `size()` and `getBuffer()`.

**Fix:** delete all three.

**m4. The class javadoc describes a class this no longer is.**

`HtmlUtils.java:743-746` says "Implementation of `ByteArrayOutputStream`", but the class extends `OutputStream`. The `java.io.ByteArrayOutputStream` import now only exists to satisfy that javadoc link. Also "that expose" should be "that exposes".

**Fix:** reword to something like "Unsynchronized `OutputStream` that accumulates into a `byte[]` and hands it out directly", and drop the `ByteArrayOutputStream` import.

**m5. The optimization in the title is on a cold path, and half of the new code cannot run there.**

`FastByteArrayOutputStream` is only constructed for a character outside `(32, 127)` in a URL, see `HtmlUtils.java:430` and `:519`. Every such character already pays a full `OutputStreamWriter` / `StreamEncoder` encode plus a `flush()`. Two uncontended monitor operations are invisible next to that. The cost side is worse than it looks: the buffer starts at `MAX_BYTES_PER_CHAR = 10` and `reset()` runs after every single character, so `ensureCapacity` and `grow` never fire for any charset that encodes one char in 10 bytes or less, which is all of them in practice. That means roughly 25 of the ~60 added lines are hand-maintained overflow arithmetic, copied from the pre-JDK-9 `ByteArrayOutputStream` body rather than the current `ArraysSupport.newLength`, that will never execute. The PR gives no measurement.

**Fix:** keep the class, it is correct and covered by tests, but trim it per m2 and m3, and drop the performance claim from the commit message unless you can back it with a number. As it stands this is hygiene, not a speedup. Reverting to the `ByteArrayOutputStream` subclass is defensible too, but that gives up the one thing the rewrite really buys: no inherited synchronization on a class nobody shares.

**m6. `StandardCharsets.ISO_8859_1.name()` replaces a literal with a per-call getter chain.**

`HtmlUtils.java:662`. `.name()` does return `"ISO-8859-1"`, I verified it, so the behavior is identical. But a `String` constant was clearer and can be folded at compile time. It also makes the two encoding checks inconsistent, because `UTF_CHARSET` just above still holds string literals.

**Fix:** hoist `private static final String ISO_8859_1 = StandardCharsets.ISO_8859_1.name();` and compare against that. This keeps the JDK as the source of truth without the per-call chain. Going back to the plain literal is fine as well.

## Nit

**n7. `UTF_CHARSET` holds one entry that can never match.**

`HtmlUtils.java:41-42`: `"x-UTF-16LE-BOM"` has a lowercase `x`, while its siblings `"X-UTF-32BE-BOM"` and `"X-UTF-32LE-BOM"` are uppercase, and the only caller uppercases the encoding first, see `HtmlResponseWriter.java:269` with `encoding.toUpperCase(Locale.ROOT)`. This is pre-existing, but the PR rewrites exactly this declaration.

**Fix:** uppercase it to `"X-UTF-16LE-BOM"` in the same commit. A separate follow-up is fine too, but please do not re-land the line with an entry that is dead by construction.

---

Summary: M1 should be fixed before merge. m2 to m4 are cheap. m5 is the main point for discussion: the title promises an optimization that this path cannot deliver, and the extra code has a real maintenance cost.

BalusC:
I'll address them.

pizzi80:
a note for 
**m2. FastByteArrayOutputStream is public but has no caller outside HtmlUtils.**

I have refactored to keep the PR inside the class, it should be extracted and used 
also in another place of Mojarra (I've found a similar inner class in another mojarra class...)

BalusC:
> a note for **m2. FastByteArrayOutputStream is public but has no caller outside HtmlUtils.**
> 
> I have refactored to keep the PR inside the class, it should be extracted and used also in another place of Mojarra (I've found a similar inner class in another mojarra class...)

You can extract at the exact moment when you need it.

BalusC:
> a message for Claude about cosmetics... :D we are human and we are slow to read code if is written out of "standard"

It helps to be consistent with surrounding code. Claude does that too. So if you introduce e.g. `if ( foo )` while rest of code is `if (foo)` then it will also briefly block active code reading/scanning by human and Claude (and me) will usually flag it.

BalusC:
Btw, I have introduced Spotless in OmniFaces and OmniHai to automatically format code conform predefined set of rules. I think I will do the same on 5.0 branches.

pizzi80:
interesting ... I bet that if you introduce Spotless here you have to rewrite half of the project, 
but could be the right time to do such a change... we have to improve this project

pizzi80:
OT: I will use OmniHai soon ... ;)

BalusC:
> interesting ... I bet that if you introduce Spotless here you have to rewrite half of the project, but could be the right time to do such a change... we have to improve this project

Indeed! :)

PR #5993, opened 2026-09-02
UrlBuilder optimizations

pizzi80:
tried to optimize at my best to avoid String and collections copy
and properly init because UrlBuilder is used in many places

### DIFF
--- a/impl/src/main/java/com/sun/faces/context/UrlBuilder.java
+++ b/impl/src/main/java/com/sun/faces/context/UrlBuilder.java
@@ -63,7 +63,7 @@ class UrlBuilder {
     // ------------------------------------------------------------ Constructors
 
     public UrlBuilder(String url, String encoding) {
-        if (url == null || url.isBlank()) {
+        if (Util.trimToNull(url) == null) {
             throw new IllegalArgumentException("Url cannot be empty");
         }
         this.url = new StringBuilder(url.length() * 2);
@@ -79,7 +79,7 @@ class UrlBuilder {
     // ---------------------------------------------------------- Public Methods
 
     public UrlBuilder addParameters(String name, List<String> values) {
-        name = trimToNull(name);
+        name = Util.trimToNull(name);
         if (name == null) {
             throw new IllegalArgumentException("Parameter name cannot be empty");
         }
@@ -91,7 +91,7 @@ class UrlBuilder {
     public UrlBuilder addParameters(Map<String, List<String>> params) {
         if (params != null && !params.isEmpty()) {
             for (Map.Entry<String, List<String>> entry : params.entrySet()) {
-                String paramName = trimToNull(entry.getKey());
+                String paramName = Util.trimToNull(entry.getKey());
                 if (paramName == null) {
                     throw new IllegalArgumentException("Parameter name cannot be empty");
                 }
@@ -103,7 +103,7 @@ class UrlBuilder {
     }
 
     public UrlBuilder setPath(String path) {
-        if (path == null || path.isBlank()) {
+        if (Util.trimToNull(path) == null) {
             throw new IllegalArgumentException("Path cannot be empty");
         }
         this.path = path;
@@ -163,7 +163,7 @@ class UrlBuilder {
         for (String pair : pairs) {
             String[] nameAndValue = Util.split(pair, PARAMETER_NAME_VALUE_SEPARATOR);
             // ignore malformed pair
-            if (nameAndValue.length != 2 || nameAndValue[0].isBlank()) {
+            if (nameAndValue.length != 2 || Util.trimToNull(nameAndValue[0]) == null) {
                 continue;
             }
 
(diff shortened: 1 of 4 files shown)


pizzi80:
note that 
```
 List<String> retValues = values.stream().filter(Objects::nonNull).collect(Collectors.toList());
 addValuesToParameter(entry.getKey().trim(), retValues, true);
```

is not needed because inside `addValuesToParameter`
the is a null check inside the list for loop

pizzi80:
but Claude will probably find an edge case where everything will explode :D

BalusC:
Thanks for the patch. The existing `UrlBuilderTest` passes 31/31 against this branch (verified locally on JDK 21).

The clear win is dropping the `values.stream().filter(Objects::nonNull)` in `addParameters(Map)`. `addValuesToParameter` already filters nulls, so that stream was pure duplication. Presizing the `ArrayList` and using `List.of(value)` are fine too. `trimToNull` is where the problems are.

## Major

**M1. `trimToNull` returns `strip()`, so control characters now survive into parameter names, which are written into the URL unencoded.**

`UrlBuilder.java:313-321` returns `str.strip()`. `Character.isWhitespace` is false for U+0000-U+0008 and U+000E-U+001B, while the old `name.trim()` removed everything up to and including U+0020. `appendQueryString()` writes `param.getKey()` raw. Only *values* go through `URLEncoder`. Verified: `addParameters("\u0000foo", List.of("bar"))` produced `?foo=bar` before, and produces `?\u0000foo=bar` now. Same for `"\u0007foo\u0007"` and `"foo\u0000"`. That URL ends up in an `href` or in a `Location` header.

**Fix:** keep the wider rejection, but restore the old value semantics:

```java
if (str == null || str.isBlank()) {
    return null;
}
String trimmed = str.trim();
return trimmed.isEmpty() ? null : trimmed;
```

This rejects Unicode-blank input (the improvement you intended) and control-character-only input (the old behavior), and it returns a trimmed name. Returning `str.strip().trim()` instead would chain two scans for no gain on realistic input.

**M2. The `isBlank()` swap makes the constructor and `setPath` accept control-character-only input that they used to reject.**

`UrlBuilder.java:66` and `UrlBuilder.java:106` change `trim().length() == 0` into `isBlank()`. Verified: `"\u0000".isBlank()` is `false`, but `"\u0000".trim().isEmpty()` is `true`. So `new UrlBuilder("\u0000")` no longer throws, and `appendPath()` appends the NUL verbatim. The inverse also flipped: `"\u2000"` is now rejected where it used to pass, which is the good half of the change.

**Fix:** use the same shape as M1 at both places, so the guard is `str.isBlank() || str.trim().isEmpty()`. Please do not simply revert to `trim().length() == 0`, because that loses the Unicode-blank rejection this PR correctly adds.

**M3. `trimToNull` has four return branches and no test, in a class that already has a 31-test suite.**

`UrlBuilderTest.addParameters_emptyOrBlankNameIsRejected` (line 281) covers blank names, but only with ASCII spaces. So every branch this PR adds is untested. The PR also changes `addParameters(Map)` from throwing NPE on a null value list into accepting it. That is an improvement, but it is untested as well.

**Fix:** extend `addParameters_emptyOrBlankNameIsRejected` with `"\u2000"` and `"\u0000"`, add a case that asserts a control-character-prefixed name is normalized, and add `addParameters_nullValueListIsTolerated`.

## Minor

**m4. `trimToNull` duplicates helpers that already exist in `Util`.**

`Util.java:555` has `isBlank(String)` and `Util.java:563` has `nullIfBlank(String)`. `UrlBuilder` already imports `Util`. The new helper is also `public static` on a package-private class, which buys nothing.

**Fix:** move it to `Util` as `Util.trimToNull`, next to `nullIfBlank`, and call it from all four places. That also fixes M2 in one place. Keeping it local as `private static` is only defensible if you want to avoid growing `Util`, but `setPath` needs it too.

**m5. The helper makes the very path the PR title claims to optimize slower.**

`addParameters(String, List)` used to do `isBlank()` plus `trim()`. That is two scans, and `trim()` returns `this` without allocating when the string is already clean. It now does `isEmpty()` plus `trim()` plus `strip()`, so three scans. Only `addParameters(Map)` gets faster, because two separate `trim()` calls became one.

**Fix:** the M1 rewrite brings it back to two scans. No separate change needed.

**m6. Six per-line comments in `trimToNull` repeat what the code says, and one of them is wrong.**

`UrlBuilder.java:314-320`. The comment `// trim remove leading and trailing control chars and whitespace` sits on a line whose result is immediately overwritten by `strip()`. So it describes behavior the method does not have, which is exactly the M1 defect.

**Fix:** delete all six. The rejection rule belongs in the test names.

**m7. Brace and spacing style differs from the rest of the file.**

`if ( str == null || str.isEmpty() ) return null;` pads the parentheses and omits the braces. The `final` on the locals in `addValueToParameter` and `addValuesToParameter` appears nowhere else in the class.

**Fix:** use braces, drop the inner padding, drop the `final`s. Note that `if ( oldValues != null ) oldValues.addAll(values);` is already in the file from an earlier PR. That one slipped through and is worth cleaning up in the same pass, rather than being treated as the local convention.

---

The javadoc `fragement` to `fragment` typo fix is unrelated to the topic of this PR, but it is harmless. Please keep it.

BalusC:
I'll apply them.

PR #5995, opened 2026-09-02
FaceletViewHandlingStrategy optimizations

pizzi80:
FaceletViewHandlingStrategy optimizations + reduce object allocation + final + fix internal classes name

### DIFF
--- a/impl/src/main/java/com/sun/faces/application/view/FaceletViewHandlingStrategy.java
+++ b/impl/src/main/java/com/sun/faces/application/view/FaceletViewHandlingStrategy.java
@@ -885,11 +885,10 @@ public class FaceletViewHandlingStrategy extends ViewHandlingStrategy {
     protected void initialize() {
         LOGGER.fine("Initializing FaceletViewHandlingStrategy");
 
-        final FacesContext context = FacesContext.getCurrentInstance();
-        final ExternalContext extContext = context.getExternalContext();
-        final Map<String, Object> appMap = extContext.getApplicationMap();
-
         initializeMappings();
+
+        // The context must be resolved per lookup: the cache outlives the request.
+        metadataCache = new Cache<>(ccResource -> createComponentMetadata(FacesContext.getCurrentInstance(), ccResource));
 
         try {
             responseBufferSize = Integer.parseInt(webConfig.getOptionValue(FaceletsBufferSize));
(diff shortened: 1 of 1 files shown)


BalusC:
# PR 5995 — FaceletViewHandlingStrategy optimizations

Target: `main`. One commit, one file: `impl/src/main/java/com/sun/faces/application/view/FaceletViewHandlingStrategy.java`. Build is green.

Nothing critical or major.

Worth keeping as-is: the `Regarget` -> `Retarget` typo renames on the private nested handlers, dropping the no-op anonymous `VariableMapperWrapper` subclass, and making `createComponentMetadata` static. That last one also makes the `metadataCache` lambda non-capturing, so it becomes a single instance instead of one per strategy. The method body touches no instance state, so `static` is sound.

## Minor

**m1. The `EnumSet` branch in `returnAsImplicitOutCome` is unreachable and allocates.**

`FaceletViewHandlingStrategy.java:2012-2016`. `ViewVisitOption` (`impl/src/main/java/jakarta/faces/application/ViewVisitOption.java:42`) has exactly one constant, `RETURN_AS_MINIMAL_IMPLICIT_OUTCOME`. So `options.length >= 2` only happens if a caller passes the same constant twice, which makes `EnumSet.of(options[0], options)` dead in practice. It also re-passes `options[0]` inside the varargs array, and it allocates a set to answer a membership question. It would NPE on a null array element, where the old `stream(options).anyMatch(...)` and the new length-0/1 branches return `false`. The method runs once per `getViews()` call, not per request, so there is no throughput to win here.

**Fix:** replace the body with a plain loop -- `for (ViewVisitOption option : options) { if (option == RETURN_AS_MINIMAL_IMPLICIT_OUTCOME) { return true; } } return false;`. That keeps the zero-allocation path the PR wants, drops the dead branch, and still works if a second enum constant is ever added. Reverting to the original `stream(...)` one-liner is the other defensible option, but it gives up the real win.

**m2. `initialize()` now assigns `metadataCache` after logging "Initialization Successful".**

`FaceletViewHandlingStrategy.java:903-904`. The reshuffle also hoists `context`, `extContext` and `appMap` to the top of the method, about 25 lines above their only use at `appMap.remove(...)` on line 908. No functional change, but the log line no longer marks the end of initialization, and the locals are separated from what they feed.

**Fix:** move the `metadataCache` assignment back above the `try` block, and keep `appMap` next to its use. A single `Map<String, Object> appMap = FacesContext.getCurrentInstance().getExternalContext().getApplicationMap();` right before the `remove` drops the two intermediate locals.

**m3. The new cache comment is ungrammatical and sits in the wrong place.**

`FaceletViewHandlingStrategy.java:903`: `// note that the Cache Factory need to retrieve always the current FacesContext instance!`. The constraint it states is real and worth keeping -- the context must not be hoisted out of the lambda -- but the wording is broken and it lands under the "Initialization Successful" log.

**Fix:** reword to one line above the assignment once m2 moves it back, e.g. `// The context must be resolved per lookup, not captured -- the cache outlives the request.`

## Nit

**n4. Formatting drifts from the file.**

`Map<String,MethodRetargetHandler>` is missing the space after the comma (line 1429). There is a stray space before the comma on the `ActionListenerRetargetHandler` entry (line 1431). `if ( options.length == 0 ) return false;` (line 2013) uses spaces inside parens with no braces; the file uses neither.

**Fix:** add the space after the comma in `Map<String, MethodRetargetHandler>`, drop the stray space before the comma, and write `if (options.length == 0) {` with braces. The repo has no formatter, checkstyle or editorconfig to defer to, so this is matching surrounding code: the file has zero occurrences of either spaces-inside-parens or `<K,V>`. Braceless single-statement `if` is the weaker of the three -- 42 occurrences across 616 `com.sun.faces` files, one of them already in this file.

BalusC:
I'll address them.

pizzi80:
for this release we're done ... I'll continue after 4.1.16 so you can deliver something in the meantime ... :D

PR #5996, opened 2026-09-02
Basic components: EVENT_NAME modernization + valueEventName constant definition

pizzi80:


pizzi80:
I found that there are some attributes that are marked for removal in 4.0 and 4.1 but are still there ...

BalusC:
Current main branch is 4.1. They're removed in 5.0. Also 5.0 got new HtmlEvents enum and defaults are centralized so these changes will conflict there.

BalusC:
# PR 5996 — Basic components: EVENT_NAME modernization

Target: `main`. One commit, 23 files under `impl/src/main/java/jakarta/faces/component/html/`.

Verified mechanically: all 23 `EVENT_NAMES` lists are element-identical to `main`, same order, no duplicates, and every new `DEFAULT_EVENT` value is present in its own list. So the pure refactoring part is clean.

## Major

**M1. `List.of(...).contains(null)` throws NPE where the old collection returned `false`.**

`UIComponentBase.addClientBehavior` (`impl/src/main/java/jakarta/faces/component/UIComponentBase.java:1899-1910`) calls `eventNames.contains(eventName)` with no null guard. `Collections.unmodifiableCollection(Arrays.asList(...))` answers `false` for a null argument; `List.of(...)` throws `NullPointerException`. I confirmed both. So `component.addClientBehavior(null, behavior)` silently no-ops today and blows up after this PR, on all 23 standard components at once.

Two mitigations, both real. `PassthroughElement.java:358` already uses `List.of` on `main`, so the hazard is established precedent rather than new. Both in-tree callers guard null first: `AjaxHandler.applyAttachedObject` (line 270) and `AjaxBehaviors.BehaviorInfo.addBehavior` (line 129). The exposure is third-party callers of the public spec API.

And it is already the situation on 5.0 regardless of this PR: `HtmlEvents` builds every event-name collection with `Collectors.toUnmodifiableList()`, which is the same `ImmutableCollections` family, and `UIComponentBase.addClientBehavior` in the faces API (line 1922 on `faces` branch `5.0`) has the same unguarded `contains`.

**Fix:** add the guard in `UIComponentBase.addClientBehavior` -- `if (eventName != null && eventNames.contains(eventName))` -- rather than reverting the 23 files. One line, it also covers third-party `ClientBehaviorHolder` implementations that adopt `List.of`, and it keeps `PassthroughElement` consistent instead of leaving one class on a different idiom. The same one-line guard is worth applying to the faces `5.0` API copy, where the NPE is already live. Reverting to `unmodifiableCollection` is the alternative, but it gives up the cleanup and leaves both `PassthroughElement` and all of 5.0 exposed anyway.

## Minor

**m2. Every hunk in this PR is in code 5.0 has already deleted.**

On mojarra `5.0` the `jakarta.faces.*` sources moved into the `faces` submodule, and on `faces` branch `5.0` the per-component `EVENT_NAMES` constant is gone entirely. `HtmlCommandButton.getEventNames()` is now `return getFacesActionSourceEventNames(getFacesContext());` and `getDefaultEventName()` is `return FacesComponentEvent.action.name();` (`api/src/main/java/jakarta/faces/component/html/HtmlCommandButton.java:828-835`). `HtmlBody` likewise delegates to `getHtmlBodyEventNames(getFacesContext())`. The whole design lives in `HtmlEvents`, driven by the `HtmlElementEvent` / `HtmlBodyEvent` / `HtmlWindowEvent` enums plus the `jakarta.faces.ADDITIONAL_HTML_EVENT_NAMES` context-param, with an application-scoped cache.

So on upmerge `main` -> `5.0` all 23 files conflict, and every conflict resolves by taking the 5.0 side. Nothing in this PR survives. 5.0 also already solves what `DEFAULT_EVENT` is reaching for, via `FacesComponentEvent.action.name()` / `.valueChange.name()`.

**Fix:** merge it for `main` if you want `main` tidy in the meantime, but flag it in the upmerge as "take ours" so nobody tries to reconcile the hunks. Given m3 below, the smaller variant is worth considering: keep the `List.of` conversion, drop `DEFAULT_EVENT`, which shrinks the conflict surface to the field declarations alone.

**m3. `DEFAULT_EVENT` buys nothing and is named off its method.**

Fifteen files gain `private static final String DEFAULT_EVENT = "action"` or `"valueChange"`, used at exactly one call site each, e.g. `HtmlCommandButton.java:826` and `:838`. javac already inlines the interned literal, so no allocation or lookup is saved. `PassthroughElement.getDefaultEventName()` -- same idiom, already on `main` -- still returns the bare literal, so the codebase ends up split. The name also does not match `getDefaultEventName()`.

**Fix:** drop the constant and keep `return "action";`. This also directly shrinks the 5.0 upmerge conflict from m2. If it stays, rename it `DEFAULT_EVENT_NAME` and apply it to `PassthroughElement` too, so the convention is uniform.

BalusC:
I'll address them.

PR #5997, opened 2026-09-02
faces.js getFormsToUpdate optimization

pizzi80:


BalusC:
Reviewed on a worktree at the PR head. I built `impl` and ran the Jest suite on both the PR head and `main`. The existing suite passes on the PR, so it is blind to both defects below. I wrote two throwaway tests that pass on `main` and fail on the PR head.

### C1 (critical). Replacing `element.getElementsByTagName(FORM)` with `document.forms` makes `add()` recurse infinitely as soon as the page holds one non-Faces form.

The `else` branch is reached when `element` *is* a form but is not valid. Previously that branch scanned `element`'s descendants; nested forms are illegal HTML, so it returned empty and the recursion terminated. Now it iterates `document.forms` and calls `add()` on every form, including the invalid one it just came from, which re-enters the same branch forever.

Input: a page with `<form id="testForm" method="post">` plus any plain `<form>` (no id, or `method="get"` — a search box, a third-party widget, or in a namespaced view any form outside `namingContainerPrefix`), then any ajax request. Observed on the PR head:

```
console.error
  malformedXML: Maximum call stack size exceeded (HTTP 200) [source: testButton]
```

The `jakarta.faces.ViewState` update is aborted, so every following postback on that page carries a stale view state. `render="@all"` hits this too: `add(document)` walks `document.forms` and the first invalid form blows the stack.

**Fix:** revert this line to `element.getElementsByTagName(FORM)`. That is also the correct scope (see M2), so one revert closes both. A visited-set guard would stop the crash but keep the wrong scope, so please do not use that.

### M2 (major). The same substitution widens `add()` from "forms inside this render target" to "every form in the document".

`add()` is called once per render client id. For `render="panel"` where `panel` is a `<div>`, every valid form in the page now lands in `formsToUpdate`. That includes forms which already hold a view-state field, which the second pass deliberately skips.

Test on the PR head: `testForm` plus an unrelated `<form id="outsideForm" method="post">` carrying a `ViewState` field, request with `render="panel"`. On the PR, `outsideForm`'s field is overwritten. On `main` it is left alone.

**Fix:** covered by the C1 revert. If updating out-of-scope forms is wanted, that is a separate behavior change and needs its own PR with a rationale, not a commit titled "optimization".

### M3 (major). The change is a pessimization, not an optimization.

With N render targets and M forms in the document, `add()` now costs O(N*M), where it was O(sum of target subtree sizes). The only real win in this PR is the second-pass change, which is a live-collection swap worth roughly nothing.

**Fix:** keep the second-pass change (`document.forms` is equivalent there and marginally cheaper), drop the change inside `add()`. If you want a measurable win, the second pass is the real cost — it runs `getHiddenStateField` over every form on every state update — but that needs a benchmark first.

### M4 (major). No test, in a file that has a Jest suite.

`impl/src/test/ts/spec/faces.ajax.test.ts` already covers `getFormsToUpdate` indirectly, e.g. "ViewState update creates hidden field if absent in other form". Neither defect above is covered, which is why `mvn clean test` is green on the PR head.

**Fix:** add two cases to `faces.ajax.test.ts`. First: an ajax request on a page containing a form that fails `isValidForm`, asserting no `onerror` fires and the ViewState is updated. Second: a `render` of a non-form container, asserting a valid form outside that container keeps its existing ViewState. These are worth adding regardless of what happens to this PR: `add()`'s termination currently rests on an unstated invariant (nested `<form>` is illegal HTML), and a test is the only thing that will hold it.

### m5 (minor). `instanceof HTMLFormElement` is fine here, but it is realm-sensitive.

Every caller passes `document`, `document.getElementById(...)`, or an element from `document.forms`, all same-realm, so this is equivalent to the `nodeName` check and cheaper. It would silently return false for a form from an iframe document.

**Fix:** keep it. One line of javadoc on `add()` stating the same-document precondition would help, so a future caller passing a node from a parsed response document does not get a silent false.

### m6 (minor). The `isValidForm` reorder is a no-op.

All four operands are side-effect-free truthiness checks, so reordering only changes which one short-circuits first. That is not measurable. `form.elements` is dead either way: it is always truthy on an `HTMLFormElement`. That part predates this PR.

**Fix:** harmless, keep it. You could drop the dead `form.elements` check while you are there.

### Verdict

Request changes. Revert the hunk inside `add()`; the other three hunks are fine.

BalusC:
I will check tomorrow, I'm afk soon.

pizzi80:
thanks!
... I'm fine also to revert if we find that was better before ... 

just some notes:
- in OmniFaces you have a similar utility that uses document.forms ... 
- this is in production since at least 1 year in a medium size webapp that 
  uses plain mojarra & omnifaces (no PF) with tons of ajax ... 
  really strange what Claude is saying but could be right anyway

BalusC:
It's because add() is recursive and document.forms returns ALL forms while element.getElementsByTagName(FORM) only returns forms among children of the given element.

BalusC:
I added unit test which ensures it fails when document.forms is used again in add() in some future.

pizzi80:
> It's because add() is recursive and document.forms returns ALL forms while element.getElementsByTagName(FORM) only returns forms among children of the given element.

you're right!!! 

it seems an uncommon situation, isn't it?

Issue #6012, opened 2026-09-15
CdiUtils can be a little bit blocker


halcsi19790320:
---

  Imnportant

  No third-party libraries are involved. The reproducer below is a plain JUnit test that runs against a stub BeanManager in Mojarra's own test source tree — no PrimeFaces, no other vendor, no CDI implementation at all on the reproducing path.

  ---

  Describe the bug

  com.sun.faces.cdi.CdiUtils keeps six static bean-resolution caches, each a synchronizedMap(new WeakHashMap<>()) (CdiUtils.java lines 108, 117, 130, 132, 134, 136 on main):

  private static final Map<BeanManager, ConcurrentMap<BeanLookupKey, Bean<?>>> RESOLVED_BEANS = synchronizedMap(new WeakHashMap<>());
  private static final Map<BeanManager, Bean<?>>                               FACES_CONTEXT_PRODUCER_BEANS = ...;
  private static final Map<BeanManager, ConcurrentMap<Object, Bean<?>>>        CONVERTER_BEANS_BY_KEY = ...;
  private static final Map<BeanManager, ConcurrentMap<Object, Bean<?>>>        VALIDATOR_BEANS_BY_KEY = ...;
  private static final Map<BeanManager, ConcurrentMap<Object, Bean<?>>>        BEHAVIOR_BEANS_BY_KEY  = ...;
  private static final Map<BeanManager, ConcurrentMap<Object, Bean<?>>>        VIEW_FACELET_BEANS_BY_KEY = ...;

  Collections$SynchronizedMap acquires its monitor for every operation, reads included. Because every Application.createConverter / createValidator / createBehavior call passes through cachedBean() — per component, per render, from every request thread — these are effectively a single process-wide monitor that all request processing serializes on, even when the cache is fully warm and no CDI resolution is left to do.

  Production thread dump (many threads in this state simultaneously):

  "default task-125" #4051 [45176] prio=5 os_prio=0 cpu=576,03ms elapsed=11,48s tid=0x0000592dbec4f050 nid=45176 waiting for monitor entry  [0x00007d699c3f9000]
     java.lang.Thread.State: BLOCKED (on object monitor)
          at java.util.Collections$SynchronizedMap.computeIfAbsent(java.base@25.0.4.1/Collections.java:3077)
          - waiting to lock <0x000000074f21a790> (a java.util.Collections$SynchronizedMap)
[...]

PR #6013, opened 2026-09-15
Make the CdiUtils bean-resolution caches lock-free on the read path

halcsi19790320:
(#6012)

The six caches were synchronizedMap(new WeakHashMap<>()), whose monitor is
taken for every operation including reads. Every Application.create* call
passes through them, so one process-wide monitor serialized request processing even on a fully warm cache.

Replace them with a volatile copy-on-write snapshot holding one cache per
application, populated only by registerBeanManager at deployment time and
dropped by unregisterBeanManager at shutdown. Nothing else can enter the map, which makes an unbounded map and a retained class loader structurally
impossible; an unregistered BeanManager resolves without caching.

Each register/unregister pair matches by identity, so no assumption is made
about BeanManager equality, which CDI leaves unspecified. Shutdown eviction
is now per-application, replacing the wholesale clearCaches().


BalusC:
The solution appears to be written with WAR-provided Mojarra impl in mind (e.g. Tomcat/Jetty) and is not compatible with the common case of container-provided Mojarra impl in forever-running (production) servers (normal JEE servers such as GlassFish/WildFly/Liberty/etc). 

In case of container-provided Mojarra, it will leak memory when WAR is undeployed from a running server and it will degrade the cache when WAR is redeployed against a running server which still contains the previous deployment leaked in memory.

The PR in its current form is unfortunately not acceptable, but the idea is definitely sound. I'll take a second look myself.

BalusC:
Thank you for the PR. 

The solution appears to be written with WAR-provided Mojarra+CDI in mind (for barebones servletcontainers like Tomcat/Jetty which don't provide Mojarra nor CDI by itself) and is not compatible with the common case of container-provided Mojarra+CDI impls in forever-running (production) servers (normal JEE servers such as GlassFish/WildFly/Payara/etc which themselves provide Mojarra+CDI JARs which are shared across all WARs). 

It'll work fine in barebones servletcontainers with WAR-provided Mojarra+CDI, but in case of container-provided Mojarra+CDI, it will leak memory when WAR is undeployed from a running server and it will degrade the cache when WAR is redeployed against the same running server which still contains the BeanManager map key of the previous deployment leaked in memory.

The PR in its current form is unfortunately not acceptable without a full rewrite, but the idea is definitely sound. I'll take a second look myself.

BalusC:
Superseded by https://github.com/eclipse-ee4j/mojarra/pull/6017

halcsi19790320:
I changed the mechanyzm too. 'createConverter' and other methods doesn't put to cache, only used. Register create a holder, and by undeploy it is removed.
```
public static void registerBeanManager(BeanManager beanManager)

public static void unregisterBeanManager(BeanManager beanManager)
```

And after unregister the cache is unused.
Unregister is called 'ConfigureListener.contextDestroyed' and 'CdiExtension.beforeShutdown'

I hoped it can work.

Issue #6014, opened 2026-09-16
Relative ui:include from exploded META-INF/resources is rejected as outside application root since Mojarra 4.1.13


martin654:
**Describe the bug**
Mojarra 4.1.13 rejects a relative ui:include when the Facelets are served from an exploded dependency’s META-INF/resources.

The resources are found correctly, but DefaultFaceletFactory rejects the resolved file: URL because it is outside the physical webapp directory:
`FacesFileNotFoundException: included.xhtml is not within the application root`

The same application works with Mojarra 4.1.12 and when the dependency is packaged as a JAR.

**To Reproduce**

Reproducer: [exploded_reproducer.zip](https://github.com/user-attachments/files/32277806/exploded_reproducer.zip)
1. Run:
```
  mvn clean package
   mvn -f webapp/pom.xml jetty:run -Pexploded
```
2. Open http://localhost:8080/reproducer/relative.xhtml.
3. Observe HTTP 500 and FacesFileNotFoundException.
4. Open http://localhost:8080/reproducer/absolute.xhtml.
5. Observe that the equivalent absolute include works.

**Expected behavior**

The relative include should resolve successfully because both Facelets are valid application resources under META-INF/resources.

**Additional context**
Mojarra 4.1.12 + exploded resources: works
Mojarra 4.1.13 + exploded resources: fails
Mojarra 4.1.13 + packaged JAR: works

BalusC:
Confirmed. This is regression of CVE security fix done in https://github.com/eclipse-ee4j/mojarra/pull/5908.

pizzi80:
I also had a lot of problems with the associated tests, in my fork I had to rewrite some of that methods ... 

I can provide a PR but I bet that you have a better overall understanding of what is going on

BalusC:
> I also had a lot of problems with the associated tests, in my fork I had to rewrite some of that methods ...
> 
> I can provide a PR but I bet that you have a better overall understanding of what is going on

They work fine for me and in CI.

Do you use Windows instead of *nix? I vaguely recall you do. There should theoretically be no problem but you never know with Windows. Sharing the exact test error/exception/stacktrace will be helpful.

pizzi80:
I'm on fedora 44 ... don't know why but on my fork the "includeRce*" tests never worked 
... if I switch to main they works 
... probably some maven configuration? 
... asked to claude and he says that there are problems with those methods

I still have a lot of things to merge but nothing is specific to paths or similar

[...]

PR #6015, opened 2026-09-16
Draft: ... resolveURL fixes with the help of Claude

pizzi80:
not working upstream but working on my forked mojarra and viceversa


pizzi80:
https://github.com/eclipse-ee4j/mojarra/issues/6014

BalusC:
According to Claude Opus 5 this happens to solve #6014 as well so I think your original problem was also #6014 itself. But the proposed solution is not acceptable as it regresses in other use cases (among others, it allows encoded traversal and incorrectly rejects paths containing spaces and brackets).

pizzi80:
yes I've created the PR only to give you some ideas ...

BalusC:
Superseded by https://github.com/eclipse-ee4j/mojarra/commit/3c9b49b2e9d4d527664a7c278d763f693c283ac5, can you verifiy if that works for you as well?

Issue #6024, opened 2026-09-18
ELFlash NPE: PreviousNextFlashInfoManager.innerMap becomes null in distributable clustered application; deleting csfcfc recovers affected client


AndrewG10i:
## Describe the bug

We intermittently encounter a `NullPointerException` in Mojarra `ELFlash` in a clustered/distributable Jakarta EE application.

The relevant exception is:

```text
java.lang.NullPointerException: Cannot invoke "java.util.Map.put(Object, Object)"
because "this.innerMap" is null
    at com.sun.faces.context.flash.ELFlash$PreviousNextFlashInfoManager.initializeBaseCase(ELFlash.java:1175)
    at com.sun.faces.context.flash.ELFlash.getCurrentFlashManager(ELFlash.java:1066)
    at com.sun.faces.context.flash.ELFlash.loggingGetPhaseMapForWriting(ELFlash.java:761)
    at com.sun.faces.context.flash.ELFlash.setKeepMessages(ELFlash.java:309)
    at com.sun.faces.application.NavigationHandlerImpl.handleNavigation(NavigationHandlerImpl.java:234)
    at com.sun.faces.application.NavigationHandlerImpl.handleNavigation(NavigationHandlerImpl.java:199)
    at com.sun.faces.application.ActionListenerImpl.invokeNavigationHandling(ActionListenerImpl.java:92)
    at com.sun.faces.application.ActionListenerImpl.processAction(ActionListenerImpl.java:61)
    at jakarta.faces.component.UIViewAction.broadcast(UIViewAction.java:507)
    at jakarta.faces.component.UIViewRoot.broadcastEvents(UIViewRoot.java:864)
    at jakarta.faces.component.UIViewRoot.processDecodes(UIViewRoot.java:1005)
    at com.sun.faces.lifecycle.ApplyRequestValuesPhase.execute(ApplyRequestValuesPhase.java:51)
    at com.sun.faces.lifecycle.Phase.doPhase(Phase.java:72)
    at com.sun.faces.lifecycle.LifecycleImpl.execute(LifecycleImpl.java:131)
    at jakarta.faces.webapp.FacesServlet.executeLifecyle(FacesServlet.java:689)
    at jakarta.faces.webapp.FacesServlet.service(FacesServlet.java:449)
```

Once this occurs for a particular browser/user, subsequent requests to the application continue failing with an error page.

Removing only the Mojarra Flash cookie:

```text
csfcfc
```

[...]

Issue #2130, opened 2011-06-23
Flash scope cookie enables data exploits


ren-zhijun-oracle:
The implementation of the flash scope in Mojarra 2.x is based on a cookie that contains an index into an application scoped map. This index is a simple integer. New values are obtained by the implementation by incrementing an AtomicInteger:

```
**ELFlash.java**private long getNewSequenceNumber() {
    long result = sequenceNumber.incrementAndGet();

    if (0 == result % numberOfFlashesBetweenFlashReapings) {
        reapFlashes();
    }

    if (result == Long.MAX_VALUE) {
        result = 1;
        sequenceNumber.set(1);
    }

    return result;
}
```

This algorithm makes the index very predictable. Since the map that contains all flash data is application scoped, an attacker only needs to known an index in order to access data from random other users. As mentioned this index is very easy to guess. One scheme would be to observe the index being handed out for a particular request (say 1500), then take a slightly higher number (say 2000) and continuously poll the server using this cookie. Eventually this number will be generated server side and it's possible the attacker's request will reach the server before that of the legitimate user.

This exploit is even more severe since all current Mojarra implementations (2.0.x and 2.1.x) contain other bugs that prevent the flash scope from being immediately expired (see [http://java.net/jira/browse/JAVASERVERFACES-1751](http://java.net/jira/browse/JAVASERVERFACES-1751)). The user has to do a couple of post backs before this expiration actually happens. This means in the current situation there's a very wide window where the attacker can access data with a guessed index.

Since the Flash scope can contain any kind of data (like e.g. personal information, credit card details, messages, etc) this seems be a rather severe problem.

One solution would be to use a GUID instead of a sequential number. Another solution could be to (optionally) store flash data in the HTTP session and thus piggyback on the security of the JSESSIONID.
#### Affected Versions
[2.1.1]

Issue #1453, opened 2009-12-09
Modify flash to support clustering


ren-zhijun-oracle:
Currently the flash implementation is entirely stored in the Application scope. This
makes it unclusterable. Make it so data stored in the flash is cluster friendly.
#### Environment
Operating System: All
Platform: All
#### Affected Versions
[2.0.2]

BalusC:
> [@javaserverfaces](https://github.com/javaserverfaces) Commented Marked as **incomplete** on Monday, December 10th 2012, 6:23:57 am

It shouldn't have been closed. It's now at least finally fixed via https://github.com/eclipse-ee4j/mojarra/pull/6025.

PR #6025, opened 2026-09-20
Fix #6024 and the clustered flash defects behind it

BalusC:
Fixes the NPE reported in #6024, and the two defects it turned out to be sitting on. Four commits, each reviewable on its own.

## The reported bug

`SessionHelper.update()` read the replicated session attribute and handed it to `ELFlash.setFlashInnerMap()` with no null check. The flash inner map is application scoped, so a session which activated without that attribute left every session on the node with a null map, and every later request threw the reported NPE out of `PreviousNextFlashInfoManager.initializeBaseCase` until the application was redeployed. `SessionHelper` has been byte identical since the 2018 initial contribution and had no test at all.

`ELFlash.getFlash()` also passed a null flash into the same block when asked not to create one, which `FlashFactory.getFlash(false)` exposes as public API. That block has now been null guarded three times over eleven years, once per null it can hold: #3625 for the session, #4431 for the helper, and this one for the flash itself.

## What that uncovered

**A cookie from one node reaches another browser's flash.** Flashes were keyed by a sequence number alone, handed out by an `AtomicLong` private to each node and started at zero, so two nodes issue the same numbers. A probe of the real `decode()` path, with two `ELFlash` instances standing in for two nodes:

```
node A first sequence number: 1
node B first sequence number: 1
node A's cookie resolved on node B to: {residentSecret=node B user's data}
it is the resident session's flash map? true
```

This needs no failover and no guessing. It is the part of #2130 that its fix does not cover: #2130 was closed in 2013 by encrypting the cookie, which stops an attacker guessing an index but does nothing about two nodes legitimately issuing the same one.

Flashes are now filed under an owner first, a truncated SHA-256 of the session id, and `PreviousNextFlashInfoManager` is given the flashes of the owner of the current request. **The owner is derived server side on every request, so the cookie keeps carrying the sequence number alone and its format does not change** - no upgrade step, and the session id never leaves the server. The same probe after the change resolves to an empty map and leaves the resident flash untouched.

**Every session replicated the whole node's flash map.** `sessionMap.put(FLASH_INNER_MAP_KEY, flash.getFlashInnerMap())` stored the application scoped map itself, the same object, in every session:

```
session A attribute keys: [sessionB-entry, sessionA-entry]
session B attribute keys: [sessionB-entry, sessionA-entry]
same object in both sessions? true
is it the application map?    true
```

So one user's replicated payload carried every other user's flash values and grew with the traffic of the node instead of with the session, and activating a session swapped that whole map in. Each session now publishes and takes back its own bucket only. Taking it back is additive, through one atomic `flashInnerMap.compute`, so a request already in flight for the same session does not lose its writes and the node keeps a `ConcurrentHashMap` as the live bucket rather than adopting a deserialized `HashMap` as shared state. Nothing swaps the application scoped map any more, so it is `final`.

Reaping follows: session flashes are dropped when the session is destroyed and re-keyed when the container rotates its id, which it does on login, where a flash is typically written before the redirect and read after it. `reapFlashes()` is left with the flashes of sessionless requests, whose sequence numbers all come from this node's counter, which is what makes the distance to that counter a measure of age in the first place.

MyFaces avoids all of this by storing flash in the session map under a random token, with no application scoped map and no reaper. That is not portable here: it forces session creation, which #3625 and #3478 deliberately went the other way on.

## Also here

`maven.compiler.source`/`target` do not constrain the API surface, so on 4.0 a call to an API that Java 11 does not have compiles into a class file claiming to run on Java 11 and only fails once it reaches a Java 11 JVM. This branch hit exactly that. Switched to `maven.compiler.release`, which 4.1 and 5.0 already use, and verified: `impl` plus 56 modules of the test reactor recompile clean under it. This commit is 4.0 only and should not be upmerged.

## Tests

`SessionHelperTest` is the first test this class has ever had. `FlashOwnerTest` covers owner isolation, session destroy, id rotation and the sessionless bucket. `ELFlashTest`'s hard-coded sequence numbers now pin the relationship rather than the literals, since the counter is seeded randomly, as MyFaces' `TokenGenerator` does, to keep two nodes apart in the sessionless bucket too.

Full impl build green, 1449 tests.

## Open

Deleting `csfcfc` recovering a single browser, as reported in #6024, does not follow from anything found here, for three reasons. The map is on the application scoped `ELFlash`, so one browser's cookie cannot scope it. Nothing assigns that field except the `ELFlash` constructor, so no sequence of requests turns it back into a map. And the direction is the other way around - with the map null, a request which carries a cookie survives, because `decode()` wraps its body in `catch (Throwable)`, while a request without one reaches `initializeBaseCase` and throws exactly the reported stack:

```
COOKIE PRESENT path (decode):
  survived; ForceSetMaxAgeZero set? true
NO COOKIE path (initializeBaseCase):
  THREW java.lang.NullPointerException: ... "this.innerMap" is null
```

The surviving path sets `ForceSetMaxAgeZero`, so the response asks the browser to drop `csfcfc`, and the next request from it takes the throwing path. That matches "every later request fails", but it means the browser reaches the cookieless state by itself.

So this branch fixes a defect whose stack matches the report exactly, and does not explain the recovery step. @AndrewG10i, two things would help: whether the failure was really limited to one browser or whether other users on the same node were failing at the same time, and whether anything else happened around the recovery (a node restart, or the requests landing on the other node). This branch is the one to test.

Refs #2130 and #1453. #1453 is where `SessionHelper` came from, and its original resolution was *incomplete*, not fixed.


AndrewG10i:
> So this branch fixes a defect whose stack matches the report exactly, and does not explain the recovery step. @AndrewG10i, two things would help: whether the failure was really limited to one browser or whether other users on the same node were failing at the same time, and whether anything else happened around the recovery (a node restart, or the requests landing on the other node). This branch is the one to test.

First of all, thank you so much for addressing this issue so quickly! We really appreciate it and I cannot wait to get the fix merged into the application server so that we can test it as soon as possible.

As per the statement below mentioned in my original report - this may not be fully accurate:

> `Deleting only csfcfc immediately restores normal operation.`

Unfortunately, we didn't have the ability - and, more importantly, didn't have sufficient time during the incidents - to perform detailed diagnostics on the affected end users' browsers.

Based on what was reported to us, the issue started affecting several users, most likely users connected to the same application node.

A few days before I opened the issue, the situation became significantly worse. As a last-resort mitigation (temporary workaround to stabilize situation with NPEs in ELFlash), we implemented `csfcfc` cookie cleanup in both directions at the Nginx reverse proxy:

1. For responses to end users, we added a header instructing the browser to delete the `csfcfc` cookie.
2. For requests forwarded upstream to the Jakarta Faces application, we stripped the `csfcfc` cookie so that it was never passed to the application.

According to our logs, the `ELFlash` NPE has not occurred again since this workaround was introduced.

Therefore, the statement in the original issue that deleting the cookie restored operation may indeed be somewhat misleading. It is quite possible that the affected node was already in a broken state for all users routed to it, rather than this being an issue isolated to an individual user's cookie/session.

